PR #2, opened 2016-05-12
Add first pass at a bootstrap UI

zeroasterisk:
See my README w/ a basic roadmap... Let's talk about what we can collaboate on to facilitate making my `FormField` wrapper work better (inherit grid, schema, and errors) and your package have a few extra features (autosave).


radekmie:
Wow, that's a lot of work! Autosave could be very handy indeed - I'll work on it in my free time.

Issue #24, opened 2016-06-14
Radio-Field should use transform-property to display the value


macrozone:
Hi there, awesome work so far, looks really promising and clean!

I found one oddity, though: In all RadioField-Variants, the value is displayed directly :

` <label>{item}</label>`

I think it should be: ` <label>{transform(item)}</label>` with `transform` beeing a prop passed to the field.

I would do a PR, but i wanted to ask, if I understand this right?

serkandurusoy:
That indeed looks like an overside.

You can take cue from the select component as a proper example and
implement a PR that also covers the styled components.

Thanks!

Sent from my iPhone



On Jun 14, 2016, at 15:59, Marco Wettstein <notifications@github.com> wrote:

Hi there, awesome work so far, looks really promising and clean!

I found one oddity, though: In all RadioField-Variants, the value is
displayed directly :

<label>{item}</label>

I think it should be: <label>{transform(item)}</label> with transform
beeing a prop passed to the field.

I would do a PR, but i wanted to ask, if I understand this right?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/vazco/uniforms/issues/24>, or mute the thread
<https://github.com/notifications/unsubscribe/AEbz3LIyuWqVoMyI0IClSKEmgQu-lcExks5qLqXHgaJpZM4I1Szi>
.

radekmie:
Yes, that's a good idea, @serkandurusoy. Looking forward your PR, @macrozone!

macrozone:
Ok here we go. 

I noticed that the published version does not include the transform-mechanism at all. Are there other tasks to do before the next release?

radekmie:
No, but currently releases are quite random - sorry.

radekmie:
I've just released `1.0.0-rc.16`.

macrozone:
awesome, thx guys!

PR #27, opened 2016-06-15
use also transform in radio-fields to display the label

macrozone:
fixes https://github.com/vazco/uniforms/issues/24


radekmie:
Good work! Thanks, @macrozone.

PR #36, opened 2016-06-17
Dev issue 35 label left

zeroasterisk:
The `modifyProps` function is just an idea, but easiest to communicate this way....

### DIFF
--- a/packages/uniforms-bootstrap3/src/components/fields/BoolField.js
+++ b/packages/uniforms-bootstrap3/src/components/fields/BoolField.js
@@ -5,11 +5,12 @@ import {connectField} from 'uniforms';
 import FormGroup from './FormGroup';
 
 // IDEA manipulate props for label on left: re #35
-const modifyProps = (props) {
+const modifyProps = (props) => {
     if (!props.label) return props;
-    if (!props.labelLeft) return Object.assign({}, props, { label: props.labelLeft });
-    return Object.assign({}, props, { label: '&nbsp;' });
-}
+    return Object.assign({}, props, {
+        label: (props.labelLeft ? props.labelLeft : <span>&nbsp;</span>)
+    });
+};
 
 const Bool = props =>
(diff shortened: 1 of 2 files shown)


radekmie:
I'm generally against such _wrappers_, so how about this:

```js
const Bool = ({label, labelBefore, ...props}) =>
    <FormGroup label={labelBefore} {...props}>
        <section className={classnames(props.inputClassName, `checkbox${props.inline ? '-inline' : ''}`)}>
            <label htmlFor={props.id}>
                <input
                    checked={props.value}
                    disabled={props.disabled}
                    id={props.id}
                    name={props.name}
                    onChange={() => props.onChange(!props.value)}
                    type="checkbox"
                />&nbsp;
                {label}
            </label>
        </section>
    </FormGroup>
;
```

It will pass your `labelBefore` (or as you already named it `labelLeft`) to the `<FormGroup />`, and leave it's current label,

packages/uniforms-bootstrap3/src/components/fields/BoolField.js
```diff
@@ -15,7 +15,7 @@ const Bool = props =>
                     name={props.name}
                     onChange={() => props.onChange(!props.value)}
                     type="checkbox"
-                />
+                />&nbsp;
                 {props.label}
```
radekmie: Now `props.label` is always `undefined` - change it to `label`.

packages/uniforms-bootstrap4/src/components/fields/BoolField.js
```diff
@@ -15,7 +15,7 @@ const Bool = props =>
                     name={props.name}
                     onChange={() => props.onChange(!props.value)}
                     type="checkbox"
-                />
+                />&nbsp;
                 {props.label}
```
radekmie: Same as above.

zeroasterisk:
updated... better?

radekmie:
Great!

Issue #35, opened 2016-06-17
bug: checkbox label doubled & needs spacing between checkbox and label


zeroasterisk:
We need to be able to control the checkbox field better.

* We should default to no label on the left (in the FormGroup)
* We should allow a schema defined custom label on the left (?prop: `labelLeft` ?)
* We should add a `&nbsp;` after the input and before the label on the right.

![ss](http://puu.sh/pw6uh/000ff35e05.png)

radekmie:
Okay, so:

0. _We should default to no label on the left (in the FormGroup)_
I'll comment #36.

0. _We should allow a schema defined custom label on the left (?prop: labelLeft ?)_
It's already done because we have `uniforms` key in SimpleSchema and it will be merged with props. In my opinion `labelLeft` is not adequate, because sometimes it is above the field. How about `labelBefore`?

0. _We should add a `&nbsp;` after the input and before the label on the right._
It's covered in #37, and surely will be merged.

PR #37, opened 2016-06-17
re #35 add space between checkbox/label

zeroasterisk:
Super simple - should be fine


radekmie:
As I understand, it's required in these themes, right?

PR #38, opened 2016-06-17
Feature react node for label

zeroasterisk:
Found a need for this, when playing w/ labels for checkboxes...

ss


radekmie:
Have you checked how `uniforms-semantic` and `uniforms-unstlyed` works with it?

zeroasterisk:
I put in the space in the other packages, but I have not tested them in each... my "examples" repo is no longer working, and I don't currently run the other formats.  I'm not 100% adding a space is a great idea or not... it might be better handled w/ CSS, but that's a "gotcha" which will mess up a lot of people.... your call.

I did split off the proptypes change to it's own PR #49 since that should be very simple to upstream. (In my case, I'm using an icon in a label)

radekmie:
I'm not saying, that you have to add that white space to all packages - I just wanted you to check, how it works (that React node `label`). I've checked it by myself - it works ok. You can remove those spaces, then I'll merge it.

radekmie:
Closing - it is already done.

PR #49, opened 2016-06-27
Add proptype support for JSX, react node, as label

zeroasterisk:
I split this off of PR #38 - so it was easier to merge (seems like a super-simple upstream)

(In my case, I'm using an icon in a label)


radekmie:
Great!

PR #50, opened 2016-06-28
Added Bootstrap docs

zeroasterisk:


radekmie:
You surprised me again! Looking forward you progress!

radekmie:
Documentation time, @zeroasterisk!

It took me some time, but [API documentation is not only a dream anymore](https://github.com/vazco/uniforms/blob/master/API.md). Right now, forms and helpers are fully documented, but fields not _(currently: 3)_. Also, bootstrap themes have a lot of styling props, so it would be great if you could write something about them.

I'm closing it, as it's quite outdated right now _(files, not documentation itself)_. I'd be glad if you could write some more in the current docs and/or review the already written.

PR #61, opened 2016-07-05
Allow defaultValue to be a function

RSchwan:


radekmie:
I'm okay with this, but [this may _break_ your schema](https://github.com/aldeed/meteor-simple-schema/blob/master/simple-schema.js#L464-L481). I would suggest using [`Object.defineProperty`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) to achieve such behaviour. Example:

```js
const Example = new SimpleSchema({
    field: Object.defineProperty({
        type: String
    }, 'defaultValue', {
        enumerable: true,
        get () {
            return 'Magic!'
        }
    })
});
```

I know, that it's a lot of boiler code, but, really - this change can _break_ your validation (`schema.clean` to be precise).

RSchwan:
`Object.defineProperty` doesn't really work. Per example random strings don't work. I decided to use `autoValue` which is also the recommended way. So need for this feature. Especially if it breaks stuff.

PR #67, opened 2016-07-12
SemanticUI: Add support for input icons

MacRusher:
Adds ability to use icons in `TextField` component.

Example usage:
```javascript
<TextField name="username" icon="user" />
<TextField name="query" iconLeft="search" />
```
There could be only one icon, and `icon` take precedence before `iconLeft`.

I could also add support for [pointer events](http://semantic-ui.com/elements/input.html#icon) if you think it would useful:
```javascript
<TextField name="username" icon="add user" onIconClick={() => {}} />
```

One problem I'm aware of is that adding `ui icon input` classes change the input width.
This could be solved by adding `fluid` class either by `TextField` itself (IMHO bad idea) or end user:
```javascript
<TextField name="username" className="fluid" icon="user" />
```

### DIFF
--- a/packages/uniforms-semantic/src/components/fields/DateField.js
+++ b/packages/uniforms-semantic/src/components/fields/DateField.js
@@ -23,6 +23,9 @@ const Date_ = ({
     findError,    // eslint-disable-line no-unused-vars
     findField,    // eslint-disable-line no-unused-vars
     findValue,    // eslint-disable-line no-unused-vars
+    icon,
+    iconLeft,
+    iconProps,
     id,
     label,
     max,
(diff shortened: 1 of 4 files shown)


MacRusher:
Few changes.

Per @radekmie recommendation I've introduced new section for the input which may or may not contain icon classes. So the problem with different input sizes is gone now.

Instead of `onIconClick` etc. I've also introduced `icomProps` that are passed directly to `<i/>` element.
Example usage:
```javascript
<TextField name="username" icon="add user link" iconProps={{ onClick: () => {} }} />
```

radekmie:
Great! Would you like to do the same for `DateField` and `NumField` also?

MacRusher:
Sure, why not. I'll take care of this tomorrow.

MacRusher:
I've added icon support for `DateField` and `NumField`

radekmie:
Great!

PR #84, opened 2016-07-25
An idea for `showInlineError` Boolean

zeroasterisk:
If true, we could include the error message as a text block with the input form.

This seems like a very useful option,
especially for forms where the errors are not rendered.

### DIFF
--- a/packages/uniforms-bootstrap4/src/components/fields/FormGroup.js
+++ b/packages/uniforms-bootstrap4/src/components/fields/FormGroup.js
@@ -17,6 +17,7 @@ const FormGroup = ({
     className,         // class name for the whole .form-group
     disabled,          // boolean, if true, show fields as disabled
     error,             // error validation response
+    errorMessage,      // errorMessage string of error
     grid,              // grid is either a int [1-11] or object {xs:6,sm:4,md:2}
     help,              // help text
     helpClassName,     // class name for the help text (default: 'text-muted')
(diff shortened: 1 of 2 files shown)


radekmie:
Yes, inline errors are useful.

packages/uniforms-bootstrap4/src/components/fields/FormGroup.js
```diff
@@ -41,11 +42,13 @@ const FormGroup = ({
             <section className={classnames(wrapClassName, gridClassName(grid, 'input'))}>
                 {children}
                 {makeHelp(help, helpClassName)}
+                {error && showError ? makeHelp(error, 'text-help') : ''}
```
radekmie: Why `makeHelp(error, ...)` and not `errorMessage`?

packages/uniforms-bootstrap4/src/components/fields/FormGroup.js
```diff
@@ -41,11 +42,13 @@ const FormGroup = ({
...
             </section>
         )}
 
         {!grid && !wrapClassName && children}
         {!grid && !wrapClassName && makeHelp(help, helpClassName)}
+        {!grid && !wrapClassName && error && showError ? makeHelp(error, 'text-help') : ''}
```
radekmie: Same as above.

radekmie:
Two more things: how about renaming it to `inlineError` or `showInlineError`?

---

Could you also implement it for bootstrap-3?

zeroasterisk:
Sure - I was just making sure you were on-board... I'll whip it up

packages/uniforms-bootstrap4/src/components/fields/FormGroup.js
zeroasterisk: It needs to be wrapped in `<span.text-help>`

radekmie: Yes, I get it, but `error` is an object, not a `String` - `errorMessage` is a `String`.

zeroasterisk:
@radekmie ok - bootstrap 3 & 4 are implemented, `showInlineError` will render the inline spans as help blocks/texts, as a sibling to any existing help blocks/texts

radekmie:
Great!

Issue #64, opened 2016-07-06
React "Unknown props" warnings


RSchwan:
I'm not sure if this is intended or not but I get a bunch of "Unknown props" errors.
Per example:
```
Warning: Unknown props `model`, `schema`, `validate` on <form> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in form
    in AutoValidatedQuickSemanticForm (created by CreateDivisionPage)
    in div (created by CreateDivisionPage)
    in div (created by Modal)
    in div (created by Modal)
    in div (created by Modal)
    in Modal (created by CreateDivisionPage)
    in CreateDivisionPage (created by RouterContext)
    in div (created by AppLayout)
```

radekmie:
It's a new feature in `React@15.2.0` - it's on my roadmap, but it will take a long time to filter **all** props. I'll do it step by step, but every help is appreciated.

RSchwan:
I could help. I just need to know how so that everything is done the same way.

radekmie:
Okay - I'll do it in one of the packages.

serkandurusoy:
@radekmie this is the list of standard dom properties that are supported by react

https://github.com/facebook/react/blob/master/src/renderers/dom/shared/HTMLDOMPropertyConfig.js#L24

Do you think we can create a wrapper around this that shares a props object to two different objects, one with the allowed standard ones and other with the remaining, custom ones.

this way, we can pass on the second one as props and use the first one where appropriate

serkandurusoy:
For reference, here's a proper `Array`:

```js
const ALLOWED_STANDARD_HTML_PROPERTIES = [
  "accept",
  "acceptCharset",
  "accessKey",
  "action",
  "allowFullScreen",
  "allowTransparency",
  "alt",
  "async",
  "autoComplete",
  "autoPlay",
  "capture",
  "cellPadding",
  "cellSpacing",
  "charSet",
  "challenge",
  "checked",
  "cite",
  "classID",
  "className",
  "cols",
  "colSpan",
  "content",
  "contentEditable",
  "contextMenu",
  "controls",
  "coords",
  "crossOrigin",
  "data",
  "dateTime",
  "default",
  "defer",
  "dir",
  "disabled",
  "download",
  "draggable",
  "encType",
  "form",
  "formAction",
  "formEncType",
  "formMethod",
  "formNoValidate",
  "formTarget",
  "frameBorder",
  "headers",
  "height",
  "hidden",
  "high",
[...]

PR #89, opened 2016-08-04
Removed `...props` from SubmitField section

zeroasterisk:
> Warning: Unknown prop `grid` on `section` tag.
> Remove this prop from the element

### DIFF
--- a/packages/uniforms-bootstrap3/src/components/fields/ErrorsField.js
+++ b/packages/uniforms-bootstrap3/src/components/fields/ErrorsField.js
@@ -5,7 +5,7 @@ import {nothing}   from 'uniforms';
 
 const ErrorsField = ({className, children, ...props}, {uniforms: {error, schema}}) =>
     (!error && !children) ? nothing : (
-        <section className={classnames('panel panel-danger', className)} {...props}>
+        <section className={classnames('panel panel-danger', className)} {...filterDOMProps(props)}>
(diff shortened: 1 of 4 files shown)


radekmie:
Since #64 (see [this](https://github.com/vazco/uniforms/issues/64#issuecomment-233732854) comment), there's a `filterDOMProps` helper. [Example](https://github.com/vazco/uniforms/blob/master/packages/uniforms-semantic/src/components/fields/BoolField.js#L19).

---

I'll add it to README later.

zeroasterisk:
updated, thanks!

packages/uniforms-bootstrap3/src/components/fields/ErrorsField.js
```diff
@@ -5,7 +5,7 @@ import {nothing}   from 'uniforms';
 
 const ErrorsField = ({className, children, ...props}, {uniforms: {error, schema}}) =>
     (!error && !children) ? nothing : (
-        <section className={classnames('panel panel-danger', className)} {...props}>
+        <section className={classnames('panel panel-danger', className)} {...filterDOMProps(props)}>
```
radekmie: You haven't imported this helper - ESlint fails.

radekmie:
I've continued your PR and done it also in `semantic` and `unstyled` packages. Thanks for pointing it out!

PR #107, opened 2016-09-28
Enable SimpleSchema2Bridge with simpl-schema package

logankoester:
Some fixes required to get the new SimpleSchema2 bridge working after updating to [v1.1.1](https://github.com/vazco/uniforms/tree/v1.1.1)

### DIFF
--- a/packages/uniforms/package.json
+++ b/packages/uniforms/package.json
@@ -100,6 +100,7 @@
     "lodash.get": "^4.4.2",
     "lodash.isequal": "^4.4.0",
     "lodash.set": "^4.3.2",
-    "lodash.xorwith": "^4.5.0"
+    "lodash.xorwith": "^4.5.0",
+    "simpl-schema": "0.0.3"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/uniforms/package.json
```diff
@@ -100,6 +100,7 @@
     "lodash.get": "^4.4.2",
     "lodash.isequal": "^4.4.0",
     "lodash.set": "^4.3.2",
-    "lodash.xorwith": "^4.5.0"
+    "lodash.xorwith": "^4.5.0",
+    "simpl-schema": "0.0.3"
```
radekmie: Adding `smpl-schema` as a dependency will require you to install it. We can _optionally_ mark it as a peer dependency.

packages/uniforms/src/bridges/SimpleSchema2Bridge.js
```diff
@@ -31,8 +31,7 @@ try {
     );
 } catch (_) { /* Ignore it. */ }
 
-// export default class SimpleSchema2Bridge extends Bridge {
-export class SimpleSchema2Bridge extends Bridge {
+export default class SimpleSchema2Bridge extends Bridge {
```
radekmie: Yeah, that's a terrible typo.

_(publish before going to bed is generally a bad idea)_

radekmie:
I've merged it, changed `simpl-schema` to a peer dependency and released 1.1.2.

Thank you, @logankoester!

PR #129, opened 2016-11-07
export wrapField lib

zeroasterisk:
I like what you did with the `wrapField` and the inline errors.

Also your warning for the old `FormGroup` was very handy.

But now my "custom" components need to be updated.... and I want access to replicate what's happening inside your components with `wrapField`


radekmie:
Sorry for the long delay, but I was thinking about it whole day... I'm still not sure about it. For now, I suggest using `import wrapField from 'uniforms-bootstrap3/lib/lib/wrapField';`.

I'll let you know in the closest future.

zeroasterisk:
Sure - I can deal w/ that - I was just thinking that if components were extensions of the framework, tools from within the framework should (maybe) be exposed...

radekmie:
Yes, but I treat `wrapField` as an _internal_ helper - if we make it public, then we cannot remove it _(or do something critical)_, because it's not backward compatible.

todda00:
I too have the same situation, I just started refactoring to use wrapField instead of FormGroup for my custom components, but want to make sure I am building the components in an expected way for this package.

radekmie:
Currently, `import wrapField from 'uniforms-bootstrap3/lib/lib/wrapField';` is the preferred way.

radekmie:
As I've finally managed some free time I've found a solution - flat build. There are no duplicates within packages, so we can make it like this `import * from 'uniforms/*'`, ex. `import wrapField from 'uniforms-bootstrap/wrapField'`. What do you think @zeroasterisk, @todda00?

todda00:
Work for me! bitmoji

zeroasterisk:
Perfect

radekmie:
It will ship in `v1.8.0`.

PR #134, opened 2016-11-13
Ant Design Theme

Gervwyk:
The basic components for antd working. 
Not sure you want to pull all the changes in the demo project and I have not had time to write/adjust the testing since this is new to me, will spend time learning this at a later stage. 
Take note. This is also my first pull request/github contribution of this sort so please advise if I'm not following best practices.  Is it better to complete all the work for enhancements before creating a pull request or is it ok to make pull requests as the project evolves?


radekmie:
At first, make the ESLint and tests pass - `npm test` in the root of the repo. I'll give you more remarks later.

Gervwyk:
Fixed all the syntax fails for testing. Got stuck on:

`Error: matchMedia not present, legacy browsers require a polyfill`

Will look into this at a later stage.

radekmie:
I've never used ANT Design before, but WickyNilliams/enquire.js#82 seems pretty similar.

Gervwyk:
So I have polished the antd module quite a bit, test still failing but I think this is related to antd so will have to investigate further. I'm also going to create a few unusual components for my apps so I will push these to the 'uniforms-antd-extended' (components such as a color picker, rating field, swap list items, geofield, there is a list which I need to complete and I will not have time to polish them as much as the standard ones). 

So I will keep the 'uniforms-antd' branch as clean and similar to the others as possible. I'm new to testing, should I make any changes to the test/combined.js files specificly for antd? Is that something I should look into or should the tests which I copied from semantic work?

radekmie:
Your tests should match your theme. Simplest way is checking the coverage.

radekmie:
Hey @Gervwyk! How is this going? Do you need any help?

Gervwyk:
Hi @radekmie . Sorry, had few deadlines to meet for Dec. Will be back on this in Jan. When I last checked I only had to finish writing the tests for this, however got stuck.. So I'll finish that soon. I like the checklist you put up on the Material pull request, I'll do the same here to keep track of what needs to be finished. 

- [ ] check tests

radekmie:
It's good to hear that, @Gevryk! OK, so see you in January. And remember, that I'll surely help you!

Gervwyk:
Thanks @radekmie . I updated the repo. I'm already using uniforms-antd in production for a while now and I'm very happy with where it is currently. For now when I run the demo, the popup elements such as date selector and select causes problems related to  react-panelgroup package.. When I first worked on it I changed the demo not to work with this package and I developed the modules using the demo project, however with the  react-panelgroup version these popups does not render the antd styles correctly. If you can possibly investigate this would be much appreciated. 

I think besides this we only need to write the test cases and update docs. There is also a few schema related point we need to discuss.

radekmie:
I'll review it after Christmas, @Gervwyk.
About that popups - I'll surely try to fix it, because we have the same problem with material ui theme.

radekmie:
@Gervwyk: Will you have some time to finish this? And about these extra components - I saw your `uniforms-antd` and `uniforms-antd-extended` packages on npm - it's a good idea to handle it this way.

Gervwyk:
Hi @radekmie. Yes I want to push to finish this asap. I'm just all over the place and don't get a lot of time for coding currently, but this will settle in Jan. 

On your review changes solved.
Solved:
- I removed the unnecessary packages
- Sorted the eslint and import markup
- Removed the improper schema examples
- Removed non-standard fields (`ColorPickerField`) from master branch

Unsolved.
- The inline styles is mainly used to refine the layout, and spacing. What would be a better alternative?
- `fieldComponent` is used to reference to a component via a reference handel, packaged in the module. I'm not familiar with schemas other than simple-schema so not sure how this would apply, but my use case is: If I want to use `HiddenField` or `ColorPickerField` in the schema definition, those fields must be imported to be applied to my schema definition. The next step for my project is to save the schema definitions in my database and instead of decoding schema definitions in the app to a saveable format, I choose to define string reference handels for these component and thus makes a simpler schema definition. In future versions of the extended fields, more many fields will be available since I'm also using uniforms in a survey tool, where various question types will refer to different fields. Thus creating a string reference for the field type made sense at the time. I'm open to change this if there is a better solution since I feel the string reference handel is not modular enough for the application especially considering approaches to `uniforms-antd-extended` packages - which should contain the non-standard fields.

On the topic of adding extra field types. How do we want to approach this? On giving this some more though, I feel that each field type should be installed as an module by itself.. such as `uniforms-antd-colorpicker` or `uniforms-antd-question-grid` or `uniforms-antd-survey-fields` so that users can add additional field types each packaged on their own. Not sure how much this makes sense in terms of maintainability. Thus `uniforms-antd` then contains the standard fields, and users can then install additional fields. I guess importing and referencing the field type in the app makes more sense in this module than my string reference handels for fields, since any field from any package can then be imported and reference in the schema. Then I just need to find a better way to store my schema definitions in the database. 

Thus the following recommendations/questions:
- I'm going to remove the `fieldComponent ` references and revert to the import method to maintain modularity.
- Does it makes sense to then package the extra field independently from the base components, for instance a package  `uniforms-antd-colorpicker` containing  only the color picker field?
- Any advice on how to save a schema definition in the database without more complex encoding/decoding function for type matching?

To complete this pull request and publish `uniforms-antd` I think I only have to:
- Write the tests 
- Fix the `react-panelgroup` in the demo project
- Amend the documentation 
Is this all or am I missing something?

radekmie:
I'm really glad to see how the development is going, @Gervwyk!

About these inline styles - it's not a problem.
About `fieldComponent` - read [this](https://github.com/vazco/uniforms#example-customautofield) _(this is also an answer for storing schemas in DB - just do it like you are doing it right now, but in your custom `<AutoField /.`)_.
About additional fields - one package per theme, like `uniforms-antd-extended` for `uniforms-antd`, is OK. It's not hard to keep it in sync with other versions _(thanks to Lerna)_ and it won't cause a bloat if you are using direct imports, like `import X from 'uniforms-theme-extended/X`.
About these _one-field-packages_ - I really appreciate the UNIX philosophy _Make each program do one thing well._, but let's keep it simple - each theme contains basic fields, needed to render any schema and another one to contain all additional fields, needed to have a nice UX _(like a colour picker or a scheduler)_.

Things that have to be done before merge:
* documentation - both `API` and `README`
* tests
* further _(rather a style)_ remarks - I'll review it today or tomorrow
* _(and finally)_ allow me to publish `uniforms-antd` and `uniforms-antd-extended`

About the demo - I've tried to solve it and it's not that easy... I'll try again and if I fail, I'll probably switch sides - it should be enough.

---

One more thing:
> Imports should be always on the top level.

This applies also to `require`s in the components - I won't accept it. Use `import`s instead and put them at the top of files.

Gervwyk:
> This applies also to requires in the components - I won't accept it. Use imports instead and put them at the top of files.

I had the antd components as imports but on ssr it caused problems related to antd. However, using `require` worked well on both ssr and client. I'll try to run the ssr with imports again as this issue might have been fixed in later versions of antd.

radekmie:
That's a big problem if it won't work... OK, I'll wait.

demo/lib/styles.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 
 const style = links => links.map((link, index) => <link key={index} rel="stylesheet" href={link} />);
 const styles = {
+    antd: style([
+        'https://cdnjs.cloudflare.com/ajax/libs/antd/2.3.2/antd.min.css'
```
benjycui: OK, it's just a demo, I make a mistake..

packages/uniforms-antd/src/DateField.js
```diff
@@ -0,0 +1,99 @@
...
+                info
+              }
+            } = this;
+        const AntD = require('antd');
+        const DatePicker = AntD.DatePicker;
+        const enUS = require('antd/lib/date-picker/locale/en_US');
+        const LocaleProvider = AntD.LocaleProvider;
+        return (
+            <FormGroup errorMessage={errorMessage} id={id} label={label} showInlineError={showInlineError} info={info} >
+                <LocaleProvider locale={enUS}>
```
benjycui: I think `LocaleProvider` should not be put in uniform field, just let developer use `LocaleProvide` to wrap their application.

packages/uniforms-antd/src/ListAddField.js
```diff
@@ -0,0 +1,23 @@
+import connectField     from 'uniforms/connectField';
+import React            from 'react';
+
+const ListAdd = ({
+    disabled,
+    parent,
+    value
+}) => {
+    const limitNotReached = !disabled && !(parent.maxCount <= parent.value.length);
+    const AntIn = require('antd');
```
benjycui: Actually, it's not a good ides to import the whole antd at once..

demo/lib/styles.js
radekmie: But still - it surely have to be updated before merge.

packages/uniforms-antd/src/DateField.js
radekmie: 👍

packages/uniforms-antd/src/ListAddField.js
radekmie: Also, there's a typo.

radekmie:
@benjycui: I'm really glad to see you and your support here!

radekmie:
Hey @Gervwyk, how is it going? Do you need some help?

Gervwyk:
I pulled the refactored work and all should now be up to date. Just the test and docs to go. @radekmie will you please look at the test when you find time. In december when I worked on it I got stuck with the Checkbox.Group giving errors when values was set as non-string (numbers) types.

packages/uniforms-antd/src/BaseForm.js
```diff
@@ -1,9 +1,22 @@
...
+            fieldData: PropTypes.object
+    };
+
     static displayName = `AntD${parent.displayName}`;
+
+    getChildContextState () {
+        return {
+            ...super.getChildContextState(),
+            fieldData: this.props.fieldData
+        };
```
radekmie: I guess it's here because of #137. I'd rather move it to the `BaseForm`, but not in this PR. And We have to think out a better name.

Gervwyk: Was also thinking that. Agreed.

Issue #140, opened 2016-11-28
reset should set initial model


macrozone:
I tried reset() for the first time.

It added a button to the QuickForm and called this.reset();.

It thought it would reset the state to the given props.model on the form, but instead, all fields were emptied. Shouldn't it reset to the model instead? if no model is given, then yes, it should empty all fields.

Edit: 

it's on this line:

https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/components/forms/AutoForm.js#L65

i think it should set it back to this.props.model instead of {},

Setting it back to props.model would be more logical to me, but this would be a breaking change. Opinions?

radekmie:
So you've tried it on `AutoForm` - there's no internal `model` in `QuickForm` :)

Yes, it is a breaking change, but every bugfix is a breaking change because something is working differently now, doesn't it? This should behave exactly the same as the initial state in the `constructor`, so as you've said - `this.props.model` in both `model` and `modelSync`.

Would you like to submit a PR for it?

radekmie:
@macrozone?

macrozone:
@radekmie: Yes, I can do this in the next days. But generally, you do agree with this change?

radekmie:
Yes, I do agree. If you are using it in a fully automatic way _(you have no model)_, then we are not changing anything. If you are using it as a _default_ model, then the new behaviour is the desired one. At last, if you are using it as a managed form, then it's still OK because it'll keep the model from props.

radekmie:
Available in `v1.8.0`.

PR #145, opened 2016-12-05
reset sets model back to initial model

macrozone:
fixes https://github.com/vazco/uniforms/issues/140

however, something is odd:

In the demo, the form already contains default-values. not from the mode, but from the bridge/schema.

reset does not seem to restore these default-values and i am unsure whether this should be the case.

### DIFF
--- a/packages/uniforms/src/components/forms/AutoForm.js
+++ b/packages/uniforms/src/components/forms/AutoForm.js
@@ -62,7 +62,7 @@ const Auto = parent => class extends parent {
 
     onReset () {
         super.onReset();
-        this.setState(() => ({model: {}, modelSync: {}}));
+        this.setState(() => ({model: this.props.model, modelSync: this.props.model}));
     }
 
     onValidate () {


radekmie:
Let me guess - you haven't noticed [this](https://github.com/vazco/uniforms/blob/master/demo/components/Application.js#L16)?

radekmie:
Also, which form in the docs? The main one and the one on the left are managed.

radekmie:
@macrozone

macrozone:
@radekmie: the one on the right and yes, I haven't noticed that, but that cleans the schema for the form on the right.

The form on the right does not restore defaults, when you call reset, but it initially contains the default values. I would think that a reset should also restore defaults from the schema.

radekmie:
@macrozone OK, now I understand. Look at [these lines](https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/components/forms/AutoForm.js#L63-L66). At first, we call [`BaseForm#reset`](https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/components/forms/BaseForm.js#L243-L250), which will rerender the whole form, including calculation of default values. To fix it, we have to switch those lines in `AutoForm` or even call `super.onSubmit()` in the `setState` callback.

radekmie:
@macrozone Will you test it, or should I do it by myself?

PR #164, opened 2017-01-14
Fix initialValue check

DimitryDushkin:
Currently there is no way set default value for SelectField, because its value by default is `''` (empty string) which is not equal to `undefined`.

### DIFF
--- a/packages/uniforms/src/connectField.js
+++ b/packages/uniforms/src/connectField.js
@@ -44,7 +44,10 @@ export default function connectField (component, {
                     return;
                 }
 
-                if (props.value === undefined && props.required) {
+                if (
+                    props.required
+                    && (this.options.ensureValue ? props.value === '' : props.value === undefined)
+                ) {
                     props.onChange(props.initialValue);
                 }
             }


radekmie:
Could make a reproduction? Or at least share schema and/or used form?

packages/uniforms/src/connectField.js
```diff
@@ -44,7 +44,7 @@ export default function connectField (component, {
                     return;
                 }
 
-                if (props.value === undefined && props.required) {
+                if (!props.value && props.required) {
```
radekmie: This should be changed - if a field has a falsy value, like `false` and `initialValue` equal to `true`, it will cause a problem. Empty string is caused by `ensureValue: true` - this flag is used to avoid `undefined` in HTML inputs.

I'd suggest
```diff
- if (props.value === undefined && props.required) {
+ if (props.required && (this.options.ensureValue ? props.value === '' : props.value === undefined)) {
```

DimitryDushkin:
I'm using graphql, so there is no default value. Component looks like following:
```js
<AutoForm
    ref={ ref => formRef = ref }
    schema={ bridge }
    onSubmit={ this.onSubmit }>
    <SelectField
        name='adapterset'
        allowedValues={ adaptersetOptions }
        transform={ getEnumOption.bind(this, sampleEnums.adapterset) }
        initialValue='ILLUMINA_GENERIC'
        required={ true } />
</AutoForm>
```

Schema:
```graphql
        input SampleInput {
            adapterset: String!
        }
```

Your suggestion looks good, I'll rewrite commit.

radekmie:
1. OK.
2. Wait, what?
3. I've never thought about passing `initialValue` to the component!
4. To be honest, that's a very good idea!
5. I'm really glad to see, that you are using GraphQL with uniforms!
6. If you are using GraphQL schema, then you can use third parameter of GraphQLBridge (example [here](https://github.com/vazco/uniforms#graphql-definition)).

DimitryDushkin:
6. Oh, I missed that option( I think this better then inlining `allowedValues` and such in components (separation of concerns and such 😃 ).

I've rewrote commit.

DimitryDushkin:
BTW graphql intergation works quite good with form! Thank you for this.) 
The only "problem" I've faced — warning from webpack of not presence dependency of `simpl-schema` module. I'll try to send PR on this someday.)

radekmie:
I've tried to get rid of it and failed few times so far. If you could handle it or help to do so - I'd love to help.

PR #169, opened 2017-01-23
Num field state

### LINKED ISSUE #167: Safari numeric input not allowing decimals
Strange issue here with Safari and numeric inputs.

In Safari when adding a numeric field using AutoField, the decimal cannot be typed unless its at the beginning of the input. So I cannot type 1.5, but could type .5. but not 0.5 . Oddly, you can type 15 then cursor back and enter a decimal to get to 1.5 . Also, the arrows up and down do work, so if you enter 1, and arrow up, 1.01 is entered (my step is 0.01). Initially I thought this was a Safari oddity with type="number" fields, but I pasted the exact HTML generated from uniforms and added it directly, and Safari allows me to insert decimals.

Schema:
```
UTR: {
    type: Number,
    optional: true,
    min: 1,
    max: 17,
    decimal: true,
    label: 'UTR',
  },
```

Usage:
```
<AutoField name="UTR" grid={{ xs: 6 }} inputClassName="form-control-sm" />
```

This generates the HTML:
```
<input type="number" class="form-control-sm form-control" id="uniforms-0000-0001" max="17" min="1" name="UTR" placeholder="" step="0.01">
```

When I add this HTML directly in the code outside of uniforms it works as expected, but not from the AutoField generated code.
safari_decimal

Chrome and Firefox do not have this issue, decimals can be entered as expected

todda00:
Add internal state to NumField for BS3 & BS4, allows trailing decimals to be added in Safari. Resolves #167


radekmie:
I wanted to provide a general solution, rather than fixing it in this very case. Also, your solution _(unfortunately)_ won't work if your model will change _(`componentWillReceiveProps` Is the answer)_. I'll respond you later, OK?

radekmie:
Additionally, tests failed.

radekmie:
It's fixed in `v1.11.0`.

PR #170, opened 2017-01-24
bootstrap 4 alpha.6 changed classes

zeroasterisk:
Bootstrap has changed a few classes to support flexbox

    form-control-label  --> col-form-label (required)
    col-{size}-{number} --> col-{number}  (optional)

I have maintained backwards compatibility
But should eventually clean it up and remove later.

Bootstrap has changed a few classes on check/radio 

    radio & radio-inline --> form-check
    <label>              --> form-check-label
    <input>              --> form-check-input

    checkbox & checkbox-inline --> form-check
    <label>                    --> form-check-label
    <input>                    --> form-check-input

### DIFF
--- a/packages/uniforms-bootstrap4/src/BoolField.js
+++ b/packages/uniforms-bootstrap4/src/BoolField.js
@@ -6,10 +6,17 @@ import wrapField from './wrapField';
 
 const Bool = ({label, labelBefore, ...props}) =>
     wrapField({label: labelBefore, ...props}, (
-        <section className={classnames(props.inputClassName, `checkbox${props.inline ? '-inline' : ''}`)}>
-            <label htmlFor={props.id}>
+        <section
+            className={classnames(
+                props.inputClassName,
+                'form-check',
+                `checkbox${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+            )}
+        >
+            <label htmlFor={props.id} className="form-check-label">
                 <input
                     checked={props.value}
+                    className="form-check-input"
                     disabled={props.disabled}
                     id={props.id}
                     name={props.name}
--- a/packages/uniforms-bootstrap4/src/RadioField.js
+++ b/packages/uniforms-bootstrap4/src/RadioField.js
@@ -7,10 +7,18 @@ import wrapField from './wrapField';
 const Radio = props =>
     wrapField(props, (
         props.allowedValues.map(item =>
-            <section key={item} className={classnames(props.inputClassName, `radio${props.inline ? '-inline' : ''}`)}>
-                <label htmlFor={`${props.id}-${item}`}>
+            <section
+                key={item}
+                className={classnames(
+                    props.inputClassName,
+                    'form-check',
+                    `radio${props.inline ? '-inline' : ''}` // bootstrap4 < alpha.6
+                )}
+            >
+                <label htmlFor={`${props.id}-${item}`} className="form-check-label">
                     <input
                         checked={item === props.value}
+                        className="form-check-input"
                         disabled={props.disabled}
                         id={`${props.id}-${item}`}
                         name={props.name}
--- a/packages/uniforms-bootstrap4/src/wrapField.js
+++ b/packages/uniforms-bootstrap4/src/wrapField.js
@@ -45,7 +45,14 @@ export default function wrapField ({
             )}
         >
             {label && (
-                <label htmlFor={id} className={classnames('form-control-label', gridClassName(grid, 'label'))}>
+                <label
+                    htmlFor={id}
+                    className={classnames(
+                        'form-control-label', // bootstrap4 < alpha6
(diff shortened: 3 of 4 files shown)


zeroasterisk:
NOTE: there are also changes to radios and checkboxes we need to look into:

http://v4-alpha.getbootstrap.com/components/forms/#using-the-grid

radekmie:
Thanks!
1. Why have you changed `section`s to `div`s?
2. You have to pass ESLint and tests.

zeroasterisk:
@radekmie I switched from `section` --> `div` because of this

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section
> Do not use the `<section>` element as a generic container; this is what `<div>` is for, especially when the sectioning is only for styling purposes. A rule of thumb is that a section should logically appear in the outline of a document.

radekmie:
Yeah... How about _every_ `section` in all packages _(in a different PR of course)_ and let this one be smaller?

zeroasterisk:
I rebased and force-pushed to a new history....

packages/uniforms-bootstrap4/src/BoolField.js
```diff
@@ -6,9 +6,19 @@ import wrapField from './wrapField';
...
+            className={classnames(
+                props.inputClassName,
+                'form-check',
+                `checkbox${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+            )}
+        >
+            <label
+                htmlFor={props.id}
+                className="form-check-label"
+            >
```
radekmie: This will fit in one line.

packages/uniforms-bootstrap4/src/BoolField.js
```diff
@@ -6,9 +6,19 @@ import wrapField from './wrapField';
...
                 <input
+                    className="form-check-input"
                     checked={props.value}
                     disabled={props.disabled}
```
radekmie: Order.

_(I have to add it to the ESLint...)_

packages/uniforms-bootstrap4/src/RadioField.js
```diff
@@ -7,9 +7,20 @@ import wrapField from './wrapField';
...
+                className={classnames(
+                    props.inputClassName,
+                    'form-check',
+                    `radio${props.inline ? '-inline' : ''}` // bootstrap4 < alpha.6
+                )}
+            >
+                <label
+                    htmlFor={`${props.id}-${item}`}
+                    className="form-check-label"
+                >
```
radekmie: This will fit in one line too.

packages/uniforms-bootstrap4/src/RadioField.js
```diff
@@ -7,9 +7,20 @@ import wrapField from './wrapField';
...
                     <input
+                        className="form-check-input"
                         checked={item === props.value}
                         disabled={props.disabled}
```
radekmie: Order.

packages/uniforms-bootstrap4/src/wrapField.js
```diff
@@ -20,13 +20,17 @@ export default function wrapField ({
     const hasWrap = !!(grid || wrapClassName);
 
     const blockError = !!(errorMessage && showInlineError) && (
-        <span className="form-text text-danger">
+        <span
+            className="form-text text-danger"
+        >
```
radekmie: And here.

packages/uniforms-bootstrap4/src/wrapField.js
```diff
@@ -20,13 +20,17 @@ export default function wrapField ({
...
             {errorMessage}
         </span>
     );
 
     const blockHelp = !!help && (
-        <span className={classnames('form-text', helpClassName || 'text-muted')}>
+        <span
+            className={classnames('form-text', helpClassName || 'text-muted')}
+        >
```
radekmie: And here.

packages/uniforms-bootstrap4/src/wrapField.js
```diff
@@ -45,13 +49,22 @@ export default function wrapField ({
...
+                >
                     {label}
                 </label>
             )}
 
             {hasWrap && (
-                <section className={classnames(wrapClassName, gridClassName(grid, 'input'))}>
+                <section
+                    className={classnames(wrapClassName, gridClassName(grid, 'input'))}
+                >
```
radekmie: And finally here.

packages/uniforms-bootstrap4/src/BoolField.js
zeroasterisk: eslint was what made me multi-line it --- that was what failed linting :(

zeroasterisk:
I rebased and force-pushed to a new history.... after reverting to not-angry eslint pattern... 

does it look ok now?

(also - thanks for keeping the quality super-high, you rule!)

PR #172, opened 2017-01-24
Section to div

zeroasterisk:
Switched from section --> div in ALL components

> Do not use the <section> element as a generic container; this is what <div> is for, especially when the sectioning is only for styling purposes. A rule of thumb is that a section should logically appear in the outline of a document.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

### DIFF
--- a/packages/uniforms-bootstrap3/src/AutoFields.js
+++ b/packages/uniforms-bootstrap3/src/AutoFields.js
@@ -25,7 +25,7 @@ AutoFields.propTypes = {
 
 AutoFields.defaultProps = {
     autoField: AutoField,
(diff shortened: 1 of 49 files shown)


radekmie:
OK, I will merge it after #170.

radekmie:
@zeroasterisk, there are conflicts, due to #170 - I'll merge it after you'll solve them.

radekmie:
@zeroasterisk will you have some time to do this?

zeroasterisk:
Ok - I re-ran this `sed` after rebasing from master -- should be ok now.

Issue #183, opened 2017-02-05
GraphQLBridge mishandles initialValue.


yourse1f-yourorg:
Array and Object initial values are suppressed.  I don't know if that's correct.

Primitive initial values should be passed through to set field initial values, but undefined is returned.

radekmie:
These are not being suppressed, but not handled at all.

PR #184, opened 2017-02-05
Fix issue #183

dude-awap:
... as requested!

My understanding of Uniforms and React is not deep.  So, more work may be required for this PR to be production ready.

### DIFF
--- a/demo/lib/presets.js
+++ b/demo/lib/presets.js
@@ -81,10 +81,10 @@ const presets = {
                 throw {details};
             }
         }, {
-            city:   {label: 'City'},
-            state:  {label: 'State'},
-            street: {label: 'Street'},
-            zip:    {label: 'Zip'}
+            city:   {label: 'City', initialValue: 'Tucson'},
+            state:  {label: 'State', initialValue: 'AZ'},
+            street: {label: 'Street', initialValue: 'Highway 61'},
+            zip:    {label: 'Zip', initialValue: '63123'}
         })
     `,
 
--- a/demo/lib/presets.js
+++ b/demo/lib/presets.js
@@ -81,10 +81,10 @@ const presets = {
                 throw {details};
             }
         }, {
-            city:   {label: 'City', initialValue: 'Tucson'},
-            state:  {label: 'State', initialValue: 'AZ'},
-            street: {label: 'Street', initialValue: 'Highway 61'},
-            zip:    {label: 'Zip', initialValue: '63123'}
+            city:   {label: 'City'},
+            state:  {label: 'State'},
+            street: {label: 'Street'},
+            zip:    {label: 'Zip'}
         })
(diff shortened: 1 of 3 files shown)


demo/lib/presets.js
```diff
@@ -81,10 +81,10 @@ const presets = {
...
             }
         }, {
-            city:   {label: 'City'},
-            state:  {label: 'State'},
-            street: {label: 'Street'},
-            zip:    {label: 'Zip'}
+            city:   {label: 'City', initialValue: 'Tucson'},
+            state:  {label: 'State', initialValue: 'AZ'},
+            street: {label: 'Street', initialValue: 'Highway 61'},
+            zip:    {label: 'Zip', initialValue: '63123'}
```
radekmie: I'd rather keep these without the `initialValue`, to have exactly the same schemas in the demo.

radekmie: I'd rather keep these without the `initialValue`, to have exactly the same schemas in the demo.

dude-awap: I would have thought that providing, some but not all initial values would be appropriate for a demo, no?  How about just : 

           street: {label: 'Street', initialValue: 'Pre-initialized St.'},

radekmie: OK, but let's do it in another issue or PR and in all schemas.

dude-awap:
Corrected with latest commit.

packages/uniforms/test/GraphQLBridge.js
```diff
@@ -36,22 +36,27 @@ describe('GraphQLBridge', () => {
 
     const schemaData = {
         author: {
+            initialValue: 'Jobs',
```
radekmie: Now I've noticed, that this is an invalid value because it's not a valid object of type `Author`.

packages/uniforms/test/GraphQLBridge.js
```diff
@@ -36,22 +36,27 @@ describe('GraphQLBridge', () => {
...
+            initialValue: 'Some Title',
             options: [
                 {label: 1, value: 'a'},
-                {label: 2, value: 'b'}
+                {label: 2, value: 'b'},
+                {label: 3, value: 'Some Title'}
             ]
         },
         votes: {
+            initialValue: '44',
```
radekmie: And here we have the same issue - it should be a number.

dude-awap: I tried with ...

    votes: {
      initialValue: 44,

... before committing, but the test only passed if I quoted the number?  I presume `bridge.getInitialValue('votes')` is returning a `String`.  I do not know how to fix that.

radekmie: It's because your test expects `'44'` In line 161.

PR #189, opened 2017-02-10
Allows material BoolFields to be checkboxes or toggles

dpankros:
This PR adds an appearance property to BoolField that allows rendering as a Toggle if appearance set to 'toggle'.  Default or any other value is 'checkbox' (present behavior) so as to not break any existing apps.
That is, ```<BoolField name='field' />``` still works as it previously did by rendering a checkbox, but now you can alternatively use toggles: ```<BoolField name='field' appearance='toggle' />```

Updated combined tests for the above (x35)

NOTE: I did **not** update any package version numbers.

### DIFF
--- a/packages/uniforms-material/src/BoolField.js
+++ b/packages/uniforms-material/src/BoolField.js
@@ -1,10 +1,11 @@
 import Checkbox       from 'material-ui/Checkbox';
-import Toggle         from 'material-ui/Toggle';
 import React          from 'react';
+import Toggle         from 'material-ui/Toggle';
 import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
 
 const Bool = ({
+    appearance = 'checkbox',
     disabled,
     id,
     inputRef,
@@ -12,11 +13,10 @@ const Bool = ({
     name,
     onChange,
     value,
-    appearance = 'checkbox',
     ...props
 }) =>
-    appearance === 'toggle' ?
-        (<Toggle
+    appearance === 'toggle' ? (
+        <Toggle
             toggled={!!value}
             disabled={disabled}
             id={id}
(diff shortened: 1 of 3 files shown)


packages/uniforms-material/test/combined.js
```diff
@@ -7,6 +7,7 @@ import {spy}      from 'sinon';
 import {stub}     from 'sinon';
 
 import MaterialCheckbox    from 'material-ui/Checkbox';
+import MaterialToggle      from 'material-ui/Toggle';
```
radekmie: Keep them sorted.

packages/uniforms-material/src/BoolField.js
```diff
@@ -1,4 +1,5 @@
 import Checkbox       from 'material-ui/Checkbox';
+import Toggle         from 'material-ui/Toggle';
```
radekmie: Keep imports sorted.

packages/uniforms-material/src/BoolField.js
```diff
@@ -11,18 +12,28 @@ const Bool = ({
     name,
     onChange,
     value,
+    appearance = 'checkbox',
```
radekmie: Props also should be sorted.

packages/uniforms-material/src/BoolField.js
```diff
@@ -11,18 +12,28 @@ const Bool = ({
...
+    :
+        (<Checkbox
+            checked={!!value}
+            disabled={disabled}
+            id={id}
+            label={label}
+            name={name}
+            onCheck={(event, value) => disabled || onChange(value)}
+            ref={inputRef}{...filterDOMProps(props)}
+        />);
```
radekmie: Two things here:
* keep props sorted
* ternary should be formatted like this:
```js
appearance === 'toggle' ? (
    <Toggle
        ...
    />
) : (
    <Checkbox
        ...
    />
)
```

radekmie:
Great! And I'm **really** glad to see, that you've written tests!

dpankros:
Updated.  
Sorry about the non-conforming code.  I didn't see any coding standards doc and these were not caught by your lint spec.  The ternary is actually in conflict with an internal standard that we have so my JS IDE wouldn't even let me do it.

radekmie:
Yeah, I have to update ESLint finally... Thanks!

Issue #209, opened 2017-02-28
Demo textarea styles breaking package specific styles


janowsiany:
This [styles](https://github.com/vazco/uniforms/blob/master/demo/components/ApplicationPropsField.js#L27-L32) need to be scoped so they will not match elements in the right panel. This issue makes #208 impossible to be checked in demo app.

PR #210, opened 2017-02-28
Scope for demo app styles

janowsiany:
Solution for issue #209

### DIFF
--- a/demo/main.css
+++ b/demo/main.css
@@ -75,7 +75,7 @@ img {
     overflow-y: scroll;
 }
 
-.left > textarea {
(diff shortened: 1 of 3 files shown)


radekmie:
As there are no `iframes` anymore let's just move it to the styles instead.

demo/main.css
```diff
@@ -74,3 +74,8 @@ img {
     width: 100%;
     overflow-y: scroll;
 }
+
+.left > textarea {
```
radekmie: How about `.panelLeft`? It would be more in align with the rest of classes.

Issue #203, opened 2017-02-23
It would be useful make select field searchable in uniforms-antd


Gervwyk:
Not sure if it makes sense to includes this as an optional, but I think it might as well be always enabled. 

Its real simple. Have a look at:
https://ant.design/components/select/#components-select-demo-search
it suggests to add:
`filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}`
to Select props. 

The multi-select option is currently searchable. I have a suspicion that the `transform` option should be included here, since users want to search by label and not by value. However the [selectfield](https://github.com/vazco/uniforms/blob/master/API.md#selectfield) documentation in lacking.

Would be useful since select list are often long (countries).

radekmie:
I'm strongly against setting it as a default, but there should be `filterDOMProps` In the `SelectField`, so you can use it already, like it's used in other fields and themes. Would you like to submit a PR for it?

Gervwyk:
Agreed. Will try some options tomorrow and send PR.

radekmie:
@Gervwyk?

Gervwyk:
Cool. It worked. Check PR.
Schema example for future users:
````
new SimpleSchema({
    size: {
        type: String,
        defaultValue: 'm',
        allowedValues: ['xs', 's', 'm', 'l', 'xl'],
        optional: true,
        uniforms: {
         showSearch: true,
         filterOption: (input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0
        }
    }
})
````

Issue #207, opened 2017-02-28
how to set "horizontal, vertical" form of "ant form"?


thearabbit:
how to set "horizontal, vertical" form of "ant form"?
```js
<AutoForm
    schema={PostSchema}
    showInlineError={true}
    onSubmitSuccess={() => alert('Success!')}
    onSubmit={doc => console.log(doc)}
>
    <AutoField name="category"/>
    <AutoField name="authors"/>
    <AutoField name="publishedDate"/>
    <AutoField name="published"/>

    <SubmitField type="primary"/>
</AutoForm>
```

![image](https://cloud.githubusercontent.com/assets/4495160/23402482/5777f8fa-fdde-11e6-9c62-94454106cd7e.png)

janowsiany:
Hey @thearabbit, i think it is not handled yet. Do you want to make a pull request?

radekmie:
Also, let's call the author: @Gervwyk.

Gervwyk:
It is not handled yet. I agree with @janowsiany ,  `labelCol` and `wrapperCol` will be this best way to implement this - will be nice to have. I won't be able to look into this until at least next week. A PR would be great.

thearabbit:
Thanks for all reply, waiting...

thearabbit:
Could example? I tried, but don't work
```js
render() {
        const formItemLayout = {
            labelCol: {span: 6},
            wrapperCol: {span: 14},
        };

        return (
            <div>
                <h2>Welcome to Meteor with Ant Design!</h2>

                <AutoForm
                    schema={PostSchema}
                    showInlineError={true}
                    onSubmitSuccess={() => alert('Success!')}
                    onSubmit={doc => console.log(doc)}
                >
                    <AutoField name="category" {...formItemLayout}/>
                    <AutoField name="authors" {...formItemLayout}/>
                    <DateField name="publishedDate" showTime={false} format="DD/MM/YYYY" {...formItemLayout}/>

                    <DateRangeField
                        name="publishedRangeDate"
                        showTime={false}
                        {...formItemLayout}
                    />

                    <AutoField name="published" {...formItemLayout}/>

                    <SubmitField type="primary"/>
                </AutoForm>

            </div>
        );
    }
```
Please help me.

radekmie:
It's not released yet.

thearabbit:
Oh, Thanks I will try again.

thearabbit:
Have problem with "BoolField"
```js
[...]

Issue #185, opened 2017-02-05
Help, please.  How to add custom attribute to field?


dude-awap:
I do my end-to-end (functional) testing with Cucumber.  Cucumber (like Selenium) can use XPath or CSS selectors on third-party web pages.  When testing one's own code however, it is recommended to select tags for test automation with explicit unique custom attributes, such as `data-cuke="title"` in the example below : 

     <input type="text" data-cuke="title" id="tfid-104-0" name="title" class="form-control" value="">

Up 'til now I have been using [tcomb-form](https://www.npmjs.com/package/tcomb-form), which permits me to do either ...

    title: {
      label: 'Title (custom label)',
      attrs: {
        'data-cuke': 'title'
      }
    },

... or ...

    <button
      data-cuke="save-item" className="btn btn-primary"
      onClick={this.submitForm}>{buttonLabel}
    </button>

**How can I achieve the same thing with Uniforms"?**

radekmie:
All fields are using `filterDOMProps` helper _(or at least should)_, which will leave non-uniforms props untouched, so code like:

```js
<SubmitField data-cuke="save-item" />
```

Should work as expected. If any component is not doing it, then let me know.

dude-awap:
Ok, so I tried with several different fields as you suggest : [Uniforms changes diff](https://github.com/yourse1f-yourorg/mmks/commit/99a5d365b8abc7bfd5469dfd5c62accdbbc7156c)

In only one case did my custom attribute appear ...

    <section class="" data-cuke="save-item">
      <input type="submit" class="btn btn-primary">
    </section>

... and on the section rather than the specific tag, :-(

radekmie:
Well, yes. Although we can't get _a lot_ better because fields like [`ListField` from uniforms-semantic](https://github.com/vazco/uniforms/blob/master/packages/uniforms-semantic/src/ListField.js) are rendering five different elements - it would be an overkill to allow settings props for each of it. I see two solutions: either leave it as it is now _(it's not a big problem to have a select element like `[data-cuke="save-item"] > input`)_ or if you _really_ want to have a direct selector, you can create a custom field.

dude-awap:
I would happily create a custom field, if I could understand how.  Are you suggesting to subclass the unstyled TextField's [input tag](https://github.com/vazco/uniforms/blob/master/packages/uniforms-unstyled/src/TextField.js#L24), and then add attributes to it like this : 

[...]

PR #211, opened 2017-02-28
Include filterDOMProps in fields

Gervwyk:
Solves #203


Gervwyk:
Oops. Sorry. Fixed.

janowsiany:
@Gervwyk I think it will be ok if we add `{...filterDOMProps(props)}` in `wrapField.js` also. It will solve #207. @radekmie What do you think about this?

packages/uniforms-antd/src/SubmitField.js
```diff
@@ -8,7 +8,6 @@ const SubmitField = ({className, inputRef, value, ...props}, {uniforms: {error,
         htmlType="submit"
         ref={inputRef}
         type="primary"
-        {...props}
```
radekmie: Why have you removed it? We shouldn't use `filterDOMProps` here as it's not wrapped with `connectField`, but it's still needed.

Gervwyk: Ahh. Sorry. That was just a typo of working on to many things at once.. My bad.

Gervwyk:
@janowsiany I was also considering #207 but still wanted to experiment with it. I'll do some tests later today then we can combine it in this PR. @radekmie what do you think of including `filterDOMProps` in `wrapField`?

radekmie:
We have to choose one way somewhere in the future _(2.0.0 probably, as it will be a breaking change)_: where _all_ unused props go. To the wrapper? To the input? To the inner wrapper _(if any)_? To nowhere _(discard)_? Right now, it's depending on the theme and field.

Gervwyk:
I struggle to see how it will not be dependant on the theme and field. However, I think its a case of either assigning props to the wrapper or to the field on schema level - which will mean a breaking change or we can say to include a object for wrapper props and all other props are assigned on field level.
Something like:
````
new SimpleSchema({
    size: {
        type: String,
        allowedValues: ['xs', 's', 'm', 'l', 'xl'],
        uniforms: {
         showSearch: true,
         filterOption: (input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0
         wrapperProps: { 
              labelCol: {span: 3, offset: 12},
              wrapperCol: {span: 3, offset: 12}
        }
    }
})
````
Where `showSearch` and `filterOption` end up as field props and `labelCol` and `wrapperCol` on the wrapper component. Think this might be less of a breaking change... Not sure how this fit in with other themes.

radekmie:
Yep, there's also #185. All in all - it shouldn't be discussed here.

Issue #208, opened 2017-02-28
LongTextField does not have a "rows" prop


tab00:
`LongTextField` is meant to render as a textarea, i.e. a multi-row field for user text input. However the API does not have any prop to set the number of rows: https://github.com/vazco/uniforms/blob/10ef9ba2445f4d23051109ab0ac19ba48bc52038/API.md#longtextfield

How can we set the number of rows?

janowsiany:
Hello @tab00 can you specify which package are you talking about?

tab00:
I am using the uniforms-bootstrap4 theme, though is the theme actually relevant?

janowsiany:
Components such as LongTextField are handled differently in each packages, specifically for the package like `bootstrap4` or `material-ui`. However as you pointed rows property is not passed down to textarea input as you can see in the [sources](https://github.com/vazco/uniforms/blob/master/packages/uniforms-bootstrap4/src/LongTextField.js#L10-L17).
Would you like to handle `rows` property there and submit a PR?

tab00:
Would simply adding `rows={props.rows}` in the section that you have referred to fix it?:

```
const LongText = props =>
    wrapField(props, (
        <textarea
            className={classnames(props.inputClassName, 'form-control', {'form-control-danger': props.error})}
            disabled={props.disabled}
            id={props.id}
            name={props.name}
            onChange={event => props.onChange(event.target.value)}
            placeholder={props.placeholder}
            ref={props.inputRef}
            value={props.value}
            rows={props.rows}
        />
   ))

```

janowsiany:
It is that simple.

Don't bother if it will not work in demo app, i see now that it has some [custom styles](https://github.com/vazco/uniforms/blob/master/demo/components/ApplicationPropsField.js#L28-L31) which should only refer to left panel. Make your PR anyway.

radekmie:
@janowsiany - use `filterDOMProps` instead of passing every prop. I'll try to cover it, but as always PRs are encouraged.

PR #213, opened 2017-03-01
Pass down rows prop of LongTextField  to textarea input

tab00:
For issue https://github.com/vazco/uniforms/issues/208


packages/uniforms-bootstrap4/src/LongTextField.js
```diff
@@ -15,6 +15,7 @@ const LongText = props =>
             placeholder={props.placeholder}
             ref={props.inputRef}
             value={props.value}
+            rows={props.rows}
```
radekmie: Please, keep props sorted.

radekmie:
@tab00 great thanks for the PR!

PR #222, opened 2017-03-04
Allow labelCol and wrapperCol props on Form Item

janowsiany:
Solves #207.

### DIFF
--- a/packages/uniforms-antd/src/ListField.js
+++ b/packages/uniforms-antd/src/ListField.js
@@ -63,8 +63,8 @@ const List = ({
                 <ListItemField
                     key={index}
                     label={null}
-                    name={joinName(name, index)}
                     labelCol={labelCol}
+                    name={joinName(name, index)}
                     wrapperCol={wrapperCol}
                     {...itemProps}
                 />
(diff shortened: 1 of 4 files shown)


radekmie:
Have you tested it?

janowsiany:
You were right, there were some warrnings about unssuported properties on elements.

Gervwyk:
This would be great if this can work. Do you have a test schema for it? I'm trying it in the demo but not finding expected results... 

I think we might have to also change `BaseForm` to render the antd `Form` component instead on default `form` ?
Please have a look at (form layout options)[https://ant.design/components/form/#components-form-demo-layout]

janowsiany:
It works with demo application and schema:
```
new SimpleSchema({
    name: {
        type: String,
        uniforms: {
            wrapperCol: {span: 12}, 
            labelCol: {span: 4}
        }
    }
})
```
It breaks a layout in ListItemField a little bit, however it might by fixed by adding `itemProps: {style: {marginTop: SOME_MARGIN}}`. 

I agree that using `Form` component of `antd` package might be more complete solution. Yet this solution seems to be more invasive?

Gervwyk:
Thanks. Sorry. Also had it like that but made a mistake in build. I think this is good. You can go ahead and fix the margin for `ListItemField`.

For now lets keep the form as is. In future we might want `inline` mode but for this PR this is good.

PR #230, opened 2017-03-11
Fixes for wrapperCol and labelCol, adding Bool field to demo app and docs update

janowsiany:
Fixing #207

### DIFF
--- a/API.md
+++ b/API.md
@@ -620,6 +620,13 @@ import NestField from 'uniforms-unstyled/NestField'; // Choose your theme packag
     //   bootstrap4
     //   semantic
     showInlineError={true}
+
+    // Add / remove "grouped" class from the field.
+    //   In Semantic, fields can be grouped using this class. By default,
+    //   this class is added to the NestField.
+    // Available in:
+    //   semantic
+    grouped={true}
 />
 ```
 
--- a/API.md
+++ b/API.md
@@ -162,6 +162,20 @@ import BoolField from 'uniforms-unstyled/BoolField'; // Choose your theme packag
     grid={{md: 5}}  // 'col-5-md' on label, 'col-7-md' on input
     grid="col-6-xl" // 'col-6-xl' on label, 'col-6-xl' on input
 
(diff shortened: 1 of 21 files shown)


janowsiany:
I am not sure what else should be placed in docs. Any ideas?

API.md
```diff
@@ -162,6 +162,20 @@ import BoolField from 'uniforms-unstyled/BoolField'; // Choose your theme packag
     grid={{md: 5}}  // 'col-5-md' on label, 'col-7-md' on input
     grid="col-6-xl" // 'col-6-xl' on label, 'col-6-xl' on input
 
+    // Field layout
+    //  The layout of label. You can set span and/or offset.
+    // Available in:
+    //   antd
+    labelCol={span: 4}   // 'ant-col-4' on label
```
radekmie: Here _(and all occurrences below)_: `{{span: 4}}` - it's an object.

packages/uniforms-antd/src/BoolField.js
```diff
@@ -6,8 +6,8 @@ import filterDOMProps from 'uniforms/filterDOMProps';
 
 import wrapField from './wrapField';
 
-const Bool = props =>
-    wrapField(props, (
+const Bool = ({labelCol, wrapperCol, ...props}) =>
```
radekmie: In #229 we've added `colon` and `required` - these also should be treated like this. By the way, we should find a better way to handle these... How about `filterDOMProps.register` them?

janowsiany:
`grouped` property is already added to API, you have even praised this [here](https://github.com/vazco/uniforms/pull/227/files#r104743614), haven't you?

packages/uniforms-antd/src/BoolField.js
janowsiany: Right, i beleive `filterDOMProps.register` is cleaner solution here. Thanks that you point them to me.

radekmie:
Well, yeah, my fault.

PR #232, opened 2017-03-12
latest material-ui as dependency (closes #231)

### LINKED ISSUE #231: Installation with latest material-ui not working
I am using the latest version of material-ui: "material-ui": "^0.17.0" and therefore I can't install uniforms-material because its dependency is material-ui@^0.16.7.

After trying to install I get the following error:
npm WARN uniforms-material@1.14.0 requires a peer of material-ui@^0.16.7 but none was installed.

h2s20:
Updates the material-ui dependency to the latest version: material-ui 0.17.0

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -17,7 +17,7 @@
     "eslint-plugin-react": "6.10.0",
     "eslint-plugin-vazco": "1.0.0",
     "graphql": "0.9.1",
-    "material-ui": "0.17.0 || 0.16.7",
+    "material-ui": "0.17.0",
     "meteor-node-stubs": "0.2.5",
     "react": "15.4.2",
     "react-dom": "15.4.2",
--- a/demo/package.json
+++ b/demo/package.json
@@ -17,7 +17,7 @@
     "eslint-plugin-react": "6.10.0",
     "eslint-plugin-vazco": "1.0.0",
     "graphql": "0.9.1",
-    "material-ui": "0.16.7",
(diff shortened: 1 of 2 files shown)


packages/uniforms-material/package.json
```diff
@@ -65,7 +65,7 @@
     "sinon": "2.0.0-pre.6"
   },
   "peerDependencies": {
-    "material-ui": "^0.16.7",
+    "material-ui": "^0.17.0",
```
radekmie: To be backwards compatible, let's use `^0.17.0 || ^0.16.7`.

demo/package.json
```diff
@@ -17,7 +17,7 @@
     "eslint-plugin-react": "6.10.0",
     "eslint-plugin-vazco": "1.0.0",
     "graphql": "0.9.1",
-    "material-ui": "0.16.7",
+    "material-ui": "0.17.0 || 0.16.7",
```
radekmie: Here...

packages/uniforms-material/package.json
```diff
@@ -54,7 +54,7 @@
     "eslint-plugin-react": "6.10.0",
     "eslint-plugin-vazco": "1.0.0",
     "jsdom": "9.11.0",
-    "material-ui": "0.16.7",
+    "material-ui": "0.17.0 || 0.16.7",
```
radekmie: ...and here, it would be better to keep only one possible version, to make tests deterministic.

radekmie:
Thank you, @h2s20!

Issue #236, opened 2017-03-21
problem with "multi select" on Material?


thearabbit:
- I try to upgrade `"uniforms": "^1.15.0"`, `"material-ui": "^0.17.1"`.
```js
// schema
categories: {
        type: [String],
        allowedValues: [
            "news",
            "image",
            "video"
        ]
    },

// Component
<SelectField
                        name="categories"
                        multiple={true}
                        hintText="Select a name"
                    />
```
- It show `Unknown prop `hintText` on <div> tag. Remove this prop from the element. For details...`
![image](https://cloud.githubusercontent.com/assets/4495160/24131513/a9a3b198-0e21-11e7-9207-21e3c3742cab.png)

- Multi Select on `"material-ui": "^0.17.1"`
![image](https://cloud.githubusercontent.com/assets/4495160/24131581/0663413c-0e22-11e7-965b-272b7e9ed032.png)

radekmie:
I'll get to it soon.

radekmie:
OK, now I get it. It's reasonable, that this warning shows up - there's no support for the multi-select within `SelectField` - there's only checkboxes _fallback_. @janowsiany, do you want to implement this?

janowsiany:
I will handle this later today.

PR #239, opened 2017-03-21
Support multiple on material ui SelectField

janowsiany:
Solves #236 
Wondering if i may do something like this to avoid triple field type checking:
```js
const renderSelect = ({
    ...props
}, context, multiple = props.fieldType === Array) => {
/* Rendering */
}
```
What do you think about this?

### DIFF
--- a/packages/uniforms-material/src/SelectField.js
+++ b/packages/uniforms-material/src/SelectField.js
@@ -90,7 +90,7 @@ const renderSelect = ({
         {allowedValues.map(value =>
             <MenuItem
                 insetChildren={props.fieldType === Array}
-                checked={props.fieldType === Array && props.value.includes(value)}
+                checked={props.fieldType === Array && props.value && props.value.includes(value)}
                 key={value}
                 primaryText={transform ? transform(value) : value}
                 value={value}
--- a/packages/uniforms-material/src/SelectField.js
+++ b/packages/uniforms-material/src/SelectField.js
@@ -71,7 +71,6 @@ const renderSelect = ({
     onChange,
     placeholder,
     transform,
-    value,
     ...props
 }) =>
     <SelectField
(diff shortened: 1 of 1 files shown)


packages/uniforms-material/src/SelectField.js
```diff
@@ -81,26 +80,33 @@ const renderSelect = ({
         fullWidth={fullWidth}
         hintText={placeholder}
         id={id}
+        multiple={props.fieldType === Array}
```
radekmie: You can add `fieldType` to spread above.

packages/uniforms-material/src/SelectField.js
```diff
@@ -81,26 +80,33 @@ const renderSelect = ({
...
         name={name}
         onChange={(event, index, value) => onChange(value)}
         ref={inputRef}
-        value={value}
+        value={props.value}
```
radekmie: I'll rather leave it as it is...

packages/uniforms-material/src/SelectField.js
```diff
@@ -81,26 +80,33 @@ const renderSelect = ({
...
         {...filterDOMProps(props)}
     >
         {allowedValues.map(value =>
```
radekmie: ...and rename this variable.

packages/uniforms-material/src/SelectField.js
```diff
@@ -81,26 +80,33 @@ const renderSelect = ({
...
             <MenuItem
+                insetChildren={props.fieldType === Array}
+                checked={props.fieldType === Array && props.value && props.value.includes(value)}
```
radekmie: Shouldn't we pass `undefined`, when `fieldType !== Array`?

packages/uniforms-material/src/SelectField.js
```diff
@@ -81,26 +80,33 @@ const renderSelect = ({
...
+    props.checkboxes
         ? renderCheckboxes(props)
         : renderSelect    (props)
 ;
 
 export default connectField(Select, {ensureValue: false});
+
+filterDOMProps.register(
+    'checkboxes'
+);
```
radekmie: It's not that obvious... I think we should extract it manually - maybe we'll use it, but let's not include it in this PR.

PR #249, opened 2017-03-28
Allow to override button classes

rafinskipg:
Using this method we can pass this options to override the value of the badge class correctly

```
let classNameAddField = { 'badge': false, btn: true};

<ListAddField className={classNameAddField} name={`${name}.$`} initialCount={initialCount} addIcon={addIcon} />
```

This way is more correct than doing 

```
let classNameAddField = { 'badge pull-right': false, btn: true, 'pull-right': true};
<ListAddField className={classNameAddField} name={`${name}.$`} initialCount={initialCount} addIcon={addIcon} />
```

### DIFF
--- a/packages/uniforms-bootstrap3/src/ListAddField.js
+++ b/packages/uniforms-bootstrap3/src/ListAddField.js
@@ -15,7 +15,7 @@ const ListAdd = ({
 
     return (
         <div
-            className={classnames('badge pull-right', className)}
+            className={classnames({ badge: true, 'pull-right': true}, className)}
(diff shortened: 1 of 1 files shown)


rafinskipg:
Sorry @radekmie I think this does not work anyway. Because `classnames({badge: true, 'pull-right': true}, {badge: false}) === 'badge pull-right'`

rafinskipg:
Yeah, I saw that it can be done by using https://github.com/JedWatson/classnames#alternate-dedupe-version

But now it can't be done with this current `classnames` package.

radekmie:
Yes I don't want to switch to the deduce version, because it will break semantic ui.

PR #254, opened 2017-03-31
Add placeholder in simpleschemabridge

rafinskipg:
I found them very useful for the selects and not being in the schema (while having the label there) is weird.


radekmie:
It's not needed, because fields should take care of this prop - it's one of the guaranteed ones. If it's not, then it's a bug.

Issue #253, opened 2017-03-31
select in antd with options fails to render label


rafinskipg:
In the `antd` SelectField component if you pass a set of options using {label: xx, value: yy} it renders the value instead of the label. In the bootstrap selectfield this does not happen

rafinskipg:
Also if the field is optional the placeholder / label is not set inside the options

radekmie:
As discussed in the issue and on Gitter - I'm closing.

PR #255, opened 2017-03-31
Fix ant-d select field

rafinskipg:
Relates to #253

### DIFF
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -4,7 +4,6 @@ import React          from 'react';
 import Select         from 'antd/lib/select';
 import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
-
 import wrapField from './wrapField';
 
 const renderCheckboxes = props =>
@@ -43,7 +42,7 @@ const renderCheckboxes = props =>
     )
 ;
 
-const renderSelect = props =>
+const renderSelect = props => 
     <Select
         allowClear={!props.required}
         disabled={props.disabled}
@@ -52,11 +51,18 @@ const renderSelect = props =>
         name={props.name}
         onChange={value => props.onChange(value)}
         ref={props.inputRef}
-        value={props.fieldType === Array ? props.value || [] : '' + (props.value || '')}
+        value={props.fieldType === Array ? props.value || [] : '' + ((props.transform ? props.transform(props.value) : props.value) || '')}
         {...filterDOMProps(props)}
     >
+
+         {(!!props.placeholder || !props.required) && (
+            <option value="" disabled={props.required} hidden={props.required}>
+                {props.placeholder ? props.placeholder : props.label}
+            </option>
+        )}
+
         {props.allowedValues.map(value =>
-            <Select.Option key={value} value={value}>
+            <Select.Option value={value} >
                 {props.transform ? props.transform(value) : value}
             </Select.Option>
         )}
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -4,6 +4,7 @@ import React          from 'react';
 import Select         from 'antd/lib/select';
 import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
+
 import wrapField from './wrapField';
 
 const renderCheckboxes = props =>
@@ -63,7 +64,7 @@ const renderSelect = props =>
         )}
 
         {props.allowedValues.map(value =>
-            <Select.Option key={value} value={value} >
+            <Select.Option key={value} value={value}>
                 {props.transform ? props.transform(value) : value}
             </Select.Option>
         )}
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -62,7 +62,7 @@ const renderSelect = props =>
         )}
 
         {props.allowedValues.map(value =>
-            <Select.Option value={value} >
+            <Select.Option key={value} value={value} >
                 {props.transform ? props.transform(value) : value}
             </Select.Option>
         )}
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -52,7 +52,7 @@ const renderSelect = props =>
         onChange={value => props.onChange(value)}
         ref={props.inputRef}
         value={props.fieldType === Array ? props.value || [] : '' +
-            ((props.transform ? props.transform(props.value) : props.value) || '')}
+            ((props.transform ? props.transform(props.value.toString()) : props.value) || '')}
         {...filterDOMProps(props)}
     >
 
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -56,9 +56,9 @@ const renderSelect = props =>
     >
 
          {(!!props.placeholder || !props.required) && (
-            <option value="" disabled={props.required} hidden={props.required}>
+            <Select.Option value="" disabled={props.required} hidden={props.required}>
                 {props.placeholder ? props.placeholder : props.label}
-            </option>
+            </Select.Option>
         )}
 
         {props.allowedValues.map(value =>
--- a/packages/uniforms-antd/src/SelectField.js
+++ b/packages/uniforms-antd/src/SelectField.js
@@ -52,8 +52,11 @@ const renderSelect = props =>
         name={props.name}
         onChange={value => props.onChange(value)}
         ref={props.inputRef}
-        value={props.fieldType === Array ? props.value || [] : '' +
-            ((props.transform ? props.transform(props.value.toString()) : props.value) || '')}
+        value={props.fieldType === Array
+            ? props.value || []
+            : props.value
+                ? props.transform ? '' + props.transform(props.value) : '' + props.value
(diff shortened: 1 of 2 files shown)


rafinskipg:
I have seen, testing in another application that this does not work correctly for some reason. 

In the demo application is working, but not in mine, so let's find the problems before merging :dancer:

packages/uniforms-antd/src/SelectField.js
```diff
@@ -52,11 +51,19 @@ const renderSelect = props =>
         name={props.name}
         onChange={value => props.onChange(value)}
         ref={props.inputRef}
-        value={props.fieldType === Array ? props.value || [] : '' + (props.value || '')}
+        value={props.fieldType === Array ? props.value || [] : '' +
+            ((props.transform ? props.transform(props.value) : props.value) || '')}
```
radekmie: It got very complex, so let's do it in this way:
```js
value={props.fieldType === Array
    ? props.value || []
    : props.value
        ? props.transform ? '' + props.transform(props.value) : '' + props.value
        : ''
}
```

packages/uniforms-antd/src/SelectField.js
```diff
@@ -52,11 +51,19 @@ const renderSelect = props =>
...
+
+        {(!!props.placeholder || !props.required) && (
+            <Select.Option value="" disabled={props.required} hidden={props.required}>
+                {props.placeholder ? props.placeholder : props.label}
+            </Select.Option>
+        )}
+
         {props.allowedValues.map(value =>
-            <Select.Option key={value} value={value}>
+            <Select.Option key={value} value={value} >
```
radekmie: Why is this space added here?

packages/uniforms/src/SimpleSchemaBridge.js
```diff
@@ -1,11 +1,10 @@
 /* global Package */
 
+import Bridge         from './Bridge';
 import cloneDeep from 'lodash.clonedeep';
+import filterDOMProps from './filterDOMProps';
 import invariant from 'fbjs/lib/invariant';
-
-import Bridge         from './Bridge';
 import joinName       from './joinName';
-import filterDOMProps from './filterDOMProps';
```
radekmie: Why you've changed it? At the top, we list all npm packages, then local dependencies.

packages/uniforms/src/SimpleSchemaBridge.js
```diff
@@ -202,7 +201,8 @@ export default class SimpleSchemaBridge extends Bridge {
             } else {
                 field = {
                     ...field,
-                    transform: value => options.find(option => option.value === value).label,
+                    transform: value => options.find(option => option.value === value) ?
+                        options.find(option => option.value === value).label : value,
```
radekmie: It's not a bug, but by design - if your field has allowed values _(indirectly, through `option`)_, then it's required to have current value there too. We can discuss it anyway, but please do it in a separate issue and PR.

packages/uniforms-antd/src/SelectField.js
```diff
@@ -4,7 +4,6 @@ import React          from 'react';
 import Select         from 'antd/lib/select';
 import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
-
```
radekmie: This line is kept to keep npm and local dependenices separated.

packages/uniforms/src/SimpleSchemaBridge.js
rafinskipg: I have an automatic sort imports that sorts all this things.

packages/uniforms-antd/src/SelectField.js
rafinskipg: changing it

packages/uniforms/src/SimpleSchemaBridge.js
rafinskipg: Ok, i do it separately

rafinskipg:
I have some problems with this when using it directly in my meteor project (with the react-blaze-helper)

```
{(!!props.placeholder || !props.required) && (
            <Select.Option value="" disabled={props.required} hidden={props.required}>
                {props.placeholder ? props.placeholder : props.label}
            </Select.Option>
        )}

```

it says that invariant is violated. But If i directly put the `select.option` without the surrounding code it works. 

In the demo it works correctly

radekmie:
What invariant exactly?

packages/uniforms/src/SimpleSchemaBridge.js
```diff
@@ -202,8 +202,8 @@ export default class SimpleSchemaBridge extends Bridge {
             } else {
                 field = {
                     ...field,
-                    transform: value => options.find(option => option.value === value).label,
-                    allowedValues: options.map(option => option.value)
+                    transform: value => options.find(option => option.value.toString() === value).label,
+                    allowedValues: options.map(option => option.value.toString())
```
radekmie: It shouldn't happen here, at schema level, because `option.value` might be an object. I know, that AntD disallows to use non-strings as options, but it should happen at the component level then.

rafinskipg: @radekmie has the component any way of doing a `toString` of the options for the transform and the allowedValues ? 

I would need some help from you to solve this PR.

thanks

radekmie: You can do it in the `SelectField` of AntD theme. If you still need some help, let me know.

rafinskipg: @radekmie I knew that, i need help in the way to do it. 

Right now i had to change the options and make the "toString" of all the data that is passed to the field. But i don't know hot to fix it in the antd selectfield

radekmie: _`packages/uniforms-antd/src/SelectField.js`_

Line 56: `.map` over this array to convert it to an array of strings.
Lines 58 and 70: convert it to string directly.

radekmie:
@rafinskipg?

packages/uniforms/src/SimpleSchemaBridge.js
rafinskipg: Hi @radekmie , I don't how that should work.  The problem is in the transform function. 

Just for being sure, this is what you suggest?

```html
    <Select
        allowClear={!props.required}
        disabled={props.disabled}
        id={props.id}
        multiple={props.fieldType === Array}
        name={props.name}
        onChange={value => props.onChange(value)}
        ref={props.inputRef}
        value={props.fieldType === Array
            ? props.value.map(item => item.toString) || []
            : props.value
                ? props.transform ? '' + props.transform(props.value.toString()) : '' + props.value
                : ''}
        {...filterDOMProps(props)}
    >

        {(!!props.placeholder || !props.required) && (
            <Select.Option value="" disabled={props.required} hidden={props.required}>
                {props.placeholder ? props.placeholder : props.label}
            </Select.Option>
        )}

        {props.allowedValues.map(value =>
            <Select.Option key={value} value={value}>
                {props.transform ? props.transform(value.toString()) : value}
            </Select.Option>
        )}
    </Select>
```

The real problem is in the `transform` function. The `value` is allways a string
The problem is that the transform function will be comparing the `value` string to the set of allowed values that might be numbers. So they will never be equal.

rafinskipg: In my project I ended doing a `toString` of the options passed to the schema...

```
uniforms: {
            options: PlaceTypeEnum.getOptions()
                .map(function(item){
                    return {
                        label: item.label,
                        value: item.value.toString() //Was an integer
                    }
                })
        }
```

radekmie: And how about:
```js
<Select
    allowClear={!props.required}
    disabled={props.disabled}
    id={props.id}
    multiple={props.fieldType === Array}
    name={props.name}
    onChange={value => props.onChange(value)}
    ref={props.inputRef}
    value={props.fieldType === Array
        ? props.value ? props.value.map(item => '' + item) : []
        : props.value
            ? props.transform ? '' + props.transform(props.value) : '' + props.value
            : ''}
    {...filterDOMProps(props)}
>

    {(!!props.placeholder || !props.required) && (
        <Select.Option disabled={props.required} hidden={props.required} value="">
            {props.placeholder ? props.placeholder : props.label}
        </Select.Option>
    )}

    {props.allowedValues.map(value =>
        <Select.Option key={value} value={'' + value}>
            {props.transform ? '' + props.transform(value) : '' + value}
        </Select.Option>
    )}
</Select>
```

rafinskipg: But when you select 1 item, and the stuff gets re-rendered  the value is already a string, no? Let me check it tomorrow.

rafinskipg: It does not work @radekmie , the problem is still there because the transform function is comparing an integer to a string, and does not return the label.

radekmie: Then we have to think out something else... I'll think about it at the weekend.

Gervwyk: I had a issue with my select only returning values and not labels. I found that the options I was passing had some of the labels undefined.. So I would recommend carefully checking the results of `PlaceTypeEnum.getOptions()` in you schema there. But also, I had problems when one of the items in the array was `null` but can't remember if it was related.

radekmie: @Gervwyk it doesn't seem to be related but if you'll recreate it, report it please.

PR #256, opened 2017-03-31
If the value is not found in the set of options(because it can be an …

rafinskipg:
…old value) don't break, show empty

### DIFF
--- a/packages/uniforms/src/SimpleSchemaBridge.js
+++ b/packages/uniforms/src/SimpleSchemaBridge.js
@@ -202,8 +202,11 @@ export default class SimpleSchemaBridge extends Bridge {
             } else {
                 field = {
                     ...field,
-                    transform: value => options.find(option => option.value === value).label,
-                    allowedValues: options.map(option => option.value)
+                    transform: value => {
+                        const option = options.find(option => option.value.toString() === value);
+                        return option ? option.label : '';
+                    },
+                    allowedValues: options.map(option => option.value.toString())
                 };
             }
         }


radekmie:
I'm still not convinced. Could you show me an example of such use case?

rafinskipg:
Well, figure that you have a set of values for a select.

houses : ['id1', 'id2', 'id3'] 

And after some months one value gets disabled and not send to the front end. The select will receive a set of options : `[{ label: 'house martin', value : 'id1'}, {label: 'house jeremy': value: 'id2'}];`. All the users that had selected `id3` won't be able to load the UI because the component will crash with the `cannot find label of undefined`.

I know that if `id3` has been deactivated from the systems the developer should remove it from the selected items, but this is a maybe. It might want to show them the value in some scenarios and show a `is deactivated, please change the house`. 


I think a warning in the console is enough.

radekmie:
Then you should add it to `allowedValues` with _Disabled because of..._ label and disallow it in your schema - otherwise, you want to allow a not allowed value.

janowsiany:
I agree with Radek, i think that package for building forms should not handle your business logic.

rafinskipg:
It is only a matter of not throwing such an exception if the item is not found on the options. 
But ok, i get the point.

radekmie:
@rafinskipg just to clarify - it's not a problem to bypass or silent this error, but then your schema _should_ disallow it because your value is not one of the allowed ones. What then? I hope you understand this decision.

PR #262, opened 2017-04-03
Update BaseField.js

rafinskipg:
Make field properties more important than schema properties

### DIFF
--- a/packages/uniforms/src/BaseField.js
+++ b/packages/uniforms/src/BaseField.js
@@ -264,8 +264,8 @@ export default class BaseField extends Component {
             parent,
             value,
 
-            ...props,
             ...schemaProps,
+            ...props,
             ...(explicitInitialValue || overrideValue) && {value},
 
             label,


rafinskipg:
In my case i was wanting to use conditional required value for a field. Maybe the way Custom works is the thing we need https://github.com/aldeed/meteor-simple-schema#custom

janowsiany:
Isn't it a breaking change?

rafinskipg:
Yea @janowsiany  maybe, and maybe it is not needed xD

I was just trying to make a field required in a conditional way but @radekmie  suggested a PR. Might it be enough using the `custom` attribute from simpleschema?
https://github.com/aldeed/meteor-simple-schema#make-a-field-conditionally-required

This should work, but it is not working.

janowsiany:
`required` property in uniforms is only dependent on `optional: true` from Simple Schema, take a look [here](https://github.com/vazco/uniforms/blob/fde6cefa7c998bb10d74486ea607c6608bc265c3/packages/uniforms/src/SimpleSchemaBridge.js#L167). 
I think that this conditinal requiredness might be handled in `Custom Field` which is IMO one biggest advantages of this package(makes it very flexible). Just quick scaffolding:
```js
const Custom => ({required, ...props}, {uniforms: {model}}) => (
  <TextField required={model.otherField !== 1} />
); 
```
This example should make field required only if value of `otherField` is not equal 1.
However, i might be wrong it was just my first tought that switching this order might be a breaking change, maybe some further investigation is needed.

rafinskipg:
@janowsiany With this code i should make one custom field for each type of fiedl i want to use with custom required... Sorry but the amount of work it makes it's not acceptable.

janowsiany:
I was completely wrong, but uniforms do support even custom validations sorry for this confusion.
@rafinskipg could you please share you `custom` validator?

rafinskipg:
This was working with aldeed autoforms. 

```
custom: function() {
            var shouldBeRequired = PlaceTypeEnum.hasPlaceId(this.siblingField('type').value);

            if (shouldBeRequired) {
                // inserts
                if (!this.operator) {
                    if (!this.isSet || this.value === null || this.value === "") return "required";
                }
                // updates
                else if (this.isSet) {
                    if (this.operator === "$set" && this.value === null || this.value === "") return "required";
                    if (this.operator === "$unset") return "required";
                    if (this.operator === "$rename") return "required";
                }
            }
        }
```

janowsiany:
Ok so here is a schema which works in [demo app](https://uniforms.tools/):
```js
new SimpleSchema({
    adult: {
        type: Boolean,
        optional: true,
        custom: function() {
            if (this.field('size').value === 'm') {
                return 'required';
            }
        }
    },

    size: {
        type: String,
        defaultValue: 'm',
        allowedValues: ['xs', 's', 'm', 'l', 'xl']
    },
})
```
Am i missing something?

radekmie:
Why did I want to see this PR? Because it's more reasonable, to make direct props more important than schema ones. I'm still thinking about consequences, but it's more like a bugfix, than a feature.

radekmie:
To be honest, I have very mixed feelings about it... I'd love to merge it, but it is a kind of breaking change... No matter what, it's still a kind of bug fix because that's how it should behave... How about this: let's release a new version _(`v1.17.1` at the moment)_ and see, how it works.

rafinskipg:
Probably it will be a breaking change yes. But, yeah.. makes sense. I have not tested that in our project so I don't know if anything is going to break. I suspect there will be major issues. Maybe it will affect more to people with custom fields.

radekmie:
OK, then I'll release `v1.17.1-beta.1` and ask few people to check it in their projects.

radekmie:
@rafinskipg I had to fix it, because after your changes `state` took precedence over `schemaProps`. You can take a look at it [here](https://github.com/vazco/uniforms/commit/4fc91894a1313d402ba66014de58f6f20471e44f). It's already released as `v1.17.1-beta.1`.

Issue #267, opened 2017-04-11
The errorMessage is shown in grey before validation fails


Gervwyk:
When using errorMessage prop, the error message is shown in grey before the validation fails on all themes. 
Using this schema: 
```
new SimpleSchema({
    STRING: {
         type: String,
         regEx: /^[A-Z]*$/,
         uniforms: {
             errorMessage: "ERROR here"
          }
     }
})
```

Do we really want to show the error before validation? It would be nice if we could have the error only show when validation fails or make it optional somehow.

radekmie:
Indeed, a lot of fields are checking for `errorMessage` instead of `error` - it's reasonable to change it, because it will allow customization like this. Would you submit a PR for it?

Gervwyk:
I'll make a PR later today. Do we want this optional? Or should we only show the message on error? If optional by what prop / value?

radekmie:
I see this in this way: [`uniforms-semantic/NumField`](https://github.com/vazco/uniforms/blob/master/packages/uniforms-semantic/src/NumField.js#L59-L63)

```diff
-        {!!(errorMessage && showInlineError) && (
+        {!!(error && showInlineError) && (
             <div className="ui red basic pointing label">
                 {errorMessage}
             </div>
         )}
```

Gervwyk:
Ok great. I agree. I'll change on all themes and submit PR.

PR #269, opened 2017-04-11
Change inline Error to show errorMessage

Gervwyk:
Solves #267 .

### DIFF
--- a/packages/uniforms-antd/src/ListField.js
+++ b/packages/uniforms-antd/src/ListField.js
@@ -11,6 +11,7 @@ import ListItemField from './ListItemField';
 
 const List = ({
     children,
+    error,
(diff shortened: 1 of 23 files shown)


Gervwyk:
I'm wondering what to do with `Errorsfield` . How do we change:
```
{schema.getErrorMessages(error).map((message, index) =>
                    <li key={index} style={{margin: '3px'}}>
                        {message}
                    </li>
)}
```
To only show `errorMessage` ? 
Also should the there be an option to hide `Errorsfield` when `showInlineError` is true? or should this be the case by default?

radekmie:
I think it's good as it is, because you can use `ErrorField` with custom content for such case.

PR #294, opened 2017-06-02
`showInlineError` is available in 'material' too

kubami:
Small docs update to reflect the code

### DIFF
--- a/API.md
+++ b/API.md
@@ -227,7 +227,6 @@ import BoolField from 'uniforms-unstyled/BoolField'; // Choose your theme packag
     //   bootstrap3
     //   bootstrap4
     //   semantic
-    //   material
     showInlineError={true}
 
     // Field and sourroundings wrap className.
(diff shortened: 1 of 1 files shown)


janowsiany:
I think that `showInlineError` is only affecting following fields: `DateField`, `LongTextField`, `NumField`, `SelectField`, `TextField` or am  i wrong? So please update this PR to match only those fields, also if you wish to add handling `showInlineError` in other type of fields then any contributions are welcome!

API.md
```diff
@@ -227,6 +227,7 @@ import BoolField from 'uniforms-unstyled/BoolField'; // Choose your theme packag
     //   bootstrap3
     //   bootstrap4
     //   semantic
+    //   material
```
janowsiany: `BoolField` is not handling `showInlineError`

API.md
```diff
@@ -477,6 +479,7 @@ import ListField from 'uniforms-unstyled/ListField'; // Choose your theme packag
...
```
janowsiany: The same with `ListField`

kubami:
Of course, you are right!

PR #303, opened 2017-06-28
301 bootstrap errorsfield styling

zeroasterisk:
https://github.com/vazco/uniforms/issues/301#issuecomment-310292017

bs4
bs3

PR #305, opened 2017-06-28
304 radios need space

zeroasterisk:
note review this after #303 so the other commits don't cause problems (I would have rebased, but needed 	86e51ca )

### DIFF
--- a/packages/uniforms-bootstrap4/src/RadioField.js
+++ b/packages/uniforms-bootstrap4/src/RadioField.js
@@ -25,6 +25,7 @@ const Radio = props =>
                         onChange={() => props.onChange(item)}
                         type="radio"
                     />
+                    {' '}
                     {props.transform ? props.transform(item) : item}
                 </label>
             </div>


demo/package.json
```diff
@@ -13,8 +13,11 @@
     "antd": "2.10.4",
     "babel-polyfill": "6.23.0",
     "graphql": "0.10.1",
+    "indexof": "0.0.1",
     "material-ui": "0.18.2",
+    "message-box": "0.0.2",
     "meteor-node-stubs": "0.2.11",
+    "moment": "^2.18.1",
```
radekmie: Same as in #303.

packages/uniforms-bootstrap4/src/ErrorsField.js
```diff
@@ -6,7 +6,7 @@ import nothing        from 'uniforms/nothing';
 
 const ErrorsField = ({className, children, ...props}, {uniforms: {error, schema}}) =>
     (!error && !children) ? nothing : (
-        <div className={classnames('card', className)} {...filterDOMProps(props)}>
+        <div className={classnames('card', 'card-outline-danger', 'text-danger', className)} {...filterDOMProps(props)}>
```
radekmie: Same as in #303.

radekmie:
Also, tests are failing.

zeroasterisk:
I dropped the un-related changes, and force pushed, only added the space.

PR #307, opened 2017-07-01
LongTextField placeholder => hintText

janowsiany:
Fixing LongTextField wrong prop name


radekmie:
@janowsiany: could you also add a note about it in the changelog?

Issue #335, opened 2017-08-22
Uncaught Error: Unsupported field type: Boolean  at invariant 


cesarve77:
if I use type Boolean in my graphql schema and AutoField i get this error in my browser : 
Uncaught Error: Unsupported field type: Boolean  at invariant 

example: 
```
const type=buildSchema(`
     type SignUp {
         name:          String!
         familyName:    String!
         email:         String!
         password:      String!
         terms:         Boolean!
     }
 # This is required by buildASTSchema
 type Query { anything: ID }
 `).getType('SignUp')


const schema = new GraphQLBridge(type, function (model) { .....
```
example :  

`  <AutoField       name="terms" type='checkbox'  label="I agree to adhere with  terms "/>`

I can solved this problem 
1) Or using Int in the schema
2) or  using CheckBox component instead of AutoField
3) or adding property component=CheckBox in AutoField

Maybe adding this line code to https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/GraphQLBridge.js

can solve the issue

` getType (name) {
        const fieldType = this.getField(name).type;

        if (fieldType instanceof graphql.GraphQLList)            return Array;
        if (fieldType instanceof graphql.GraphQLObjectType)      return Object;
        if (fieldType instanceof graphql.GraphQLInputObjectType) return Object;
        if (fieldType instanceof graphql.GraphQLScalarType) {
            if (fieldType.name === 'Int')    return Number;
            if (fieldType.name === 'Float')  return Number;
            if (fieldType.name === 'String') return String;
        ++if (fieldType.name === 'Boolean') return Boolean;
        }

        return fieldType;
    }
`

radekmie:
I'm surprised that it came out that late. Do you want to submit a PR with it?

cesarve77:
Sure. I will tomorrow i live in Australia and is time to sleep

PR #337, opened 2017-08-23
Add support for Boolean type in GraphQLBridge

cesarve77:
Refer to issue:
https://github.com/vazco/uniforms/issues/335

### DIFF
--- a/packages/uniforms-semantic/src/BoolField.js
+++ b/packages/uniforms-semantic/src/BoolField.js
@@ -37,10 +37,8 @@ const Bool = ({
         </div>
 
         {!!(error && showInlineError) && (
-            <div>
-                <div className="ui red basic pointing label">
-                    {errorMessage}
-                </div>
+            <div className="ui red basic pointing label">
+                {errorMessage}
             </div>
         )}
     </div>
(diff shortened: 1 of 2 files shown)


packages/uniforms-semantic/src/BoolField.js
```diff
@@ -37,8 +37,10 @@ const Bool = ({
         </div>
 
         {!!(error && showInlineError) && (
-            <div className="ui red basic pointing label">
-                {errorMessage}
+            <div>
+                <div className="ui red basic pointing label">
+                    {errorMessage}
+                </div>
```
radekmie: This will be merged in #339 so let's remove it from here.

packages/uniforms/src/GraphQLBridge.js
```diff
@@ -182,6 +182,7 @@ export default class GraphQLBridge extends Bridge {
             if (fieldType.name === 'Int')    return Number;
             if (fieldType.name === 'Float')  return Number;
             if (fieldType.name === 'String') return String;
+            if (fieldType.name === 'Boolean') return Boolean;
```
radekmie: Could you keep it in align with lines above?

Issue #338, opened 2017-08-23
Uniforms-Semantic: Wrong position for inline error in BoolField


cesarve77:
![captura de pantalla 2017-08-23 12 54 38](https://user-images.githubusercontent.com/3970983/29597071-abf2e8d6-8803-11e7-8c66-ff6d4c19b034.png)

PR #339, opened 2017-08-23
BoolField inline error position correction

cesarve77:
Fix for https://github.com/vazco/uniforms/issues/338


packages/uniforms/src/GraphQLBridge.js
```diff
@@ -182,6 +182,7 @@ export default class GraphQLBridge extends Bridge {
             if (fieldType.name === 'Int')    return Number;
             if (fieldType.name === 'Float')  return Number;
             if (fieldType.name === 'String') return String;
+            if (fieldType.name === 'Boolean') return Boolean;
```
radekmie: This will be merged in #337 so let's remove it from here.

PR #346, opened 2017-09-13
Fix TextField type prop for antd  (closes #345)

### LINKED ISSUE #345: Textfield type="password" not working for antd style
The `'uniforms-antd/TextField` does not properly pass the `type` prop. See below.

For demonstration, I imported both antd and unstyled versions. The antd does not mask the input field as expected.  

```jsx
import AutoForm from 'uniforms-antd/AutoForm';
import TextField from 'uniforms-antd/TextField';
import TextField2 from 'uniforms-unstyled/TextField';

<AutoForm schema={loginFormSchema} onSubmit={submitLogin}>
 <TextField name="email" />
 <TextField name="password" type="password" />
 <TextField2 name="password" type="password" />
 <SubmitField>Submit</SubmitField>
</AutoForm>
```

purplecones:
Should fix #345 

@radekmie Let me know if this PR is okay.

I didn't see a need to touch `filterDOMProps` especially since its being used by all other style versions. It seems it was just a missed prop.

### DIFF
--- a/packages/uniforms-antd/src/TextField.js
+++ b/packages/uniforms-antd/src/TextField.js
@@ -15,6 +15,7 @@ const Text = props =>
             placeholder={props.placeholder}
             ref={props.inputRef}
             value={props.value}
+            type={props.type}
             {...filterDOMProps(props)}
         />
     ))
--- a/packages/uniforms-antd/src/TextField.js
+++ b/packages/uniforms-antd/src/TextField.js
@@ -14,8 +14,8 @@ const Text = props =>
             onChange={event => props.onChange(event.target.value)}
             placeholder={props.placeholder}
             ref={props.inputRef}
-            value={props.value}
             type={props.type}
+            value={props.value}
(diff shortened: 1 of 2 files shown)


purplecones:
Also, I'm not quite sure how I would fully test this locally by importing my version of the library into a sample react app. Right now I'm purely going by the jest test.

@radekmie how do you test new changes locally in a react app?

packages/uniforms-antd/src/TextField.js
```diff
@@ -15,6 +15,7 @@ const Text = props =>
             placeholder={props.placeholder}
             ref={props.inputRef}
             value={props.value}
+            type={props.type}
```
radekmie: Could you keep it ordered?

radekmie:
Either by linking it to an existing app (symlink or ’npm link`) or in the demo app.

radekmie:
I'll release it this weekend.

PR #348, opened 2017-09-26
added checkbox option for BoolField for antd

purplecones:
PR adds ability to create an antd `Checkbox` with uniforms `BoolField` instead of just the antd `Switch`.

### Usage
```javascript
<BoolField
    checkbox
    name="myCheckbox"
    label="Check here if you agree"
/>
```

### DIFF
--- a/packages/uniforms-antd/__tests__/BoolField.js
+++ b/packages/uniforms-antd/__tests__/BoolField.js
@@ -1,19 +1,27 @@
 import React   from 'react';
 import Switch  from 'antd/lib/switch';
+import Checkbox  from 'antd/lib/checkbox';
 import {mount} from 'enzyme';
 
 import BoolField from 'uniforms-antd/BoolField';
 
 import createContext from './_createContext';
 
-test('<BoolField> - renders an input', () => {
+test('<BoolField> - renders a switch input', () => {
     const element = <BoolField name="x" />;
     const wrapper = mount(element, createContext({x: {type: Boolean}}));
 
     expect(wrapper.find(Switch)).toHaveLength(1);
 });
 
-test('<BoolField> - renders a input with correct id (inherited)', () => {
+test('<BoolField> - renders a checkbox input', () => {
+    const element = <BoolField checkbox name="x" />;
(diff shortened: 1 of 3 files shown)


packages/uniforms-antd/__tests__/BoolField.js
```diff
@@ -1,83 +1,155 @@
 import React   from 'react';
 import Switch  from 'antd/lib/switch';
+import Checkbox  from 'antd/lib/checkbox';
 import {mount} from 'enzyme';
```
radekmie: These should be aligned.

packages/uniforms-antd/src/BoolField.js
```diff
@@ -1,26 +1,38 @@
 import Icon           from 'antd/lib/icon';
 import React          from 'react';
 import Switch         from 'antd/lib/switch';
+import Checkbox       from 'antd/lib/checkbox';
 import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
```
radekmie: These should stay in order.

packages/uniforms-antd/src/BoolField.js
```diff
@@ -1,26 +1,38 @@
...
+            /> :
+            <Switch
+                checked={props.value}
+                disabled={props.disabled}
+                id={props.id}
+                name={props.name}
+                onChange={() => props.onChange(!props.value)}
+                ref={props.inputRef}
+                {...filterDOMProps(props)}
+            />
```
radekmie: How about `React.createElement(props.checkbox ? Checkbox : Switch, {...})`? It'll reduce this duplicated code.

packages/uniforms/src/filterDOMProps.js
```diff
@@ -2,6 +2,7 @@ const unwantedProps = [
     // These props are provided by BaseField
     'changed',
     'changedMap',
+    'checkbox',
```
radekmie: These are global, so it's a breaking change. I'd rather go with destructuring props in `BoolField` - it won't be needed then.

purplecones:
Thanks for feedback, learning something new every day.

PR #354, opened 2017-10-14
Functional set state

janowsiany:
This PR i aimed to use setState with a function only, all across `uniforms` package.

### DIFF
--- a/packages/uniforms/__tests__/ValidatedForm.js
+++ b/packages/uniforms/__tests__/ValidatedForm.js
@@ -33,16 +33,17 @@ describe('ValidatedForm', () => {
 
     describe('when reset', () => {
         it('removes `error`', () => {
-            const wrapper = mount(
-                <ValidatedForm model={model} schema={schema} onSubmit={onSubmit} />
-            );
-
             validator.mockImplementationOnce(() => {
                 throw new Error();
             });
 
+            const wrapper = mount(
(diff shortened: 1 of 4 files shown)


packages/uniforms/__tests__/ValidatedForm.js
```diff
@@ -6,11 +6,14 @@ import ValidatedForm from 'uniforms/ValidatedForm';
 jest.mock('meteor/aldeed:simple-schema');
 jest.mock('meteor/check');
 
-// FIXME: This fail sometimes when ran simultaneously.
 describe('ValidatedForm', () => {
     const onChange   = jest.fn();
     const onSubmit   = jest.fn();
-    const onValidate = jest.fn();
+    const onValidate = jest.fn((m, e, cb) => {
```
radekmie: Could you use full names here?

packages/uniforms/src/ValidatedForm.js
```diff
@@ -83,11 +83,9 @@ const Validated = parent => class extends parent {
         super.componentWillReceiveProps(...arguments);
 
         if (this.props.schema !== schema || this.props.validator !== validator) {
-            this.setState({
-                validator: this
-                    .getChildContextSchema()
-                    .getValidator(validator)
-            }, () => {
+            this.setState(({bridge = this.getChildContextSchema()}) => ({
```
radekmie: Is it needed?

PR #386, opened 2018-01-30
Update bootstrap to v4.0.0

janslowik:
Becouse the bootstrap released the final stable 4 version i decided to update `uniforms-bootstrap4` from alpha-5 to version 4.0.0. I tested mostly on `demo` app and tried to keep similar appearance and behaviour like with alpha version. I tested also with some more complicated schemas  and imho looks good, but maybe something i missed so i will be happy on every review or remark.

### DIFF
--- a/packages/uniforms-bootstrap4/src/BoolField.js
+++ b/packages/uniforms-bootstrap4/src/BoolField.js
@@ -11,6 +11,7 @@ const Bool = ({label, labelBefore, ...props}) =>
                 props.inputClassName,
                 'form-check',
                 `checkbox${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+                {'text-danger': props.error}
             )}
         >
             <label htmlFor={props.id} className="form-check-label">
--- a/packages/uniforms-bootstrap4/src/BoolField.js
+++ b/packages/uniforms-bootstrap4/src/BoolField.js
@@ -10,7 +10,7 @@ const Bool = ({label, labelBefore, ...props}) =>
             className={classnames(
                 props.inputClassName,
                 'form-check',
-                `checkbox${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+                `checkbox${props.inline ? ' custom-control-inline' : ''}`,
                 {'text-danger': props.error}
             )}
         >
--- a/packages/uniforms-bootstrap4/src/BoolField.js
+++ b/packages/uniforms-bootstrap4/src/BoolField.js
@@ -10,8 +10,11 @@ const Bool = ({label, labelBefore, ...props}) =>
             className={classnames(
                 props.inputClassName,
                 'form-check',
-                `checkbox${props.inline ? ' custom-control-inline' : ''}`,
-                {'text-danger': props.error}
+                'checkbox',
+                {
+                    'text-danger': props.error,
+                    'custom-control-inline': props.inline
+                }
             )}
         >
             <label htmlFor={props.id} className="form-check-label">
--- a/packages/uniforms-bootstrap4/src/ErrorField.js
+++ b/packages/uniforms-bootstrap4/src/ErrorField.js
@@ -6,7 +6,7 @@ import nothing        from 'uniforms/nothing';
 
 const Error = ({children, className, error, errorMessage, ...props}) =>
     !error ? nothing : (
-        <div className={classnames('card', className)} {...filterDOMProps(props)}>
+        <div className={classnames('card', 'mb-3', className)} {...filterDOMProps(props)}>
             <div className="card-body">
                 {children ? (
                     children
--- a/packages/uniforms-bootstrap4/src/ErrorField.js
+++ b/packages/uniforms-bootstrap4/src/ErrorField.js
@@ -7,7 +7,7 @@ import nothing        from 'uniforms/nothing';
 const Error = ({children, className, error, errorMessage, ...props}) =>
     !error ? nothing : (
         <div className={classnames('card', className)} {...filterDOMProps(props)}>
-            <div className="card-block">
+            <div className="card-body">
                 {children ? (
                     children
                 ) : (
--- a/packages/uniforms-bootstrap4/src/RadioField.js
+++ b/packages/uniforms-bootstrap4/src/RadioField.js
@@ -12,7 +12,7 @@ const Radio = props =>
                 className={classnames(
                     props.inputClassName,
                     'form-check',
-                    `radio${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+                    `radio ${props.inline ? ' custom-control-inline' : ''}`,
                     {'text-danger': props.error}
                 )}
             >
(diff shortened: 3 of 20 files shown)


packages/uniforms-bootstrap4/src/ErrorField.js
```diff
@@ -6,8 +6,8 @@ import nothing        from 'uniforms/nothing';
 
 const Error = ({children, className, error, errorMessage, ...props}) =>
     !error ? nothing : (
-        <div className={classnames('card', className)} {...filterDOMProps(props)}>
-            <div className="card-block">
+        <div className={classnames('card', 'mb-3', className)} {...filterDOMProps(props)}>
```
radekmie: Is this `mb-3` part really needed? We shouldn't enforce your layout.

It's also a few times more below.

packages/uniforms-bootstrap4/src/RadioField.js
```diff
@@ -12,7 +12,8 @@ const Radio = props =>
                 className={classnames(
                     props.inputClassName,
                     'form-check',
-                    `radio${props.inline ? '-inline' : ''}` // bootstrap4 < alpha.6
+                    `radio${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
```
radekmie: IMHO we should get rid of these alpha fixes.

radekmie:
@janslowik?

packages/uniforms-bootstrap4/src/ErrorField.js
janslowik: At version `4.0.0` `margin-bottom` was removed from some components. This `mb-3` class is added here to provide proper display the demo page. Without this submit buttons etc are stick to the form what imho looks bad.

packages/uniforms-bootstrap4/src/RadioField.js
janslowik: done.

packages/uniforms-bootstrap4/src/BoolField.js
```diff
@@ -10,7 +10,8 @@ const Bool = ({label, labelBefore, ...props}) =>
             className={classnames(
                 props.inputClassName,
                 'form-check',
-                `checkbox${props.inline ? '-inline' : ''}`, // bootstrap4 < alpha.6
+                `checkbox${props.inline ? ' custom-control-inline' : ''}`,
+                {'text-danger': props.error}
```
radekmie: In such case, use `'custom-control-inline': props.inline`.

packages/uniforms-bootstrap4/src/gridClassName.js
```diff
@@ -3,32 +3,59 @@ import filterDOMProps from 'uniforms/filterDOMProps';
...
+        }
+    }
+
+    if (a === 'lg') {
+        if (b === 'xl') {
+            return -1;
+        }
+    }
+
+    return 1;
```
radekmie: Do these have to be in order? If so, maybe let's have an array with correct order and order them by an index in this array: `sizes.indexOf(a) - sizes.indexOf(b)`?

packages/uniforms-bootstrap4/src/gridClassName.js
```diff
@@ -38,7 +65,14 @@ export default function gridClassName (grid, side) {
 
     // Example: {xs: 6, sm: 4, md: 3}
     if (typeof grid === 'object') {
-        return Object.keys(grid).map(size => gridClassNamePart(size, grid[size], side)).join(' ');
+
```
radekmie: Extra newline.

packages/uniforms-bootstrap4/src/gridClassName.js
```diff
@@ -38,7 +65,14 @@ export default function gridClassName (grid, side) {
...
+        if (!grid['xs']) {
+            grid['xs'] = grid['sm'] || grid['md'] || grid['lg'] || grid['xl'];
+        }
```
radekmie: Two things here:
* Should we care about it?
* If so, use dot notation instead.

packages/uniforms-bootstrap4/src/gridClassName.js
janslowik: It was intentionally.

packages/uniforms-bootstrap4/src/gridClassName.js
janslowik: ok.

Maybe?
```
const gridOrder = {
    xs: 1,
    sm: 2,
    md: 3,
    lg: 4,
    xl: 5
};

function compareSizeClass (a, b) {
    return gridOrder[a] - gridOrder[b];
}
```

packages/uniforms-bootstrap4/src/gridClassName.js
janslowik: Yes it is needed (becouse we use this `xs` to count the default `col-`.

radekmie: It doesn't mean it's needed - it means that it's required.

packages/uniforms-bootstrap4/src/gridClassName.js
```diff
@@ -38,7 +39,13 @@ export default function gridClassName (grid, side) {
 
     // Example: {xs: 6, sm: 4, md: 3}
     if (typeof grid === 'object') {
-        return Object.keys(grid).map(size => gridClassNamePart(size, grid[size], side)).join(' ');
+        if (!grid.xs) {
+            grid.xs = grid.sm || grid.md || grid.lg || grid.xl;
```
radekmie: It just hit me: you can't mutate `grid` here!

packages/uniforms-bootstrap4/src/gridClassName.js
```diff
@@ -38,7 +39,10 @@ export default function gridClassName (grid, side) {
 
     // Example: {xs: 6, sm: 4, md: 3}
     if (typeof grid === 'object') {
-        return Object.keys(grid).map(size => gridClassNamePart(size, grid[size], side)).join(' ');
+        const completeGrid = {xs: grid.sm || grid.md || grid.lg || grid.xl, ...grid};
```
radekmie: Move it into an `if`. This way we'd make a lot of unnecessary copies.

janslowik: Do you prefer
```
        if (!grid.xs) {
            const completeGrid = {xs: grid.sm || grid.md || grid.lg || grid.xl, ...grid};
            return Object.keys(completeGrid || grid)
                .sort(compareSizeClass)
                .map(size => gridClassNamePart(size, completeGrid[size], side)).join(' ');
        }

        return Object.keys(grid)
            .sort(compareSizeClass)
            .map(size => gridClassNamePart(size, grid[size], side)).join(' ');
```
or
```
const completeGrid = grid.xs
            ? grid
            : {xs: grid.sm || grid.md || grid.lg || grid.xl, ...grid};

        return Object.keys(completeGrid)
            .sort(compareSizeClass)
            .map(size => gridClassNamePart(size, completeGrid[size], side)).join(' ');
```

radekmie: I think our ESLint allows overriding the argument. If so, then

```js
if (!grid.xs) {
    grid = {xs: grid.sm || grid.md || grid.lg || grid.xl, ...grid};
}
```

If not, then the second one.

PR #388, opened 2018-02-05
Add JSON schema to demo app

janowsiany:
This is a small enhancement which presents the usage of `JSONSchemaBridge` inside demo app.

### DIFF
--- a/demo/imports/components/ApplicationPreviewField.js
+++ b/demo/imports/components/ApplicationPreviewField.js
@@ -14,13 +14,13 @@ class ApplicationPreview extends Component {
         this.state = {model: undefined};
 
         this.onModel = this.onModel.bind(this);
-        this._schema = eval(`${this.props.value.schema}`);
+        this._schema = eval(`(${this.props.value.schema})`);
     }
 
     componentWillReceiveProps (props) {
         if (this.props.value.schema !== props.value.schema) {
             this.onModel({});
-            this._schema = eval(`${props.value.schema}`);
+            this._schema = eval(`(${props.value.schema})`);
         }
     }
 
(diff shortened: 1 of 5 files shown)


demo/imports/components/ApplicationPreviewField.js
```diff
@@ -14,13 +14,13 @@ class ApplicationPreview extends Component {
...
         this.onModel = this.onModel.bind(this);
-        this._schema = eval(`(${this.props.value.schema})`);
+        this._schema = eval(`${this.props.value.schema}`);
     }
 
     componentWillReceiveProps (props) {
         if (this.props.value.schema !== props.value.schema) {
             this.onModel({});
-            this._schema = eval(`(${props.value.schema})`);
+            this._schema = eval(`${props.value.schema}`);
```
radekmie: This way in both cases you could eval `props.value.schema` directly.

demo/imports/lib/schema.js
```diff
@@ -21,17 +23,21 @@ MessageBox.defaults({messages: {en: {syntax: ''}}});
...
-    global.buildASTSchema = buildASTSchema;
-    global.parse          = parse;
+    global.Ajv              = Ajv;
+    global.GraphQLBridge    = GraphQLBridge;
+    global.JSONSchemaBridge = JSONSchemaBridge;
+    global.SimpleSchema     = SimpleSchema;
+    global.SimpleSchema2    = SimpleSchema2;
+    global.buildASTSchema   = buildASTSchema;
+    global.parse            = parse;
 }
```
radekmie: More and more of these... How about using `const scope = Meteor.isClient ? window : global;` and update this code accordingly?

demo/package.json
```diff
@@ -5,6 +5,7 @@
     "start": "meteor run"
   },
   "dependencies": {
+    "ajv": "^6.1.1",
```
radekmie: Pin it, please.

PR #391, opened 2018-02-15
Add extra check on Match

MaxGuitet:
When using uniforms in a Meteor/React project and testing with Jest, in the Jest environment Meteor is not defined. This results in an error being raised `Cannot read property 'Match' of undefined`. Even if you use `simpl-schema`, the file is being imported and fails on `SimpleSchemaBridge.js:38`.

### DIFF
--- a/packages/uniforms/src/SimpleSchemaBridge.js
+++ b/packages/uniforms/src/SimpleSchemaBridge.js
@@ -33,7 +33,7 @@ try {
     }
 } catch (_) { /* Ignore it. */ }
 
-if (SimpleSchema) {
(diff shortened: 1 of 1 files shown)


radekmie:
Hello @MaxGuitet. Why you've closed this PR?

MaxGuitet:
I realised it was probably more with the setup of the tests and that there was a workaround in jest. So didn't want to mess with the core package only for testing, but if you think it would be an improvement, we might want to reopen.

radekmie:
Then reopen it, please. We should check it as SimpleSchema doesn't depend on the `match` package.

PR #404, opened 2018-04-06
AntD: Customizable style on AutoField

webpn:
Making AutoField, and all other components enhanced via `wrapField`, customizable through a `style` prop.

### DIFF
--- a/packages/uniforms-antd/src/wrapField.js
+++ b/packages/uniforms-antd/src/wrapField.js
@@ -14,6 +14,7 @@ export default function wrapField ({
     labelCol,
     required,
     showInlineError,
+    style,
     wrapperCol
 }, children) {
     const labelNode = !!label && (
@@ -39,7 +40,7 @@ export default function wrapField ({
             label={labelNode}
             labelCol={labelCol}
             required={required}
-            style={{marginBottom: '12px'}}
+            style={style || {marginBottom: '12px'}}
             validateStatus={error ? 'error' : undefined}
             wrapperCol={wrapperCol}
         >


radekmie:
Thank you, @webpn!

radekmie:
@webpn: nope, my fault. I've published incorrect (old) version. Will do it in a moment.

_I have to bind the build to the publish script..._

webpn:
Yesterday I noticed an issue that affects my pull req: since I wanted to be able to customize the hardcoded style (`{ marginBottom: '12px' }`), I added a custom prop `style` on `wrapField`. 


However, the `style` prop is applied also to the wrapped `<Input />` element, so that, if you set a `border` in the `style` prop of the wrapper, you will get a border on the wrapper `<Form.Item />`, and the same border will be applied on the wrapped `<Input />` element.

You can see an example on [this CodeSandbox](https://codesandbox.io/s/n48zz5397l).

I didn't notice this issue before because I was testing only with `margin` style customization (setting `margin: 0`).

How can I fix this? I thought it would be better if the prop on the wrapper function was called `wrapperStyle`, rather than `style`.

radekmie:
Well, @webpn, it's a problem. You can submit a PR for it, but I have to think about it a bit more: it's a breaking change. Not a big deal, as there wasn't such prop for a long time and there is, which is also a breaking change, but it's a new feature, so semantic versioning allows it.

webpn:
Ok, I created https://github.com/vazco/uniforms/pull/414
I had the same concern about the fact that renaming the style prop could be considered a breaking change: that's why I preferred to discuss this before making any other PR.
But, since #404 introduced a bug (that is a potentially breaking change), maybe its fix won't be considered as a breaking change itself. #414 may be released in 1.24.3

PR #425, opened 2018-05-15
date field: support dateFormat to parse string

MohammedEssehemy:
fix error value.toISOString is not a function when passing string.


MohammedEssehemy:
Hi @radekmie 
We use uniforms  to update  JSON data coming from server
so we get date values as string and when passing this to uniforms  we get the error describned above

radekmie:
Then you should adjust the model before passing it to the form or use `modelTransform` ([documentation](https://github.com/vazco/uniforms/blob/master/INTRODUCTION.md#model-transformations)).

PR #440, opened 2018-06-21
Use the official wording :)

oliviertassinari:


radekmie:
Good one, @oliviertassinari! :smile:

PR #442, opened 2018-06-26
Drop dependency on react-tap-event-plugin

janowsiany:
This dependency is not needed anymore. Used to be required until [`material-ui@0.20.0`](https://github.com/mui-org/material-ui/releases/tag/v0.20.0)


radekmie:
Already included in #443.

Issue #433, opened 2018-06-04
meteor 1.7.0.1 - TypeError: Class constructor _class cannot be invoked without 'new'


macrozone:
this line fails in meteor 1.7.0.1:

```
import AutoFormBase from 'uniforms/AutoForm';

const Auto = parent => class extends AutoFormBase.Auto(parent) {
  static Auto = Auto;
};

```

any idea?

full error:

```
modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:98346 Uncaught (in promise) TypeError: Class constructor _class cannot be invoked without 'new'
    at new _class (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:98346)
    at new _class (AutoForm.js:9)
    at modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80870
    at measureLifeCyclePerf (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80651)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80869)
    at ReactCompositeComponentWrapper._constructComponent (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80860)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80763)
    at Object.mountComponent (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:73688)
    at ReactCompositeComponentWrapper.performInitialMount (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80946)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=441a59febac0d39753b5eebcca0bbdcb08b83a80:80833)
```

radekmie:
It's probably related to different versions of Babel used by uniforms and Meteor. While I'm not sure, I'd ask for a reproduction too (we can track it in the Meteor issue if you want).

macrozone:
@radekmie does your example project work under meteor 1.7? did not have time to setup a reproduction yet

radekmie:
Haven't tried it yet. I'll give it a shot later this week.

radekmie:
Unluckily, I've made two attempts and failed twice. Any progress on your site?

macrozone:
[...]

PR #443, opened 2018-06-26
Switched to Babel 7 and updated dependencies.

radekmie:
It's a big one.

Resolves: #433.
Supersedes: #442.


macrozone:
Looking good! 🤩

PR #460, opened 2018-08-30
Fix extracting the type from GraphQLNonNull in getProps

jgoux:
The internal type stored in `GraphQLNonNull` was never extracted as the whole `field` needs to be passed according to `extractFromNonNull` definition.

### DIFF
--- a/packages/uniforms/__tests__/GraphQLBridge.js
+++ b/packages/uniforms/__tests__/GraphQLBridge.js
@@ -15,7 +15,8 @@ describe('GraphQLBridge', () => {
         input Author {
             id:        ID!
             confirmed: Boolean
-            decimal:   Float!
+            decimal1:  Float
+            decimal2:  Float!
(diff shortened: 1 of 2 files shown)


radekmie:
It should return the underlying type, so `GraphQLScalarType` and friends. Can we start with adding the tests which should fail in the current implementation? Also, I don't clearly understand, when the current code is not working - could you provide an example?

radekmie:
Huh, I'm not sure anymore. I'm not really sure about _what the implementation should be_, but I think that two things are important here: one, we need to pass `fieldType` in the same manner, two, cases like `decimal` should be working correctly.

I'll get to it later today.

radekmie:
OK, so:

```js
const fieldType = extractFromNonNull(field).type;
```

Seems to work. Can you check if it works for you too?

jgoux:
It works! 👍

PR #472, opened 2018-09-26
Added catch-all to validation promises (closes #471)

### LINKED ISSUE #471: Using the error prop causes console errors on validations
When using the `error` prop with a `ValidatedForm`, `ValidatedQuickForm` and `AutoForm`, an uncaught promise error is logged to the console whenever the model is validated after a submit (such as when manipulating model data).

Demo: https://codesandbox.io/s/lx4rz1001q

The problem seems to lie in [`ValidatedForm.onValidateModel`](https://github.com/vazco/uniforms/blob/30c37a13343261846464fc749d1570a721864ea9/packages/uniforms/src/ValidatedForm.js#L143). Whenever the `error` prop is set or schema validation throws, it inevitably leads the returned promise to reject , and both calls to `this.onValidate()` and `this.onValidateModel()` in [`componentWillReceiveProps`](https://github.com/vazco/uniforms/blob/30c37a13343261846464fc749d1570a721864ea9/packages/uniforms/src/ValidatedForm.js#L69) are unhandled, causing the console error.

```
componentWillReceiveProps ({model, schema, validate, validator}) {
    super.componentWillReceiveProps(...arguments);

    if (this.props.schema !== schema || this.props.validator !== validator) {
        this.setState(({bridge}) => ({
            validator: bridge.getValidator(validator)
        }), () => {
            if (validate === 'onChange' || validate === 'onChangeAfterSubmit' && this.state.validate) {
                this.onValidate();
            }
        });
    } else if (!isEqual(this.props.model, model)) {
        if (validate === 'onChange' || validate === 'onChangeAfterSubmit' && this.state.validate) {
            this.onValidateModel(model);
        }
    }
}

onValidateModel (model) {
    model = this.getModel('validate', model);

    let catched = this.props.error || null;
    try {
        this.state.validator(model);
    } catch (error) {
        catched = error;
    }

    return new Promise((resolve, reject) => {
        this.props.onValidate(model, catched, (error = catched) => {
            // Do not copy error from props to state.
            this.setState(() => ({error: error === this.props.error ? null : error}), () => {
                if (error) {
                    reject(error);
                } else {
                    resolve();
                }
            });
        });
    });
}
```

idmadj:


radekmie:
Thank you @idmadj!

idmadj:
No worries, thanks for the quick merge!

PR #480, opened 2018-10-31
Support date and datetime-local inputs

Floriferous:
It'd be helpful for DateField to also support `type: 'date'` inputs, instead of only `datefield-local`. Currently, you can change the type, but the data does not appear, as it has to be formatted properly: `YYYY-MM-DD`.
 
I propose a working, though not very elegant solution, that checks for `inputProps` being passed to the material-ui `Input` component. [(source)](https://stackoverflow.com/questions/14647290/html-5-date-field-shows-as-mm-dd-yyyy-in-chrome-even-when-valid-date-is-set)

So to use this you have to do (with simple schema):

```js
new SimpleSchema({
  dateField: { type: Date, uniforms: { inputProps: { type: 'date' } } },
  dateTimeField: { type: Date },
})
```

It works in chrome, haven't tested it in other browsers yet.

### DIFF
--- a/packages/uniforms-material/src/DateField.js
+++ b/packages/uniforms-material/src/DateField.js
@@ -7,7 +7,8 @@ import connectField   from 'uniforms/connectField';
 import filterDOMProps from 'uniforms/filterDOMProps';
 
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
-const dateFormat = value => value && value.toISOString().slice(0, -8);
+const dateFormat = value => value && value.toISOString().slice(0, -14);
+const dateTimeFormat = value => value && value.toISOString().slice(0, -8);
 const dateParse = (timestamp, onChange) => {
     const date = new DateConstructor(timestamp);
     if (date.getFullYear() < 10000) {
@@ -48,7 +49,7 @@ const Date = ({
             placeholder={placeholder}
             ref={inputRef}
             type="datetime-local"
-            value={dateFormat(value)}
+            value={props.inputProps && props.inputProps.type === 'date' ? dateFormat(value) : dateTimeFormat(value)}
             {...filterDOMProps(props)}
         />
         {showInlineError && error ? (


radekmie:
Hi @Floriferous. The code itself is good. I'm just not sure if this is the right way. Right now, each theme consequently implements _some_ date picker - such change tends to implement more and more full-blown ones for each theme. It's more of an opinion thing, just let me think about it.

Also, what about `time`? It's also _easy_ achievable, in the same way. On the other way, it's easy to do so with `modelTransform` - have you thought about it in your case?

Floriferous:
You're right, I agree it doesn't feel like the right way to do this. 

Ideally, I should be able to simply specify what html `type` this is, and then it should work. But with the required `formatDate` transform you apply on `value` it gets a bit more complex : /

radekmie:
Ping @Floriferous.

Floriferous:
How do you suggest we do this:

* Add support for native `type='date'`, `type='datetime-local'`, and `type='time'`, even though we're manipulating the value with custom functions
* Custom type: `time, date, date-time`
* Other idea?

The schemas will probably have to be defined as follows:

```js
new SimpleSchema({
  dateField: { type: String, uniforms: { type: 'date' } },
  dateTimeField: { type: String, uniforms: { type: 'time' } },
});
```

Or should we store them as `Date` types, without date (`hh:mm`)/without time respectively (`dd-mm-yyyy`)?

radekmie:
As a rule, custom component is the best here - it can handle the in (model ~> DOM) and out (DOM ~> model) logic, parametrized with props (defined in the schema or directly in props). Now, we'd like to enhance the current flaky implementation _a bit_ to support also `date`. It's not covered on purpose, because one, there's a ton of **better** date/date-time/time pickers, two, [it's not really cross-browser](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#Browser_compatibility).

We have to chose now, if we want to support more than one as well or not. IMO `datetime-local` (currently supported) and `date` (this PR) are enough. If so, we have to do it in **every theme**.

There's even a special paragraph on that, on [Material-UI docs](https://material-ui.com/demos/pickers/#date-pickers):

> We are currently falling back to native input controls. If you are interested in implementing or have implemented a rich Material Design Picker with an awesome UX, please, let us know on mui-org/material-ui#4787 and mui-org/material-ui#4796! We could add a link to or a demo of your project in the documentation. Here are some components that are promising:
> 
> [material-ui-pickers](https://github.com/dmtrKovalenko/material-ui-pickers): date pickers and time pickers.
> [material-ui-time-picker](https://github.com/TeamWertarbyte/material-ui-time-picker): time pickers.
> [material-ui-next-pickers](https://github.com/chingyawhao/material-ui-next-pickers): date pickers and time pickers.
> 
> ⚠️ Native input controls support by browsers isn't perfect.

All in all, I'd rather **not** do it and always suggest to use an existing picker or create a custom field for that.

Floriferous:
Oh, so should we just close this PR then?

radekmie:
I thought about it once again and yes, I'm closing.

Issue #481, opened 2018-11-01
Allow passing props to material-ui's InputLabel


Floriferous:
We're using the material-ui variant of uniforms, and we've had the issue where we want to use `displayEmpty` of the Select, which conflicts with the InputLabel, as can be seen in this issue: https://github.com/mui-org/material-ui/issues/12336

The `shrink` prop on `InputLabel` would fix it, but there's currently no way to customize it, as all `SelectField` props are spread on the `Select` component only.

Not sure what's best, picking all props that should be on InputLabel (`shrink` `required` `disableAnimation` for example), or spread a new prop `labelProps` ?

radekmie:
Hi @Floriferous. Sounds good, as long as each `InputLabel` will be handled, not only the `SelectField` one. In this case, `labelProps` would be better, as there might be conflicts in prop names. Would you like to submit a PR for that?

radekmie:
Closed in #485.

PR #485, opened 2018-11-08
Pass labelProps to DateField and SelectField

Floriferous:
Addresses #481. 

It turns out there was only `DateField` and `SelectField` that can use the extra `labelProps`. The `TextField` component, used in most places, can already get a `InputLabelProps` prop that will be passed to its `InputLabel`.

For the sake of consistency, I changed the `FormLabel` to `InputLabel` on `DateField`, as I don't think it is required. I'll revert it if there was a specific reason behind it.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [Table of Contents](#table-of-contents)
 - [Fields](#fields)
     - [`AutoField`](#autofield)
     - [`AutoFields`](#autofields)
@@ -325,6 +326,11 @@ import DateField from 'uniforms-unstyled/DateField'; // Choose your theme packag
     //   receive a field component rather than an HTML input. If you need an input ref,
     //   use this prop instead.
     inputRef={ref => {}}
+
+    // Props for the FormLabel
+    // Available in:
+    //   material-ui
+    labelProps={{ shrink: true, disableAnimation: true }}
 
     // Maximum value.
     //   Date object.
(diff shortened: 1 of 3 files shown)


packages/uniforms-material/src/SelectField.js
```diff
@@ -43,6 +43,7 @@ const renderSelect = ({
     showInlineError,
     transform,
     value,
+    labelProps,
```
radekmie: Please keep these ordered.

packages/uniforms-material/src/DateField.js
```diff
@@ -1,6 +1,6 @@
 import FormControl    from '@material-ui/core/FormControl';
 import FormHelperText from '@material-ui/core/FormHelperText';
-import FormLabel      from '@material-ui/core/FormLabel';
+import InputLabel     from '@material-ui/core/InputLabel';
 import Input          from '@material-ui/core/Input';
```
radekmie: Please keep these ordered.

API.md
```diff
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [Table of Contents](#table-of-contents)
```
radekmie: It's a good addition. Leave it as it is now, but it's better not to mix purposes of PR.

Floriferous: Oh, vscode did this automatically, didn't mean to do that.

Floriferous:
Should be ok.

PR #496, opened 2019-01-01
Add ability to pass variant prop to FormControl. Fixes #495

### LINKED ISSUE #495: Ability to pass variant prop to Material-UI SelectField to support other field types
Material-UI 3 now supports Selects with both the new Outlined and Filled text field variants when used with [Select React component](https://material-ui.com/api/select/), see [demo](https://material-ui.com/demos/selects/#simple-select).

However, it does not seem that Uniforms MUI SelectField supports ability to pass in the `variant` prop such as `variant="outlined"` like you can with `fullWidth` or `required` props. The MUI [FormControl component](https://material-ui.com/api/form-control/) supports the `variant` prop so maybe just adding this prop to `SelectField` with logic for `input` is what is required. 

If there is another way to support using these field variants today please let me know or if this will require a PR.

janowsiany:
Fixes: #495 
Introduces `wrapField` helper to reduce code duplication. This is similar to other packages.

### DIFF
--- a/packages/uniforms-material/__tests__/wrapField.js
+++ b/packages/uniforms-material/__tests__/wrapField.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import FormControl from '@material-ui/core/FormControl';
+import FormHelperText from '@material-ui/core/FormHelperText';
+import {mount} from 'enzyme';
+
+import wrapField from 'uniforms-material/wrapField';
+
+test('<wrapField> - renders wrapper', () => {
+  const element = wrapField({}, <div />);
+  const wrapper = mount(element);
+
+  expect(wrapper.find(FormControl)).toHaveLength(1);
+});
+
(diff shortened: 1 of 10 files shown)


radekmie:
Hi @janowsiany! I'd like to merge it, really, but `React.Fragment` is a `16.2.0` feature, uniforms are still supporting [`0.14.0`](https://github.com/vazco/uniforms/blob/master/packages/uniforms/package.json#L26). Maybe it's crazy, to go _that far_, but, at least for now, 16 is a maximum.

radekmie:
@janowsiany, @skirunman I've changed the way how `wrapField` works. There's no need for `React.Fragment` in such case. If you both are good with that, I'll merge it and release a new version.

I've also pushed to an incorrect origin: ignore `continuous-integration/travis-ci/push`, only `continuous-integration/travis-ci/pr` is valid. Codecov seems to work (but the status is not updated).

PR #498, opened 2019-01-04
Stop using deprecated "raised" variant, use "contained" instead

Floriferous:
Material-ui has started throwing warnings when using the old `raised` variant.

### DIFF
--- a/packages/uniforms-material/package.json
+++ b/packages/uniforms-material/package.json
@@ -22,7 +22,7 @@
     "src/"
   ],
   "peerDependencies": {
-    "@material-ui/core": "^3.0.0 || ^1.0.0",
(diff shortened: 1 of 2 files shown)


radekmie:
Hi @Floriferous. Is there a way to actually check the version of `@material-ui/core`? The `contained` variant was introduced in `1.2.0`, so a dependency range have to be changed.

Floriferous:
Not sure how to do that? But since 1.2.0 was released 7 months ago, can't you put a hard constraint on 1.2.0+ today? There's no hard upgrades since this package already uses 1.0, so I don't think there's a good reason for people not to follow material-ui's updates ?

radekmie:
OK then, let's change the dependency version to start from 1.2.0.

PR #509, opened 2019-02-01
Use material-ui icons to make IconButtons have a square aspect-ratio

Floriferous:
I've always been annoyed by the list add-del fields in uniforms-material, because they aren't perfectly circular.

Using the real material-ui icons solves this issue.

We're actually using other icons, which I believe are even cleaner, but maybe that's going too far:

screenshot 2019-02-01 at 14 19 56

### DIFF
--- a/packages/uniforms-material/package-lock.json
+++ b/packages/uniforms-material/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "uniforms-material",
-	"version": "1.30.0",
+	"version": "1.31.0",
 	"lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


radekmie:
Hi @Floriferous. It's a design decision, that this theme is not using icons by default. Even I myself had a project with custom icons, where this theme worked really well. As I understand, it's a not clear thing, I'm still against this change.

As a _dirty_ workaround, you can override the `defaultProps` directly in your project. Or, a cleaner way, pass these props explicitly (e.g. with a schema default).

PR #514, opened 2019-02-20
Fix SelectField required state. Fixes #512

### LINKED ISSUE #512: uniforms-material SelectField not displaying "required" asterisk on label
Using SelectField in uniforms-material I try to pass the `required` prop to show an asterisk on the label. However, it is not displayed. Maybe`required` is not being passed correctly to [underlying MUI TextField](https://github.com/vazco/uniforms/blob/master/packages/uniforms-material/src/SelectField.js#L55).

I verified that both `select` and `required` props are passed correctly to MUI TextField on [MUI doc website](https://material-ui.com/demos/text-fields/), which does display the asterisk correctly on the label. May want to double check this works correctly with other fields built on MUI TextField like DateField.

SelectField was refactored in PR https://github.com/vazco/uniforms/pull/505.

janowsiany:
Fixes #512

### DIFF
--- a/packages/uniforms-material/__tests__/SelectField.js
+++ b/packages/uniforms-material/__tests__/SelectField.js
@@ -28,6 +28,14 @@ test('<SelectField> - renders a Select with correct disabled state', () => {
   expect(wrapper.find(FormControl).prop('disabled')).toBe(true);
 });
 
+test('<SelectField> - renders a Select with correct required state', () => {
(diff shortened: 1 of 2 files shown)


skirunman:
@janowsiany Does [DateField](https://github.com/vazco/uniforms/blob/master/packages/uniforms-material/src/DateField.js) also need explicit `required` prop or is it OK?

PR #525, opened 2019-04-15
use transform for keys in material view

jansenignacio:
If you pass Objects as options, you'll get this error. This is because of `key={item}` which assigns an Object to the key which translates to `[object Object]`.

``Warning: Encountered two children with the same key, `.$.1=2$[object Object]`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.``

### DIFF
--- a/packages/uniforms-material/src/RadioField.js
+++ b/packages/uniforms-material/src/RadioField.js
@@ -40,7 +40,7 @@ const Radio = ({
       {allowedValues.map(item => (
         <FormControlLabel
           control={<RadioMaterial {...filteredProps} />}
-          key={item}
+          key={transform ? transform(item) : item}
           label={transform ? transform(item) : item}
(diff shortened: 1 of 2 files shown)


radekmie:
Hi @jansenignacio. I'm OK with the code itself (therefore the accept), but I'd like to discuss a bit the change itself, as I'm not sure about it. It's not a common scenario, but the labels (`transform` results) can be identical (and IMO it's more possible than a scenario with same values). A good example here is a dropdown with people data - their IDs will differ, but names can be the same. In the end, I'd rather suggest using direct data as the values (like IDs or even indexes) or creating a custom component for that. Thoughts?

radekmie:
In the end, there's a big problem with this solution and no response so far. I'm closing.

Issue #528, opened 2019-04-19
no input possible in <TextField> when wrapped inside Pose component


monkeydri:
[Pose](https://popmotion.io/pose) is a  library to make animations. It uses props passed to Pose components as triggers for styles. Pose components are created from standard HTML and React components with the `posed` method.

```js
import posed from 'react-pose';
const Field = posed.li({ open: { opacity: 1 }, closed: { opacity: 0 }});
```
When wrapped inside a Pose component such as `Field`, `TextField` does not allow keyboard input, it works ok for `NumField`.

```js
import SimpleSchema from 'simpl-schema';
import AutoForm from 'uniforms-unstyled/AutoForm';
import NumField from 'uniforms-unstyled/NumField';
import TextField from 'uniforms-unstyled/TextField';

const schema = new SimpleSchema(
{
	delay:
	{
		type: Number,
		min: 0,
		max: 62
	},
	email:
	{
		type: String,
	}
});
```

```jsx
<AutoForm schema={schema}>
	<Field>
		<NumField name="delay"/>
	</Field>
	<Field>
		<TextField name="email"/>
	</Field>
</AutoForm>
```

If putting `<TextField>`s as direct children of `<AutoForm>` both form works well.

Tried several things but I am starting to lose my hair on this issue...

**edit** : I updated the issue to narrow the issue to `<TextField>` rather than `<AutoField>`. Original title : no input possible in <AutoField> of type text when wrapped inside Pose component.

monkeydri:
from React Dev tool the difference between the two corresponding `<input>`s (for `<NumField>` and `<TextField>`) is that the `onChange` method does not seem bounded for `<TextField>` :

- `<TextField>` => onChange: onChange()
- `<NumField>` => onChange: bound onChange()

monkeydri:
After looking at the source code for [TextField](https://github.com/vazco/uniforms/blob/master/packages/uniforms-unstyled/src/TextField.js) and [NumField](https://github.com/vazco/uniforms/blob/master/packages/uniforms-unstyled/src/NumField.js) I start to wonder if this is not related to class-properties.

I am using [CRA](https://github.com/facebook/create-react-app) and class-properties [are not enabled](https://github.com/facebook/create-react-app/issues/4648) for 3rd party code. Therefore I had to import using relative module path (see #518) :

> use import X from 'uniforms/X' instead of import {X} from 'uniforms'

monkeydri:
[...]

PR #529, opened 2019-04-19
refactor TextField with state

monkeydri:
This fix issue #528 

It adds internal state to TextField so it updates even when wrapped inside Pose component (see issue for more details).


radekmie:
Closing as stated in #528.

PR #533, opened 2019-04-24
fix: Datefield prints date in wrong timezone

macrozone:
This is a suggestion how to fix https://github.com/vazco/uniforms/issues/532. I only added it to unstyled, maybe we introduce helpers for these functions?


radekmie:
As you can see, the tests are failing. I gave it a big thought and I'm still against it. It's a design choice to treat them as UTC. Also, with this change SSR is _not happy_, as the server might have a different timezone and the rehydration will have a different value.

As it's easy to create a custom component for that (or even a wrapper), I'd suggest doing so. It also requires to be mentioned in the docs.

I'll give it one more thought and let you know.

radekmie:
Sorry, it won't work. As I've said, we've tested it with the SSR and it is a problem.

PR #535, opened 2019-05-15
Nesting Arrays breaks props from schemaMetadata

gerwinbrunner:
Hi,

when you try to create labels for nested arrays in a schema the replacement code broke. Setting the regex replacement to global fixes it. 

This did not work prior to the fix since only the first index, in this case, the `0` is replaced - the second `3` is not touched.
` .getProps("arrayFieldA.0.innerArray.3.description")`
but works after the fix.


gerwinbrunner:
Fixed by PR
https://github.com/vazco/uniforms/issues/536

PR #537, opened 2019-05-15
chore: update docusaurus

endiliey:
## Motivation

Update to alpha.15

You might want to pull this PR and try it out locally first. I might have missed something

Anyway I recommend not to customize a lot since we're still experimenting a lot with the API. Try to use preset-classic and not install the plugin one by one because preset-classic will most likely be stable and what everyone will be installing

working


radekmie:
Thanks @endiliey. I'm about to check it out.

PR #543, opened 2019-06-12
Material ui v4

### LINKED ISSUE #542: Material-UI v4
Congrats on the v2 launch!

Anything holding this package back from using material-ui v4?

Floriferous:
Here's the migration checklist: https://material-ui.com/guides/migration-v3/

It doesn't seem to me that this package uses almost any of the changed features. I fixed a couple tests, only the target had to change.

Also, instead of pinning material-ui I let it loose, they have followed semver very well with proper deprecations before any braking changes in their major versions. Since they regularly add new features in their minor versions, that will give users of this package a bit more flexibility.

Closes #542.


Floriferous:
Alright, I updated the package requirements, any reason to undo the top-level `package.json` version changes?

radekmie:
Yeah, actually two: one, this change has almost completely rewritten the `package-lock.json` file (even not related to this update), two, as a rule, each change should be minimal. Also, this change added two not needed (and problematic in a monorepo) files, namely `website/package-lock.json` and `packages/uniforms-material/package-lock.json`.

Floriferous:
Okay, just so I get this right, I undo all the changes (including tests, since the ones I fixed depend on `material-ui@4`), except the `package.json` in `uniforms-material/package.json` ?

radekmie:
Updating package dependencies is enough. If you’d like to add these tests too, then leave the top level dependencies updated but update the lock file as little as possible. I can do it later today.

radekmie:
Done. I had to update `enzyme` and adapter too, because of airbnb/enzyme#2146.

PR #548, opened 2019-07-02
Add support for label classes

bigchirv:
For being able to alter the looks of the label without depending on 'grid'.


radekmie:
Hi @bigchirv. Really small change looks good (docs!). Can you add it to the bootstrap3 theme too? A simple test would also be great.

bigchirv:
Thanks for pointing out writing the tests, really forgot about them. :smiley:

On what file should I add the documentation on the labelClassName prop for bootstrap 3/4?

radekmie:
It should be in [`docs/api-fields.md`](https://github.com/vazco/uniforms/blob/master/docs/api-fields.md).

bigchirv:
Okay, I added the entries in `api-fields.md` as suggested.

radekmie:
Thank you @bigchirv!

PR #565, opened 2019-07-31
Documentation fixes

kestarumper:
- Fixed bug which caused to crash website after clicking submit button in RangeField example (model was undefined in validation before first input).
- Added default date in RangeField example.
- Set label to be always shrunk when Material-UI  is selected as a theme (before it would fall and overlap with date picker text).
- Unified emojis displayed in comparison matrices.


docs/examples/CustomFields/RangeField.js
```diff
@@ -37,6 +53,9 @@ export default function ExamplesRangeField() {
 
   return (
     <AutoForm
+      model={{
+        range: { start: today, stop: tommorow }
+      }}
```
radekmie: It'd be better to create the model only once, to keep the reference. Also, using a fixed pair of dates (not _today_, _tomorrow_) will simplify it a bit.

PR #574, opened 2019-08-10
Call to action

kestarumper:
image
image

### DIFF
--- a/website/components/Home/CallToAction/CallToAction.module.css
+++ b/website/components/Home/CallToAction/CallToAction.module.css
@@ -1,0 +1,21 @@
+.container {
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  margin-bottom: 2rem;
+}
+
+#get-started.container {
+  margin-top: -1.5rem;
+  margin-bottom: 3rem;
+}
+
+.questions {
+  background-color: var(--ifm-footer-background-color);
+  padding: 5rem 0;
+  margin: 0 0;
+}
+
+.questions h6 {
+  margin: -1em 0 1em 0;
+}
--- a/website/components/Home/CallToAction/index.js
+++ b/website/components/Home/CallToAction/index.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import Link from '@docusaurus/Link';
+import classNames from 'classnames';
+
+import styles from './CallToAction.module.css';
+
+const Button = ({ title, to, ...rest }) => {
+  const classes = Object.keys(rest).map(key => `button--${key}`);
+  return (
+    <Link to={to}>
+      <button class={classNames('button', ...classes)}>{title}</button>
+    </Link>
+  );
+};
+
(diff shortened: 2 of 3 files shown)


website/components/Home/CallToAction/CallToAction.module.css
```diff
@@ -0,0 +1,43 @@
...
+
+.questions {
+    background-color: #eee;
+    padding: 5rem 0;
+    margin: 0 0;
+}
+
+.questions h6 {
+    margin: -1em 0 1em 0;
+}
```
radekmie: Add a trailing newline at the end of files. As the amount of CSS in here is growing, maybe we should configure [`stylelint`](https://github.com/stylelint/stylelint)?

website/components/Home/CallToAction/index.js
```diff
@@ -0,0 +1,56 @@
...
+  </section>
+);
+
+export function SectionButton({ title, vertical }) {
+  return (
+    <Section
+      style={{
+        marginTop: '-3rem',
+        marginBottom: '4rem'
+      }}
```
radekmie: Why these two styles are inline?

website/components/Home/CallToAction/index.js
```diff
@@ -0,0 +1,56 @@
...
+      className={vertical && styles.vertical}
+    >
+      <div>
+        <Link to="/docs/installation">
+          <Button title={title} large />
+        </Link>
+      </div>
+    </Section>
```
radekmie: It makes a `section > div > a > span`. Can't we simplify it?

website/components/Home/CallToAction/index.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function SectionQuestions({ title, subtitle }) {
+  return (
+    <Section className={styles.questions}>
+      <div style={{ textAlign: 'center' }}>
+        {title && <h2>{title}</h2>}
+        {subtitle && <h6>{subtitle}</h6>}
+        <a href="https://gitter.im/vazco/uniforms">
+          <Button title="GITTER" large outlined primary />
```
radekmie: It's "Gitter", not "GITTER".

website/docusaurus.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
       apiKey: '513292b2605e0bc7bfdf63163dfde1c4',
       indexName: 'uniforms'
     },
-    googleAnalytics: { trackingID: 'UA-136559762-4' },
+    gtag: { trackingID: 'GTM-M8RKXLG' },
```
radekmie: I've removed commits introducing that. Can you get rid of them too?

website/components/Home/CallToAction/index.js
kestarumper: Because otherwise I would have to put them in separate class but these are only needed in this specific single situation.

PR #575, opened 2019-08-12
Add stylelint

kestarumper:


.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "stylelint-config-recommended",
+  "rules": {
+    "indentation": 2,
+    "no-missing-end-of-source-newline": true
```
radekmie: I'd rather use [`prettier/stylelint-prettier`](https://github.com/prettier/stylelint-prettier) along with [`prettier/stylelint-config-prettier`](https://github.com/prettier/stylelint-config-prettier) to match our prettier config.

Issue #577, opened 2019-08-13
GraphQL example is missing labels


wtrocki:
When creating GraphQL example labels will not be provided. 
After checking source code I found that extra object needs to be passed.
For example:

```
    city:   {label: 'City'},
    state:  {label: 'State'},
    street: {label: 'Street'},
    zip:    {label: 'Zip'}
```

Can labels be derived from ast values?
Example: https://astexplorer.net/#/gist/aaf298cda868080a86c1e1838c9fb382/a98e0667d362bdba99fafeee99818b8c411c05be

wtrocki:
Workaround for the moment is to use something like this:
```
visit(parse(schema), {
    FieldDefinition(node) {
        schemaData[node.name.value] = {
            label: node.name.value
        };
    }
})
```

radekmie:
Hi @wtrocki. It's something I thought about it for a long time. It's not the best solution, as the schema names often are not very user-friendly, but they can definitely be a sensible default. It'll also unify DX between bridges.

wtrocki:
Hmm.. I do not need to be in the core, but we can provide helper to wrap functions as above. Community contribution?

radekmie:
Well, yeah, but it really makes sense to put it in the core. `GraphQLBridge#getProps` to be exact.

wtrocki:
I have put some thought into entire graphql implementation and I think it needs some work to make it useful for production cases.

Main pain points I see now is:

### Performance
Generally parsing GraphQL schema (as oposed to documents) is quite bad idea. 
Getting thru complex AST was giving me extra 62 ms webpage rendering time on browser and 151ms on random phone. Using json schema do not add such overhead

## GraphQL specific issues 

Wrapping schema into form seems easy in hello world scenario but for normal use cases schema will diverge from form. Example? ID field in the schema is very rarely used as ID is assigned by server?

### How to address this issues

Generally GraphQL integration seems like just starting. I think the best GraphQL intergartion is no integration at all.

Imagine that I will put very complex schema into some cmd on build time and generate json schema. Then I can extend/modify json schema, push it to repo etc. without performance or usability penality. We can add extra directives to generator engine that can cover validation use cases etc. 

What this means is that uniforms will still be supporting graphql but in more usable way.

I'm working now on integrating uniforms with http://graphback.dev and I decided to not use GraphQL bridge

wtrocki:
[...]

PR #580, opened 2019-08-16
Added GraphQL default labels derived from schema.

kestarumper:
Targeting [issue 577](https://github.com/vazco/uniforms/issues/577)

Given schema:
```graphql
scalar Scalar

enum AccessLevel {
    Admin
}

input Author {
    id:        ID!
    confirmed: Boolean
    decimal1:  Float
    decimal2:  Float!
    firstName: String = "John"
    lastName:  String = "Doe"
    level:     AccessLevel
    tags:      [String]!
}

input Category {
    owners: [Author!]
}

input Post {
    id:       Int!
    author:   Author!
    custom:   Scalar
    title:    String
    votes:    Int
    example:  ID
    category: [Category!]!
}

# This is required by buildASTSchema
type Query { anything: ID }
```
## Feature
GraphQL bridge will now derive default labels from schema AST.
Works as follows: ~~splits AST field name string by capital letters `/(?=[A-Z])/` and joins them with spaces. Excessive  spaces are removed~~. Uses lodash's `_.lowerCase` to format string into lowercase word pieces. Finally ~~capitalizes whole string~~ uppercases first letter.
## Example
Schema path `author.firstName` will generate label: `First name`.


packages/uniforms-bridge-graphql/src/GraphQLBridge.js
```diff
@@ -126,14 +127,15 @@ export default class GraphQLBridge extends Bridge {
 
     const field = this.getField(nameGeneric, false);
     const fieldType = extractFromNonNull(field).type;
+    const defaultLabel = toHumanLabel(field.name) || '';
```
radekmie: I'm strongly against adding such functionality to the core package. How about using [`_.lowerCase`](https://lodash.com/docs/4.17.15#lowerCase) instead? It's almost it:

```js
function toHumanLabel(label) {
  // label === 'x.y.z.firstNameField';
  label = joinName(null, label);
  // label === ['x', 'y', 'z', 'firstNameField']
  label = label[label.length - 1];
  // label === 'firstNameField'
  label = _.lowerCase(label);
  // 'first name field'
  label = _.upperFirst(label);                     // Option A.
  label = label[0].toUpperCase() + label.slice(1); // Option B.
  // 'First name field'
  return label;
}
```

There's also [`_.startCase`](https://lodash.com/docs/4.17.15#startCase), but I like it a bit less. And let's keep this function local to this file, just like `extractFromNull`. There's no need for explicit tests of it - simply covering it in the other tests (as it is now) is enough.

radekmie:
A short follow up on the internal discussion: we'd have to remove merging `props` from `GraphQLBridge.getProps` as it is in the other bridges. It's not needed there, as `BaseField.getFieldProps` is resolving it anyway. We've decided to merge this logic as it is and get back to that later.

PR #591, opened 2019-09-05
Landing fixes

kestarumper:
image


website/components/LandingPage/Header/index.js
```diff
@@ -67,10 +68,6 @@ function Showcase() {
     addressLine1: {
       type: String
     },
-    addressLine2: {
-      type: String,
-      optional: true
-    },
```
radekmie: With this removed, I think we should rename the `addressLine1`.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -55,62 +56,74 @@ function cacheSet(key, data) {
...
+
+    const length =
+      (to.getFullYear() - from.getFullYear()) * 12 +
+      (to.getMonth() - from.getMonth());
+
     const dates = Array.from({ length })
-      .map((_, i) => from.getFullYear() + i + '-01-01')
+      .map(() =>
+        new Date(from.setMonth(from.getMonth() + 1)).toISOString().slice(0, 10)
+      )
```
radekmie: `Array.from` has a second argument. This `map` won't be needed with it.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -55,62 +56,74 @@ function cacheSet(key, data) {
...
 
+  const dates = dateRanges(from, to);
+
+  return Promise.all(
+    dates.map(range =>
+      fetch(`https://api.npmjs.org/downloads/point/${range}/uniforms`)
+        .then(response => response.json())
+        .then(({ downloads }) => downloads || 0)
+    )
+  ).then(sums => sums.reduce((acc, curr) => acc + curr));
```
radekmie: This fires quite a lot of requests at once. We can keep it as it is now, but let's just keep it in mind.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -55,62 +56,74 @@ function cacheSet(key, data) {
...
-          ['downloads']
-        ).then(({ downloads }) => downloads || 0)
-      )
-    )
-      .then(sums => sums.reduce((acc, curr) => acc + curr))
-      .then(downloads => setDownloads(downloads.toLocaleString('en-US')));
+
+    cached(`npm-${from}--${today}`, () => getNPMDownloads(from, today)).then(
+      downloads => setDownloads(downloads.toLocaleString('en-US'))
+    );
```
radekmie: Maybe it's too much to do it now, but we could cache them per-month for longer and then just update the latest data.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -43,73 +43,96 @@ function cacheGet(key) {
   }
 }
 
-function cacheSet(key, data) {
-  const twoMinutes = Date.now() + 2 * 60 * 1000;
+function cacheSet(key, data, length = 120000) {
+  const expires = Date.now() + length;
```
radekmie: I'd rather pass `expires` as an argument.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -43,73 +43,96 @@ function cacheGet(key) {
...
   if (data) return Promise.resolve(data);
 
-  return fetch(url)
+  return fallback().then(data => {
+    cacheSet(key, data, length);
+    return data;
+  }, console.error);
+}
+
+function yyyymmdd(date) {
```
radekmie: How about renaming it to `formatDate`? We can add `formatNumber` too (`.toLocaleString('en-US')`).

website/components/LandingPage/OpenSource/index.js
```diff
@@ -43,73 +43,96 @@ function cacheGet(key) {
...
+
+function dateRanges(from, to) {
+  from = new Date(from);
+  to = new Date(to);
+
+  const dates = [yyyymmdd(from)];
+  while (from < to) {
+    dates.push(yyyymmdd(new Date(from.setMonth(from.getMonth() + 1))));
+  }
+  dates.concat([yyyymmdd(to)]);
```
radekmie: `.push`, not `.concat`.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -43,73 +43,96 @@ function cacheGet(key) {
...
+}
+
+function dateRanges(from, to) {
+  from = new Date(from);
+  to = new Date(to);
+
+  const dates = [yyyymmdd(from)];
+  while (from < to) {
+    dates.push(yyyymmdd(new Date(from.setMonth(from.getMonth() + 1))));
+  }
```
radekmie: Using `for` loop here would be simpler (and would get rid of these two reassignments above). Actually using `for` here could also get rid of the `reduce` below.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -43,73 +43,96 @@ function cacheGet(key) {
...
+      new Date(todayDate.setMonth(todayDate.getMonth() - memoizedMonths))
+    );
+
+    Promise.all([
+      cached(
+        `npm-${start}--${mid}`,
+        () => getNPMDownloads(start, mid),
+        memoizedMonths * 31 * 24 * 60 * 60 * 1000
+      ),
+      cached(`npm-${mid}--${today}`, () => getNPMDownloads(mid, today))
```
radekmie: This code is firing even ore parallel requests. Let's discuss it later.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -37,43 +37,45 @@ function cacheKey(key) {
 function cacheGet(key) {
   try {
     const { expires, data } = JSON.parse(localStorage.getItem(cacheKey(key)));
-    if (expires > Date.now()) return data;
+    return { expired: expires < Date.now(), data };
   } catch (error) {
     // Nothing.
   }
+  return {};
```
radekmie: It'd be better to return `{ expired: true, data: null }` directly in the `catch` block.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -37,43 +37,45 @@ function cacheKey(key) {
...
-  const data = cacheGet(key);
-  if (data) return Promise.resolve(data);
+function cached(key, fallback, expires) {
+  const { expired, data } = cacheGet(key);
 
-  return fallback().then(data => {
-    cacheSet(key, data, length);
-    return data;
-  }, console.error);
+  if (data === undefined || expired === true) {
```
radekmie: With the above comment `if (expired) {` would be enough.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -37,43 +37,45 @@ function cacheKey(key) {
...
+      })
+      .catch(() => data || 0);
+  }
+
+  return Promise.resolve(data);
 }
 
 function formatDate(date) {
-  return date.toISOString().slice(0, 10);
+  return date && date.toISOString().slice(0, 10);
```
radekmie: `''`, 'undefined`, `false` or `null` is a very bad result in here. Why is this check actually needed?

website/components/LandingPage/OpenSource/index.js
```diff
@@ -37,43 +37,45 @@ function cacheKey(key) {
...
 }
 
 function formatNumber(number) {
-  return number.toLocaleString('en-US');
+  return number && number.toLocaleString('en-US');
```
radekmie: Same as above. I'd rather try to find and fix the origin of this problem.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -88,14 +90,27 @@ function dateRanges(from, to) {
 }
 
 function getNPMDownloads(from, to) {
+  function getRange(range, expires) {
```
radekmie: It can be hoisted. My proposed name: `getNPMDownloadsInRange`.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -88,14 +90,27 @@ function dateRanges(from, to) {
...
 
   return Promise.all(
-    dates.map(range =>
-      fetch(`https://api.npmjs.org/downloads/point/${range}/uniforms`)
-        .then(response => response.json())
-        .then(({ downloads }) => downloads || 0)
-    )
+    dates
+      .map(range => getRange(range, Date.now() + oneYear))
+      .concat(getRange(lastRange, Date.now() + oneDay))
```
radekmie: I'd rather use the date object methods instead of operating on milliseconds directly.

website/components/LandingPage/OpenSource/index.js
```diff
@@ -117,29 +132,16 @@ function useStats() {
     cached('github', getGitHubStats).then(({ forks, stars }) => {
       setForks(formatNumber(forks));
       setStars(formatNumber(stars));
-    });
+    }, Date.now() + 2 * 60 * 1000);
```
radekmie: Only two minutes?

website/docusaurus.config.js
```diff
@@ -1,8 +1,12 @@
 const path = require('path');
 
 module.exports = {
+  customFields: {
+    email: 'hello@uniforms.tools',
+    keywords: ['forms', 'react', 'schema']
+  },
```
radekmie: Very minor thing: can we move it _a bit _ down? Right after the `url`.

website/components/LandingPage/OpenSource/index.js
kestarumper: Github api after rate limit has been surpassed returns empty object `{}` as a response, resulting in `stars` and `forks` being `undefined`. I will move this check somewhere else.

website/components/LandingPage/OpenSource/index.js
kestarumper: In the end I still have to convert them to timestamp. This saves a lot of date object instances and conversions.

website/components/LandingPage/OpenSource/index.js
kestarumper: This is the minimum which won't cause rate limiting set by Github. It depends on what we would like to accomplish. I choose to go with the most up to date statistics. What time interval would you suggest?

radekmie: OK, right, we've already discussed it.

website/components/LandingPage/OpenSource/index.js
radekmie: OK, good.

PR #592, opened 2019-09-09
WIP Accelerate reproductions

### LINKED ISSUE #587: Accelerate reproductions
It would be nice if you could do something like: https://codesandbox.io that includes a dummy setup of Uniforms running.

So if someone wants to show a bug or something automatically forks your sandbox and leaves the link inside GitHub.

kestarumper:
closes #587

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
-<!-- 
+<!--
   Before you open an issue, please check if a similar one already exists or has been closed before.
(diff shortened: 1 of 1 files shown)


radekmie:
How about using [CodeSandbox import](https://codesandbox.io/docs/importing#import-from-github) feature? If we'd provide a complete repo in a sub-directory, CodeSandbox can import it on demand. The current solution is good, but it'd require us to maintain the code in two places, possibly with versioning problems.

kestarumper:
We're going to provide a demo directory compatible with CodeSandbox.

PR #597, opened 2019-09-17
Fixed top padding when entering docs.

kestarumper:
Before:
image

After fix:
image

### DIFF
--- a/website/components/LandingPage/index.module.css
+++ b/website/components/LandingPage/index.module.css
@@ -7,10 +7,6 @@
   }
 }
 
-:global #__docusaurus {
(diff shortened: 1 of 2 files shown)


website/components/LandingPage/index.module.css
```diff
@@ -86,7 +82,8 @@
   --ifm-hero-background-color: var(--dark-blue-grey);
 }
 
-.landing .overflow-hidden {
+.landing .overflow-hidden,
+.overflow-hidden {
```
radekmie: Is this `.landing` prefixed rule still needed?

PR #600, opened 2019-09-18
Added eslint-plugin-import

radekmie:
Finally, imports will be at least partially organized. Sorting is not yet possible: benmosher/eslint-plugin-import#389.

**EDIT:** I've sorted them semi-automatically using [renke/import-sort](https://github.com/renke/import-sort).


radekmie:
Reference order should be like this:

```js
// No bindings.
import 'path';
// Namespaces.
import * as X 'path';
import * as Y 'path';
import * as z 'path';
// Default imports (and optional named).
import A from 'path';
import B, { Any1, any1, b1 } from 'path';
import C from 'path';
import a from 'path';
import b, { Any2, any2, b2 } from 'path';
// Only named.
import { C, D } from 'path';
import { E } from 'path';
import { e, f } from 'path';
import { g } from 'path';
```

Issue #609, opened 2019-10-18
Separate, de-coupled Themes, or Theme contribution guide


refactorized:
**Background**

I am trying to fork the uniforms-material package and running into a lot of trouble.

1) I can't easily fork just the theme as it lives in a monorepo.  Forking the entire repo makes for a difficult time referencing the specific package locally.  I am surely missing some things, but I don't know what. 

2) If I just copy the typescript files defining the theme I have to then somehow get type script working in the smaller project, or in my larger project in which the components just live under the main src directory.  At best this means divining the proper configuration from the monorepo but I have not got TS working yet. 

3) If I skip TS altogether and pull the 4.2.1 versions of the components I get problems because the components _extend_ objects found in the base uniforms package.  This means I am locked into that specific version of uniforms, or have to extract the extended code into the theme because I can't extend a compiled es5 object in es6 code with babel.

**Requests / Suggestions**
1) Write themes that only depend on the proper API, not some upstream implementation detail. If this means duplicating too much functionality, then expose that functionality as helper methods that can be invoked (by anything) rather than classes that require source comprehension to extend.

and / or

2) Documentation on how to fork and improve the themes in the monorepo.

radekmie:
Hi @refactorized. It's a tough one. For all this time I've tried to encourage custom themes and make it as easy as possible. Let me address your points one by one and then try to give a broader answer.

1. This one boils down to your idea of _forking_. If you mean forking the repo, then it's fairly easy - just do it. Want to add a new theme? Simply copy or edit one of the existing ones. If you mean something different, I'll _probably_ get to that later.
2. We've transitioned to TypeScript just recently. There are very few types at the moment (but we are working on it) and there's no special configuration (there is, but almost every line of it is used to make it work as desired in the monorepo).
3. Version 4.2.1 of Material UI I assume?
    > [...] of the components I get problems because the components _extend_ objects found in the base uniforms package. This means I am locked into that specific version of uniforms, or have to extract the extended code into the theme because I can't extend a compiled es5 object in es6 code with babel.

    What objects? Why are you locked? Is there some runtime problem? Did you mean tests? What kind of _extend_ operation? I'm completely lost here, sorry.

1. There is *no dependency* on any implementation detail! If there is any (I may overlook something), then we definitely have to fix it.
2. I'll get to it below.

[...]

PR #612, opened 2019-10-23
Added ES6 targets (closes #433).

### LINKED ISSUE #433: meteor 1.7.0.1 - TypeError: Class constructor _class cannot be invoked without 'new'

kestarumper:
We've added separate compilation target for ES6 standard which enables classes.

We would like to ask @macrozone @refactorized to check if everything works fine.
1. Clone this branch.
2. Run `npm ci`.
3. Copy built files directly into `node_modules` to check if it works.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,6 +6,7 @@
 /npm-debug.log*
 /packages/*/*.d.ts
 /packages/*/*.js
+/packages/*/es6
 /packages/*/node_modules
 /website/.cache-loader
 /website/.docusaurus
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /npm-debug.log*
 /packages/*/*.d.ts
 /packages/*/*.js
+/packages/*/es6
 /packages/*/node_modules
 /website/.cache-loader
 /website/.docusaurus
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,6 +6,7 @@
 /npm-debug.log*
 /packages/*/*.d.ts
 /packages/*/*.js
+/packages/*/es6
 /packages/*/node_modules
 /website/.cache-loader
 /website/.docusaurus
--- a/packages/uniforms-antd/package.json
+++ b/packages/uniforms-antd/package.json
@@ -3,6 +3,7 @@
   "version": "2.5.0",
   "license": "MIT",
   "main": "index.js",
+  "module": "es6/index.js",
   "description": "Ant Design UI components for uniforms.",
   "repository": "https://github.com/vazco/uniforms/tree/master/packages/uniforms-antd",
   "bugs": "https://github.com/vazco/uniforms/issues",
--- a/packages/uniforms-antd/package.json
+++ b/packages/uniforms-antd/package.json
@@ -19,6 +19,8 @@
   "files": [
     "*.d.ts",
     "*.js",
+    "es6/*.d.ts",
+    "es6/*.js",
     "src/*.ts",
     "src/*.tsx"
   ],
--- a/packages/uniforms-antd/tsconfig.es6.json
+++ b/packages/uniforms-antd/tsconfig.es6.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "es6",
+    "target": "es6",
+    "tsBuildInfoFile": "../../node_modules/.cache/uniforms-antd.es6.tsbuildinfo"
+  }
+}
--- a/packages/uniforms-bootstrap3/package.json
+++ b/packages/uniforms-bootstrap3/package.json
@@ -3,6 +3,7 @@
   "version": "2.5.0",
   "license": "MIT",
   "main": "index.js",
+  "module": "es6/index.js",
   "description": "Bootstrap3 UI components for uniforms.",
   "repository": "https://github.com/vazco/uniforms/tree/master/packages/uniforms-bootstrap3",
   "bugs": "https://github.com/vazco/uniforms/issues",
--- a/packages/uniforms-bootstrap3/package.json
+++ b/packages/uniforms-bootstrap3/package.json
@@ -19,6 +19,8 @@
   "files": [
     "*.d.ts",
     "*.js",
+    "es6/*.d.ts",
+    "es6/*.js",
     "src/*.ts",
     "src/*.tsx"
   ],
--- a/packages/uniforms-bootstrap3/tsconfig.es6.json
+++ b/packages/uniforms-bootstrap3/tsconfig.es6.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "es6",
+    "target": "es6",
+    "tsBuildInfoFile": "../../node_modules/.cache/uniforms-bootstrap3.es6.tsbuildinfo"
+  }
+}
--- a/packages/uniforms-bootstrap4/package.json
+++ b/packages/uniforms-bootstrap4/package.json
@@ -3,6 +3,7 @@
   "version": "2.5.0",
   "license": "MIT",
   "main": "index.js",
+  "module": "es6/index.js",
   "description": "Bootstrap4 UI components for uniforms.",
   "repository": "https://github.com/vazco/uniforms/tree/master/packages/uniforms-bootstrap4",
   "bugs": "https://github.com/vazco/uniforms/issues",
--- a/packages/uniforms-bootstrap4/package.json
+++ b/packages/uniforms-bootstrap4/package.json
@@ -19,6 +19,8 @@
   "files": [
     "*.d.ts",
     "*.js",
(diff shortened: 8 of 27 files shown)


macrozone:
@kestarumper thank you so much, will check if that solves my issues!  Will give you an update

radekmie:
I've checked it with the repro from @refactorized (see https://github.com/vazco/uniforms/issues/609#issuecomment-545079686) and it worked with indirect imports (`import { X } from 'uniforms'` instead of `import X from 'uniforms/X'`). If this will work also for @macrozone, I think we'll have to deprecate it and suggest named imports. As it would be a breaking change, we'll change it as soon as in version 3.

refactorized:
@radekmie That is fantastic news - how do I test this out?  Clone the repo, run CI and try a filesystem import?

refactorized:
And/Or can you push the changes to the skel repo, if it's not too much trouble.  You don't even need to PR it, I added you as a contributor.

macrozone:
> I've checked it with the repro from @refactorized (see [#609 (comment)](https://github.com/vazco/uniforms/issues/609#issuecomment-545079686)) and it worked with indirect imports (`import { X } from 'uniforms'` instead of `import X from 'uniforms/X'`). If this will work also for @macrozone, I think we'll have to deprecate it and suggest named imports. As it would be a breaking change, we'll change it as soon as in version 3.

I think we can introduce it asap, it schould be non-breaking. Also you can still import from `import X from 'uniforms/X', it shoud pull in the es6 code, when the bundler supports the `module` entrypoint.

macrozone:
hmm i could not manage to make it work:

- i cloned the branch, 
- npm install
- copy packages/uniforms to my project node_modules, removing the uniforms there first
- i verified the existence of the build in packages/uniforms/es6 and the package.json with the "module": "es6/index.js" entry point
- removed the transform-classes again
- replaced ever "import x from 'uniforms/x'" with "import {x} from 'uniforms'"
- started the project: "Class constructor _class cannot be invoked without 'new'"

so i verified that indeed still the old es5 code is loaded, by deleting that file (outside of es6 folder) and it would complain that the file was not found. So it stills loads the es5 code. :-/ any idea what i am doing wrong?

radekmie:
@refactorized: I can't do it, as I wouldn't commit `node_modules`. I've simply copied the repo, run `npm run build` and then copied the package into the `node_modules` by hand. @macrozone instructions are more detailed.

@macrozone: Try to remove `node_modules/.cache` - it did the work on my side. Also, importing form `'uniforms'` is hitting the `main` or `module` file (depends on the bundler), but `'uniforms/X'` will always load a `node_modules/uniforms/X.js` file. In general, it's a good addition, but we should definitely mention it in the docs.

macrozone:
> @refactorized: I can't do it, as I wouldn't commit `node_modules`. I've simply copied the repo, run `npm run build` and then copied the package into the `node_modules` by hand. @macrozone instructions are more detailed.
> 
> @macrozone: Try to remove `node_modules/.cache` - it did the work on my side. Also, importing form `'uniforms'` is hitting the `main` or `module` file (depends on the bundler), but `'uniforms/X'` will always load a `node_modules/uniforms/X.js` file. In general, it's a good addition, but we should definitely mention it in the docs.

i tried that, but .cache did only contain code from storybook (of course i tried to delete it anyway). Maybe next.js uses another cache. ... anyway i will try again. Your changes look correct to me and as mentioned, i  did the same in another library project.... so it should work, i will try again soon.

refactorized:
Example of working project:
https://github.com/refactorized/uniforms-skel/tree/working-with-special-uniforms-branch

You should be able to clone this, follow the directions (changing only the local file location in your package.json), and get a working form.

macrozone:
@radekmie is it possible to get  pre-release on npm? the class-problem gives us still a lot of headaches and installing it without npm seems also to be a bit hard

radekmie:
@macrozone I'd like to do a normal relase today anyway. Sorry for that, I also don't like to delay it so much, but we've struggled with some consequences of that change.

macrozone:
@radekmie thank you, i tried out 2.6.0 

unfortunatly, it does not solve my issue. i still get `Uncaught TypeError: Class constructor _a cannot be invoked without 'new'`

It's a nextjs app. 

The problematic code is this:


```

import { AutoForm as UniformsAutoForm } from "uniforms";


export const Auto = (parent): any =>
  class extends UniformsAutoForm.Auto(parent) {
    public static Auto = Auto;
  };

// ....
````

**Comment by radekmie** [comment]
@macrozone I've just tested it in our `website` sub-project. It's a Docusaurus project, which involves Webpack 4, and it just works. Could you provide a reproduction and file it in a new issue?

**Comment by macrozone** [comment]
@radekmie yes, i should make a reproduction. But are you 100% sure that it uses the non-transpiled code? E.g can you try the problematic class pattern from above?

My next attempt will be to remove everything but the es6 folder. My guess is that webpack priorizes index.js in the root directory over main or module. It would be better anyway if the source would be in a src directory and the builds in a directory each

Sorry, on my phone, i will continue tomorrow

**Comment by radekmie** [comment]
@macrozone yes, I've checked it twice, including the sources loaded in the browser (search for `uniforms/es6`). I've also made an [example CodeSandbox project](https://codesandbox.io/s/purple-bush-9yoo5), which includes your code.

**Comment by macrozone** [comment]
@radekmie super weird. I just don't know why it does not work in my case. 

now, i tried to delete the es5-files in the root directory of node_modules/uniforms, but let the es6 in place.

It now throughs an interesting error:

```
./node_modules/uniforms-bridge-json-schema/es6/register.js
Module not found: Can't resolve 'uniforms' in '/path/to/project/node_modules/uniforms-bridge-json-schema/es6'
```

which is interesting, because it throws inside of es6-folder in uniforms-bridge-json-schema and fails to resolve `import { Bridge, joinName } from 'uniforms';`

so it looks like that at least uniforms-bridge-json-schema gets resolved correctly to its module entry point. :-o

i'll keep you updated, will investigate

macrozone:
wow, i think its this: https://github.com/zeit/next.js/issues/9323

-.-

macrozone:
I give up. @radekmie is there a way, to create a custom styled autoform and fields without this `extends class` nonsense? :-((

radekmie:
@macrozone That's a really weird bug (?) in Next.js. If you need a custom styling, create all of the fields (so your `NestField` will use your `AutoField`) and override the `render` of given form, e.g. `class X extends AutoForm { render() { /*...*/ } }` or even `AutoForm.prototype.render = /* ... */`.

macrozone:
@radekmie ok, es target seems to work.

The problem is now the other way around:

I can't extend AutoForm or BaseField, because my code is already transpiled, by it pulls in it the untranspiled code with native classes and you can't extend a native class with a non-native-class.

I thought that webpack and babel would first pull in the modules and would transpile the result based on the env, but that seems not to be the case...

radekmie:
It looks like @JanMP has the same problem. Maybe try to import `uniforms/index` instead of `uniforms`? It'll import the ES5 build. Otherwise, I'm really confused, what's actually happening.

macrozone:
@radekmie i created a reproduction: https://github.com/macrozone/uniforms-next-babel-reproduction

it throws the dreaded error: "TypeError: Class constructor _a cannot be invoked without 'new'"

solution is to use https://github.com/martpie/next-transpile-modules and transpile uniforms

PR #622, opened 2019-10-30
Fixed preservation of mixin types (fixes #617).

### LINKED ISSUE #617: typescript error in uniforms-materialui
given 

```
import { AutoForm, AutoFields } from 'uniforms-material';

// later

   <AutoForm
         ...
        >
          <AutoFields />
        </AutoForm>

```

yields:

`JSX element type '{ [x: string]: any; }' is not a constructor function for JSX elements.`
`Type '{ [x: string]: any; }' is missing the following properties from type 'ElementClass': render, context, setState, forceUpdate, and 3 more.`

the generated type of `AutoForm` is:

```
declare const _default: {
    new (): {
        [x: string]: any;
    };
    [x: string]: any;
    Auto: (parent: any) => any;
};
export default _default;
```

it seems to complain about the type of `new()` which should be a react component

radekmie:
As reported in #617, `AutoForm` (other form components too) have non-JSX compatible types. With this change base class (`Component`) and mixin helpers are preserved. It's not an ideal solution but it'll do for now. In the future, we'd have to define `BaseForm` generic type and other forms should simply derive from it, reusing generics.

### DIFF
--- a/packages/uniforms/src/AutoForm.tsx
+++ b/packages/uniforms/src/AutoForm.tsx
@@ -4,10 +4,21 @@ import isEqual from 'lodash/isEqual';
 import omit from 'lodash/omit';
 import set from 'lodash/set';
 
+// FIXME: Fix configuration of ESLint in TypeScript files.
+// eslint-disable-next-line no-unused-vars
+import BaseForm from './BaseForm';
 import ValidatedQuickForm from './ValidatedQuickForm';
 
-const Auto = (parent: any) =>
-  class extends parent {
+function Auto<
+  T extends {
+    new (...args: any[]): BaseForm;
+    childContextTypes?: {};
+    defaultProps?: {};
+    displayName?: string;
+    propTypes?: {};
+  }
+>(parent: T): T & { Auto: typeof Auto } {
+  return class extends parent {
     static Auto: any = Auto;
 
     static displayName: string = `Auto${parent.displayName}`;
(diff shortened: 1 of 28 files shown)


packages/uniforms/src/AutoForm.tsx
```diff
@@ -4,10 +4,21 @@ import isEqual from 'lodash/isEqual';
...
+  T extends {
+    new (...args: any[]): BaseForm;
+    childContextTypes?: {};
+    defaultProps?: {};
+    displayName?: string;
+    propTypes?: {};
+  }
+>(parent: T): T & { Auto: typeof Auto } {
+  return class extends parent {
     static Auto: any = Auto;
```
kestarumper: Probably this `any` could be deleted.

packages/uniforms/src/BaseForm.tsx
```diff
@@ -201,7 +201,7 @@ export default class BaseForm extends Component<any, any> {
     return changedKeys(root, valueA, valueB);
   }
 
-  getNativeFormProps() {
+  getNativeFormProps(): { [prop: string]: unknown } {
```
kestarumper: There is a nice utility type provided by typescript: [`Record<K,T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt). So you could use `Record<string, unknown>`.

PR #643, opened 2019-11-12
Fixed RatingField example in docs

yurtsiv:
Fixes a problem with `RatingField` example in the documentation (the field isn't rendered).

image


docs/examples/CustomFields/RatingField.js
```diff
@@ -14,11 +14,10 @@ const Rating = ({
   onChange
 }) => (
   <section className={classnames('ui', { disabled, required }, className)}>
-    {[...Array(max)]
-      .map((_, index) => index + 1)
```
yurtsiv: Babel has different behavior than browser's one an this `map` has no effect

Issue #661, opened 2019-12-06
SubmitField and ErrorsField throw in TypeScript


chmac:
```
Foo.tsx
TypeScript error in Foo.tsx(13,10):
Property 'children' is missing in type '{}' but required in type '{ [x: string]: any; children: any; disabled?: any; inputRef?: any; label?: any; value?: any; }'.  TS2741
```

The typing that I see for the `SubmitField` from material is like so:

```typescript
/// <reference types="react" />
declare const SubmitField: {
    ({ children, disabled, inputRef, label, value, ...props }: {
        [x: string]: any;
        children: any;
        disabled: any;
        inputRef: any;
        label: any;
        value: any;
    }, { uniforms: { error, state } }: {
        uniforms: {
            error: any;
            state: any;
        };
    }): JSX.Element;
    contextTypes: {
        uniforms: any;
    };
    defaultProps: {
        label: string;
        variant: string;
    };
};
export default SubmitField;
```

If I make all the parameters optional on the first example (by editing the definition file) then my error goes away. I'm not sure how these definition files are built, so I'm not sure where these changes could be made. I guess there's typing missing in the material `SubmitField` component...

radekmie:
Hi @chmac. Indeed, currently inferred types are... Not suitable to work with. We'll try to do as much as possible now, but more general support will come with v3.

chmac:
OK. Will wait to see how it works in v3. Right now I worked around the issue by just replacing the `<SubmitField` with a simple `<Button`. I don't have the whole lerna setup running for uniforms so won't be able to do much in terms of useful PRs.

PR #662, opened 2019-12-06
Add ugly, any typing to `SubmitField`.

chmac:
This might address the issue I reported in #661.

I haven't tested this. I copied the code locally into my own app and it seems to work, but that's as good as I can offer at the moment I'm afraid...


radekmie:
Hi @chmac. Besides the linting issue, it's OK. But as the types are incomplete anyway, I'd simply go with `any` for now and tidy it up in v3. I'll close this one as soon as a superseded PR will arrive.

Issue #681, opened 2020-01-30
SemanticUI BoolField should be "fitted" if it has no label


emlai:
For consistency with [semantic-ui-react](https://react.semantic-ui.com/modules/checkbox/#variations-fitted):

> The fitted class is automatically applied if there is no label prop.

PR #687, opened 2020-02-07
Update semantic BoolField (fixes #680, #681).

### LINKED ISSUE #680: SemanticUI BoolField is missing wrapClassName

kestarumper:
This pull request fixes #680 and #681 by introducing `wrapClassName` prop into `BoolField` and logic for adding _fitted_ class name  if `label` prop is falsy (instead of only if `label` is `undefined`, not passed). This last functionality differs from the one specified [here](https://react.semantic-ui.com/modules/checkbox/#variations-fitted).

### DIFF
--- a/packages/uniforms-semantic/src/BoolField.tsx
+++ b/packages/uniforms-semantic/src/BoolField.tsx
@@ -22,7 +22,14 @@ const Bool = ({
     className={classnames(className, { disabled, error, required }, 'field')}
     {...filterDOMProps(props)}
   >
-    <div className={classnames('ui', wrapClassName, 'checkbox')}>
+    <div
+      className={classnames(
+        'ui',
+        wrapClassName,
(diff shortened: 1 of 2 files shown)


packages/uniforms-semantic/src/BoolField.tsx
```diff
@@ -15,13 +15,21 @@ const Bool = ({
...
   <div
     className={classnames(className, { disabled, error, required }, 'field')}
     {...filterDOMProps(props)}
   >
-    <div className="ui checkbox">
+    <div
+      className={classnames(
+        'ui',
+        wrapClassName,
+        !label && label !== true && 'fitted',
```
radekmie: Is there a way to put `true` in here? I'm pretty sure in this case a default value (e.g. from schema) is taken instead.

PR #702, opened 2020-04-01
Reworked website examples

radekmie:
Until now, each interactive example in the docs required us to copy its source into the `.mdx` files. This change adds a new component called `CodeSection`, which should replace all of the duplicated code with using `raw-loader` (importing source instead of the parsed file).

Additionally, I've added _section_ concept, which allows cutting only a section of the file. In the code noted as `<sectionName>`/`</sectionName>` tags.

- [x] Rewrite all examples.
- [x] Remove old `Code` component and `code` helper (`website/components/Code`).

---

```ts
type CodeSectionProps = {
  // Source language (for highlighting).
  language: string;
  // Map of patterns to replace.
  replace?: Record<string, string>;
  // Section.
  section?: string;
  // Source. Object variant is useful to work with CommonJS imports.
  source: string | { default: string };
};
```

Ping @martachrzanowska, @Monteth.


docs/examples/Tutorial/GuestFormWithErrorFields.js
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
-
-import GuestSchema from './GuestSchema';
 import {
   AutoForm,
```
martachrzanowska, line 3: What about the order of imports here?

radekmie: After rendering in `ExampleCustomizer` all `universal` imports are renamed to the correct theme (`uniforms-semantic`, etc.). I wanted to keep the correct order on the rendered docs more than in the code.

docs/examples/Tutorial/GuestFormWithFieldsInSchema.js
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
-
-import GuestSchema3 from './GuestSchema3';
 import {
   AutoForm,
```
martachrzanowska, line 3: Same as above.

docs/examples/Tutorial/GuestFormWithHeading.js
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
-
-import GuestSchema from './GuestSchema';
 import {
   AutoForm,
```
martachrzanowska, line 3: Same as above.

docs/examples/Tutorial/GuestFormWithErrorFields.js
martachrzanowska: Alright :)

Issue #51, opened 2016-06-28
Feature Request: onSubmit callback or onSubmitSuccess/onSubmitFailure events


zeroasterisk:
If I am submitting a form via a method call (for example) and it's `async` --- I would like the ability to pass down a callback and act on it's response.

The new API for `formRef.submit()` is great, but I can not see any way to act on results from that submit.

Here's my current use case:

1. Normal submit the form, saves a document and shows a success notification, doesn't navigate
2. Clicking on a `Save & Continue` button, should submit the form, and on success, navigate to the next page.

The problem is, I can trigger `formRef.submit()` but I have no way to wait on the response of it, from the React component level, without involving internal state.

Here is what I'm doing now to handle this:

```js
  constructor(props) {
    super(props);
    this.state = { afterSaveContinue: false };
    this.onSubmit = this.doSubmit.bind(this);
  }
  doSubmit(data) {
    const { afterSaveContinue } = this.state;
    const { myDocument, handleSubmit } = this.props;
    if (!afterSaveContinue) {
      return handleSubmit(data);
    }
    return handleSubmit(data, (err, res) => {
      if (err) return;
      browserHistory.push(`/myDocument/${myDocument._id}/nexpage`);
    });
  }
  setAsClicked() {
    this.setState({ afterSaveContinue: true });
  }
```

But, I could imagine enabling some alternate submit-like handler, when clicking on the "continue" button, which passed either extra arguments into our `onSubmit` function, or received standard events or callbacks from it.

radekmie:
Yep, that's a quite common use case. How about simply doing a `return` at the end of `onSubmit`? This could suffice your needs because your `doSubmit` could return a `Promise` or something similar.

---

I don't really get your idea with `onSubmitSuccess` / `onSubmitFailure` - could you describe it more precisely?

zeroasterisk:
_I typed in a great comment on this last week... but never submitted it and I don't feel like re-typing :/_

I think the idea of a `doSubmit` returning a `Promise` and resolving... but I honestly don't know how to do that...

Can you provide an example of: onSubmit, sets the button to disabled until submit handler resolves?

radekmie:
### Introduction

Let's assume, that `BaseForm#onSubmit` returns `props.onSubmit(...)`. Then consider the following example:

```js
class Example extends React.Component {
    constructor (props) {
        this.onSubmit     = this.onSubmit.bind(this);
        this.onSubmitNext = this.onSubmitNext.bind(this);
    }

[...]

Issue #17, opened 2016-06-04
[Feature Request] Expose validation context to onSubmit() for manually adding async errors


serkandurusoy:
Simple schema runs validations synchronously, but some validations require server roundtrips or other async calls.

For example, I may want to check a given input if it conforms to some security related constraints or uniqueness constraints etc. And the only way to do that may be to use a server side method. And if it does not, I would like to show an error message to the user.

For such cases, simple schema allows [manually adding validation errors](https://github.com/aldeed/meteor-simple-schema#manually-adding-a-validation-error)

Currently, uniforms only provides the doc to onSubmit callback so there is no way to access the current validation context of the simple schema.

So, it would be great if onSubmit callback had access to the simple schema validation context. In fact, there may be other valuable information that can be provided. So, it would even be better if we had a way to access the whole autoform context in there, not just the document.

radekmie:
For such a validation you can pass an `error` prop to form. Provided error should be compatible with your schema (`ValidationError` for `SimpleSchema`). Also, `onSubmit` is a final action.

serkandurusoy:
Hmm, do you mean something like:

```javascript

const PostSchema = new SimpleSchema({
  title: {type: String},
  body: {type: String}
})

class AddPost extends Component {
  constructor(props) {
    super(props)
    this.state={
      asyncErrors: null
    }
  }

  submitPost(doc) {
    PostSchema.clean(doc)

    Meteor.call('checkIfPostTitleIsUnique', doc, (err,res) => {
      if (err) {
        this.setState({
          asyncErrors: new ValidationError([{name: 'title', type: 'notUnique'}])
        })
      }
    })
  }

  render {
    return <AutoForm schema={PostSchema} 
                     error={this.state.asyncErrors}
                     onSubmit={doc => this.submitPost(doc)} />
  }

}
```

Does this override the form errors or does it get combined with existing errors?

And what do you mean by onSubmit being a final action? In that case, how do you propose we do asynchronous type checks and validatinons?

I can't give you an answer right now about why I might need the form context, but obviously, aldeed:autoform provides necessary mechanisms and also hooks that give you full access to the form before, during and after its submission. And there may be cases such flexibility is valuble.

serkandurusoy:
Edit:

I think a way to access the form's instance or at least the "simple schema validation context" it creates, is very important. That way, we can even define async validations within the schema itself!

radekmie:
[...]

PR #711, opened 2020-04-22
Asyncified validate + submit flow

### LINKED ISSUE #644: Validation Errors should not rely on throw/catch
Validators currently work by throwing an object that is at a minimum `{ details: arrayOfErrors}`

This is not best practice as validation errors are not actually runtime errors but the proper result of a working validator against a currently invalid model.

In future versions it would be better to handle this via returning errors directly (versus returning nothing to indicate success), using promises, error first callbacks, etc.

**Side note:** 

The current system also requires throwing a somewhat nonstandard Error object that cannot be created via the Error() constructor alone.  If I am using, for instance the JSONSchemaBridge and want to throw a standard Error object (as to appease a linter) I would need the following code:

```
throw Object.assign(new Error('Validation Error'), { details: validator.errors })
```

Of course one could argue throwing a standard Error object is perhaps worse because, once again, this is not an actual error condition.  In my case I settle on just disabling ESLint for the line in question.

radekmie:
Let's start with a short flashback from #644:
> [...] Well, the first bridge ever was the SimpleSchema (the Meteor one), and its validator throws an error instead of returning it. It was perfectly fine, up to the point when we wanted to add `onSubmitSuccess`/`onSubmitFailure` and async validation. Of course, we managed to do so, but the solution is not very elegant.

This _not very elegant solution_ got implemented in #51. Why it's not elegant? Because it's redundant. There's no need for such hooks if one can simply add it directly to the `onSubmit`. Another problem (non-elegancy) is the asynchronous validation - invented in #17 and implemented in 4a56596f8a2e296f891316430cdba104cd21f8e6. It's weird and is using callbacks.

This PR introduces a new, clean flow (including https://github.com/vazco/uniforms/pull/711#issuecomment-618319641):
* **Bridge validators can to be asynchronous.**
* **Bridge validators have to return errors instead of throwing them.** More background is in #644.
* **The `onValidate` is now `(model, error?) => error? | Promise<error?>`.** This is far better and obvious than the callback version before.
* **Removed `onSubmitSuccess` and `onSubmitFailure`.** I'm pretty sure there's no viable case for them at all. Migration is straightforward:
  ```diff
  -onSubmit={onSubmit}
  -onSubmitSuccess={onSubmitSuccess}
  -onSubmitFailure={onSubmitFailure}
  +onSubmit={model => {
  +  const result = onSubmit(model);
  +  result.then(onSubmitSuccess, onSubmitFailure);
  +  return result;
  +}}`
  ```
* **Separated `submitting` and `validating` states.** Now `validating` is `true` only while the bridge validator or `onValidate` are currently running. Same for `submitting` - it's try only while `onSubmit` is running.

**TODO:**
- [x] Perform some manual tests to check if there's no visible difference with this change. It might be the case as few things got _asyncified_.
- [x] Run benchmarks to ensure no performance degradation.
- [x] Add missing tests.
- [x] Update `CHANGELOG.md`.

---

Closes #644.

### DIFF
--- a/packages/uniforms/__tests__/ValidatedForm.tsx
+++ b/packages/uniforms/__tests__/ValidatedForm.tsx
@@ -343,29 +343,31 @@ describe('ValidatedForm', () => {
   });
 
   describe('validation flow', () => {
-    beforeAll(jest.useFakeTimers);
-    afterAll(jest.useRealTimers);
-
     const variantGroups = [
       {
         'fail-async': () => Promise.resolve(error),
         'fail-sync': () => error,
         'good-async': () => Promise.resolve(null),
+        'good-async-silent': () => Promise.resolve(),
         'good-sync': () => null,
+        'good-sync-silent': () => {},
       },
       {
         'fail-async': () => Promise.resolve(error),
         'fail-sync': () => error,
         'good-async': () => Promise.resolve(null),
+        'good-async-silent': () => Promise.resolve(),
         'good-sync': () => null,
+        'good-sync-silent': () => {},
         'pass-async': (_, error) => Promise.resolve(error),
         'pass-sync': (_, error) => error,
       },
       {
         'fail-async': () =>
           new Promise((_, reject) => setTimeout(() => reject(error))),
-        'good-async': () => new Promise(resolve => setTimeout(resolve)),
-        'good-sync': () => {},
+        'good-async': () =>
+          new Promise(resolve => setTimeout(() => resolve('ok'))),
+        'good-sync': () => 'ok',
       },
     ] as const;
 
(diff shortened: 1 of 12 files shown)


radekmie:
During benchmarks, it turned out that making everything async has one flaw: `setState` happens before `then`. It means that we have at least one additional render. Right now I'm working on a solution similar to [the one in `BaseForm`](https://github.com/vazco/uniforms/blob/04e1ff427998f1369f39945ad416f443c2c0ff7a/packages/uniforms/src/BaseForm.tsx#L281-L288) to handle both sync and async validators, `onValidate`, and `onSubmit` in any configuration.

Another problem araised while working with rejected promises. The thing is that it's not better than `throw` as we still cannot type it anyhow. We won't be able to type it correctly anytime soon (I don't have an idea on how to type errors as those are created in bridges, used in forms, and then analyzed by bridges), but having such a possibility would be nice. I'll experiment with actually returning errors instead, as discussed in #644.

packages/uniforms/__tests__/ValidatedForm.tsx
```diff
@@ -389,4 +341,122 @@ describe('ValidatedForm', () => {
...
+      return Object.keys(x) as (keyof X)[];
+    }
+
+    const cases = cartesian(
+      [true, false] as [true, false],
+      cartesian(
+        keys(variantGroups[0]),
+        cartesian(keys(variantGroups[1]), keys(variantGroups[2])),
+      ),
+    ).map(([a, [b, [c, d]]]) => [a, b, c, d] as const);
```
kestarumper, line 391: This is clever, but unreadable. I don't think anyone in the future will know what is going on here (except maybe the author). But let's leave it as is.

packages/uniforms/src/ValidatedForm.tsx
```diff
@@ -132,34 +121,24 @@ export function Validated<Base extends typeof BaseForm>(Base: Base) {
...
-            // If validation fails, or `super.onSubmit` doesn't touch `submitting`, we need to reset it.
-            this.setState(state =>
-              state.submitting ? { submitting: false } : null,
-            );
-          }
+        // Validation failed (i.e. returned an error), so no error is present
+        // both in the props nor the state.
+        return super.onSubmit().catch(error => {
+          this.setState({ error });
+          throw error;
```
kestarumper, line 135: Why is this `throw`ed here if we follow up with `.catch(noop)` later on (line 139)?

packages/uniforms/__tests__/ValidatedForm.tsx
radekmie: I'll extract it to a separate function with an appropriate name - it should help.

packages/uniforms/src/ValidatedForm.tsx
radekmie: Think of it as of a _re-rejection_ (as in _re-throwing_ an error). Not doing that would effectively swallow `onSubmit` errors.

The `result.catch(noop);` below is there just to suppress the `Unhandled rejection` errors. Every validation error leads to a rejected promise (that `Promise.reject` in line 128) but it is _technically_ handled (i.e. it's already in the state).

packages/uniforms/src/ValidatedForm.tsx
```diff
@@ -171,41 +150,54 @@ export function Validated<Base extends typeof BaseForm>(Base: Base) {
...
+    if (value instanceof Promise) {
+      callIfAsync();
+      return value.then(fn);
+    }
+
+    return fn(value);
+  }
+
+  return then;
+}
```
Monteth, line 199: That's brilliant 👑

Issue #717, opened 2020-05-01
Optimize AutoField


radekmie:
One performance issue that uniforms using new context API have, is the number of context consumers (more in #716). In a fully customized form (i.e. without `AutoField`s) there's no way to reduce their number (other than props drilling, of course). Let's think about `AutoField` then.

Currently, it's a smart wrapper, that can route the props to a correct component (it'll probably become smarter with #640). What we could do, is to make it render the component (`X` of `XField`) directly, skipping additional context access and props calculation. It's possible and non-distinguishable as long as the field doesn't contain more context consumers (e.g. fields). To make it opt-in, we could add a flag to `connectField` to mark such fields. It'd also exposing the wrapped component. Here's an example of how this could look like:

```tsx
const CustomTextInput = ...;
const CustomTextInputField = connectField(CustomTextInput, {
  // This means that there won't be any context consumer in the
  // rendered subtree and AutoField can render it directly.
  leaf: true
});

// Component and options passed to connectField.
// Are those part of the public API?
CustomTextInputField.Component;
CustomTextInputField.options;
```

It's colliding with `includeInChain: true`, those two couldn't be used together. This leads to another idea, raised in #139: separation of simple (here called _leaf_) and complex (called _relative_ in #139) fields. I think that there are actually three types:
1. _Leaf_ fields: actual fields, without form-related children, spiritually more advanced `<input>`s. These can definitely be directly rendered by `AutoField`.
2. _Wrapper_ fields: fields groups, similar to `ListField` and `NestField`. Currently, the ones using `includeInChain: false`. These cannot be directly rendered by `AutoField` (for example due to the propagation of `disabled` and other flags).
3. _Node_ fields: parts of the layout, complete blocks composed of fields. There are no such fields in the built-in themes. An example is the [`CompositeField`](https://uniforms.tools/docs/examples-custom-fields#compositefield) from the docs.

PR #721, opened 2020-05-09
Simplified ListField

radekmie:
As said in #717, there are different types of fields. `ListField` is very special due to the `ListAddField` and `ListDelField`. A typical `ListField` looks like this:

```tsx
// This is not valid usage! All names are absolute for presentational purposes!
<ListField name="list">
  <ListAddField name="list.$" />
  <ListItemField name="list.0">
    <ListDelField name="list.0" />
    <AutoField name="list.0.fieldA" />
    <AutoField name="list.0.fieldB" />
  </ListItemField>
  <ListItemField name="list.1">
    <ListDelField name="list.1" />
    <AutoField name="list.1.fieldA" />
    <AutoField name="list.1.fieldB" />
  </ListItemField>
</ListField>
```

The problem that `ListAddField` and `ListDelField` have is that they require parent (`list` in this example) props to work. In v2 every field receives its parent as a part of the guaranteed props. In v3 we removed that functionality, as every field can use `useField` hook multiple times to get all it needs. Well, it works in most cases, but here it may cause problems as `useField` always appends the name from the context, therefore there's no way of getting parent props at all. I see three ways of dealing with that:

1. Introduce a special symbol to make `joinName` _pop_ a name. Something like `joinName('list.0', POP) === 'list'`. It will cover this case and allow granular stepping up the name chain. The downside is it makes name handling more complex.
2. Introduce a flag for `useField` to treat the received name as a top-level one. It will cover this case and allow referencing arbitrary fields. The downside is that it may lead to some very complex relations, but we could teach users not to overuse that.
3. Introduce a new helper, which will work just like the `useField` does but with the `context` as a parameter instead of a name. It'd make the usage slightly more complex (`useField` + adjust the context + use the new helper), but also more performant (one `useContext` instead of two). 
4. Do nothing. This will make `ListField` a special and probably error-prone while implementing in user themes.

---

This is the second step of removing `includeInChain` option of `connectField`. For the first one, see #720.

### DIFF
--- a/packages/uniforms/src/useField.tsx
+++ b/packages/uniforms/src/useField.tsx
@@ -28,10 +28,14 @@ export function useField<
   Props extends Record<string, any>,
   Value = Props['value'],
   Model = Record<string, any>
->(fieldName: string, props: Props, options?: { initialValue?: boolean }) {
+>(
+  fieldName: string,
+  props: Props,
+  options?: { absoluteName?: boolean; initialValue?: boolean },
+) {
   const context = useForm<Model>();
 
-  const name = joinName(context.name, fieldName);
+  const name = joinName(options?.absoluteName ? '' : context.name, fieldName);
   const state = mapValues(context.state, (prev, key) => {
     const next = props[key];
     return next === null || next === undefined ? prev : !!next;
--- a/packages/uniforms/src/useField.tsx
+++ b/packages/uniforms/src/useField.tsx
@@ -18,7 +18,10 @@ function propagate(
   const resultValue =
     typeof prop === 'string'
       ? prop
-      : prop === false || (prop === undefined && !state) || schemaDisabled
+      : prop === null ||
+        prop === false ||
+        (prop === undefined && !state) ||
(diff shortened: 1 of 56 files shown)


radekmie:
After an internal discussion we've agreed to try out the _absolute name_ solution.

radekmie:
As discussed in today's meeting: @kestarumper will continue rewriting other themes.

packages/uniforms/src/useField.tsx
```diff
@@ -18,7 +18,10 @@ function propagate(
   const resultValue =
     typeof prop === 'string'
       ? prop
-      : prop === false || (prop === undefined && !state) || schemaDisabled
+      : prop === null ||
+        prop === false ||
+        (prop === undefined && !state) ||
+        schemaDisabled
```
Monteth, line 24: Can we extract it to function with descriptive name?

radekmie: Will do when we'll come up with one.

PR #739, opened 2020-06-03
Removed modelSync from AutoForm state.

radekmie:
The `modelSync` was introduced back in da56d32a9d0ee03e6a061a8b674492beac4b3fc6, without any additional comment. I remember it was required to make the async validation work at all. I've run some extensive testing and it looks like it's not the case in the current React version.

Performance-wise, this removes one complete re-render of the `AutoForm`, which is the most used form component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Next
+
+- **Breaking:** Removed `modelSync` from `AutoForm` state. [\#739](https://github.com/vazco/uniforms/issues/739)
+
 ## [v3.0.0-alpha.4](https://github.com/vazco/uniforms/tree/v3.0.0-alpha.4) (2020-06-03)
 
 - **Breaking:** Removed `injectName` helper. It was used to force context updates and got replaced by the new context directly. [\#720](https://github.com/vazco/uniforms/issues/720)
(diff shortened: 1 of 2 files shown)


mvgijssel:
Just ran into a problem with this particular change as `changed` and `changedMap` are never being updated by the BaseForm. This happens because the `model` state is updated immediately by AutoForm (this change) and only after this state change the `onChange` handler of the `BaseForm` is called. The `onChange` handler of the BaseForm now checks if there's a difference between the newly passed value and the currently value (https://github.com/vazco/uniforms/blob/v3/packages/uniforms/src/BaseForm.tsx#L185), but as the new state is already persisted this will never be the case.

radekmie:
That's a great finding, @mvgijssel! It does make sense. I'll definitely look into this next week - maybe we could call `onChange` earlier... I'll have to check that. Maybe you have an idea?

radekmie:
OK, it looks like it _just works_. I have to run some more tests, but that's for the next week. @mvgijssel - could you maybe test it in your project?

mvgijssel:
Hi @radekmie,

Did the following on https://github.com/vazco/uniforms/pull/765

```bash
git clone git@github.com:vazco/uniforms.git
cd uniforms/packages/uniforms/
git checkout `v3-autoform-changed`
yarn link 
# switch to other project
yarn link "uniforms"
```

And got the following error:

> Error: Invalid hook call. Hooks can only be called inside of the body of a function component.

image

Likely unrelated to this change and more a problem of how I included the dependency to the project. Will do some proper testing later this week, but the change seems logical! Calling the onChange before the changes are persisted to the `state`, so the diff in state works again.

radekmie:
@mvgijssel that's... Weird. I've added tests for that (and this change actually fixes them), but this problem seem unrelated. I'll wait for your findings!

Issue #716, opened 2020-05-01
Performance regressions on v3


radekmie:
Step by step, commit by commit - we're getting closer to a state, where v3 could be released (finally!). You don't have to run benchmarks to see, that the new version is far slower! Let's dissect the problems and think about possible solutions.

### Consequences of new context API
Using `useField`, or `useForm` directly, relies on `useContext` underneath. While the new context API is far better and more composable, it differs in the way of context propagation. The old context was more or less another set of props. This caused a lot of issues with other packages using `PureComponent`/`memo` (like #528). New one updates all consumers directly, skipping all nodes in between. This means, that there's no way to actually stop an update on the consumer side, and existing [`shouldComponentUpdate`](https://github.com/vazco/uniforms/blob/04e1ff427998f1369f39945ad416f443c2c0ff7a/packages/uniforms/src/BaseField.tsx#L93-L169) won't help us now.

As there's no React way to deal with that _yet_ (there are RFCs though, like https://github.com/reactjs/rfcs/pull/119), we have to do _something_. There are userspace implementations, like [dai-shi/use-context-selector](https://github.com/dai-shi/use-context-selector). I think we could use that, but the actual gains need to be confirmed in benchmarks. This would also mean, that the actual context usage may change with time, and we should no longer treat the context as a part of public API (of course, `useForm` and `useField` are still available).

### Consequences of new lifecycle APIs
Using `componentWillReceiveProps` one can access both previous and next props as well as the current state. New `getDerivedStateFromProps` differs a little: there's no access to previous props. This means, that the `BaseForm` cannot recreate the bridge only when needed! [I've signalized it already](https://github.com/vazco/uniforms/blob/0ad207a677e1566d426d0b60a206360808a46f80/packages/uniforms/src/BaseForm.tsx#L86-L92) and we should consider actually implementing that.

**EDIT:** Other solution can be actually removing `createSchemaBridge` completely. This would eliminate the awkward `Bridge.check` as well as the bridge duplication problem because one would have to create the bridge manually beforehand. It leads to a slightly more complicated usage with SimpleSchema, but it changes nothing with JSON schema GraphQL and schemas.

### Other problems
I'm sure there are more things we could do. We have to take the opportunity of making breaking changes (as small as possible!) and look for them now. One thing we could take a look at is the `modelSync` of `AutoForm` - it may be not needed with the Suspend API.

radekmie:
[...]

PR #741, opened 2020-06-06
Added kind option to connectField.

radekmie:
This change introduces a `kind: 'leaf' | 'node'` flag for the `connectField`, closing #717. This information is later used by `AutoField`, allowing it to skip one components layer (namely, `Field` created in `connectField`), effectively getting rid of one `useField` per field.

Rendering time (see #716) reduced noticeably, especially the upper bounds:
|  | v3 1f70af5f56235d4875611ee813c61aeaa2b0ae46 | v3 895130a2d89c709bf96e10473ffc1a8c375fa00b | v2 759a0a782bfbd1a6cad42bff3cbac8b4193fc136 |
| - | -:| -:| -:|
| **Nested** min | 42ms | 40ms | 55ms |
| **Nested** max | 70ms | 53ms | 62ms |
| **Flat** min | 51ms | 43ms | 22ms |
| **Flat** max | 73ms | 56ms | 57ms |

This is still a proof of concept, though. Things that are left:
- [x] ~Decide if `Component` and `options` are part of the public API.~ We'll introduce it as an internal API, therefore it could be changed without a major version change. In the following v3.x versions it may become public (and obtain stability guarantees).
- [x] ~Decide if `kind` is a good name and way of specifying that. Maybe something more specific like `mayHaveSubfield`, `isLeaf` or `plainField` would make more sense.~ We'll stick to `kind` for now.
- [x] Mark fields with `kind: 'leaf'` whenever possible.
- [x] Update other themes.
- [x] Write tests.
- [x] Update changelog.


packages/uniforms-bootstrap4/src/AutoField.tsx
```diff
@@ -57,5 +57,7 @@ export default function AutoField(originalProps: AutoFieldProps) {
     }
   }
 
-  return createElement(component, originalProps);
+  return 'options' in component && component.options?.kind === 'leaf'
```
kestarumper, line 60: Actually, why do we need `'options' in component` if we use `component.options?.` operator? Doesn't that mean conditional property access operator checks if `options` exist?

radekmie: It's required to narrow `ComponentType<any> | ReturnType<typeof connectField>` down to `ReturnType<typeof connectField>`.

PR #758, opened 2020-07-01
Material ui v4

kestarumper:
This PR
* updates `@material-ui/core` package to the newest (`4.9.9` -> `4.11.0`) version and drops support for lower versions (`^3.0.0 || ^1.2.0`)
* fixes an error that `material-ui` component starts as an uncontrolled when `value` prop is `undefined`


packages/uniforms-material/src/BoolField.tsx
```diff
@@ -2,7 +2,7 @@ import Checkbox, { CheckboxProps } from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormGroup from '@material-ui/core/FormGroup';
 import FormLabel from '@material-ui/core/FormLabel';
-import React, { ReactNode } from 'react';
+import React, { ReactNode, RefObject } from 'react';
```
radekmie, line 5: We do use `Ref` elsewhere.

packages/uniforms-material/src/BoolField.tsx
```diff
@@ -17,6 +17,7 @@ export type BoolFieldProps = Override<
     legend?: string;
     onChange?(value: any): void;
     transform?(label: string): string;
+    inputRef?: RefObject<HTMLButtonElement>;
```
radekmie, line 20: Keep these sorted.

packages/uniforms-material/src/SelectField.tsx
```diff
@@ -126,7 +127,7 @@ function Select(props: SelectFieldProps) {
                   name={name}
                   // FIXME: There's a problem with SelectFieldProps.
                   onChange={() => disabled || onChange!(xor(item, value))}
-                  ref={inputRef}
+                  inputRef={inputRef}
```
radekmie, line 130: It should stay as `ref`. `inputRef` is our internal name for forwarded ref.

Now there's `forwardRef` for that and we should consider using it. Anyway, it'll be a separate PR.

packages/uniforms-material/src/RadioField.tsx
```diff
@@ -46,7 +46,7 @@ function Radio({
       name={name}
       onChange={(event: any) => disabled || onChange(event.target.value)}
       ref={inputRef}
-      value={value}
+      value={value || ''}
```
radekmie, line 49: I'm not sure if `RadioField` works with numbers, but `??` may be better here (consider `0`).

packages/uniforms-material/src/RadioField.tsx
```diff
@@ -46,7 +46,7 @@ function Radio({
       name={name}
       onChange={(event: any) => disabled || onChange(event.target.value)}
       ref={inputRef}
-      value={value}
+      value={value ?? 0}
```
radekmie, line 49: Ah! I meant `?? ''` instead of `|| ''` because of the `0`.
```suggestion
      value={value ?? ''}
```

PR #763, opened 2020-07-05
Improved filterDOMProps return type

radekmie:
This change enhances `filterDOMProps` return type from `Partial<T>` to `Omit<T, keyof FilterDOMProps>`. In short, this allows us to properly exclude all filtered props, including the ones in the user app. New props have to be registered in the type as well:

```ts
declare module 'uniforms' {
  interface FilterDOMProps {
    customPropToFilter: never;
  }
}

filterDOMProps.register('customPropToFilter');
```

This showed a few unnecessarily passed props.

### DIFF
--- a/packages/uniforms/__tests__/filterDOMProps.ts
+++ b/packages/uniforms/__tests__/filterDOMProps.ts
@@ -11,6 +11,7 @@ describe('joinName', () => {
   });
 
   it('removes registered props', () => {
+    // @ts-ignore: Do not register its type not to pollute it.
     filterDOMProps.register('__special__');
 
     expect(filterDOMProps({ __special__: true })).toEqual({});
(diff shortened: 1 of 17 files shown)


packages/uniforms/__tests__/filterDOMProps.ts
```diff
@@ -11,11 +11,18 @@ describe('joinName', () => {
...
 
   it('removes registered props', () => {
+    // @ts-ignore: Do not register its type not to pollute it.
     filterDOMProps.register('__special__');
 
     expect(filterDOMProps({ __special__: true })).toEqual({});
   });
 
+  it('ignores double registers', () => {
+    const { length } = filterDOMProps.registered;
```
Monteth, line 21: Isn't it simpler? 
```suggestion
    const length = filterDOMProps.registered.length;
```

radekmie: Ah, I'm also never sure about the readability of that... The generated code is the same, though. Let's leave it for now - it's only in tests.

Issue #640, opened 2019-11-11
make it more easy to replace components without requiring to implement a custom AutoField


macrozone:
If you just want to replace one field (e.g. replace the ListField with a custom one for all schemas), you need to reimplement the whole AutoField.

It would probably easier, if you could pass a `getFieldComponent` to a AutoForm, which would hook into here: https://github.com/vazco/uniforms/blob/master/packages/uniforms-material/src/AutoField.tsx#L24 and would take `props` and could return a Component. If no component is returned, it would fallback to the normal logic.

radekmie:
Hi, @macrozone. I'm somewhat against it, as I don't like the idea of modifying it by reference. The broader problem is that `AutoField` renders `NestField` (or `ListField`), which renders `AutoField` again. Due to this cycle, it's not obvious who should change. For now, a custom theme is a go-to choice for most cases; another one is to automatically set the `component` prop at the schema level for all fields.

In the future (v3), it'll be easier to extend the context (and typing it!), so maybe the `AutoField` (or the function to determine the component?) could be a part of the context data.

macrozone:
@radekmie i don't understand the reasoning at all, but let's see what v3 brings

radekmie:
You can easily create a custom `AutoField` right now - simply extend the current one and provide your custom `render` method, falling back to the `super.render` if needed. Passing it to the form is also possible. The problem is that if it'll render `NestField`, this `NestField` will render the default one, imported from the theme.

macrozone:
> You can easily create a custom `AutoField` right now - simply extend the current one and provide your custom `render` method, falling back to the `super.render` if needed. Passing it to the form is also possible. The problem is that if it'll render `NestField`, this `NestField` will render the default one, imported from the theme.

ah yes, that's make it more clear. And in v3 the fields will come from the context?

radekmie:
Not sure if they _will_, but it'll be easier to do so, probably even in the project scope (no core change will be needed).

Floriferous:
We'd love something like this as well, we have pretty much rewritten all the uniforms-material components at this point, sometimes just to change one or two lines. 

If there was a way to at least only rewrite components once, and then have them used by all recursive calls would be great. 

Maybe an API like this could be helpful:

```js
import { makeAutoField, FIELD_TYPES } from 'uniforms';

const AutoField = makeAutoField({
  getComponent: type => {
    if (type === FIELD_TYPES.NESTED) {
      return CustomListField;
    }
[...]

PR #764, opened 2020-07-06
Added FieldProps types

### LINKED ISSUE #760: Feature (v3): Add Field properties type
In our project we defined this:

```ts
export type FieldProps<Value, Extra = {}, InputElementType = HTMLInputElement, ElementType = HTMLDivElement> = Override<
    HTMLProps<ElementType>,
    GuaranteedProps<Value> & {
        inputRef?: Ref<InputElementType>;
    } & Extra
>;
```
Which lets met define the properties for most custom components like this 
```ts
export type CustomFieldProps = FieldProps<string>;
```
or for more complex Fields with extra props:
```ts
export type SelectFieldProps = FieldProps<
    string | string[],
    { allowedValues?: string[]; checkboxes?: boolean; transform?(value: string): string },
    HTMLSelectElement,
    HTMLSelectElement,
>;
```

It might be a good idea to add this to uniforms. This makes it clearer which props are specific for the component and which are generic

BTW: I was a bit suprised that `inputRef` is not part of `GuaranteedProps` is that intentional?

radekmie:
This change adds two types: `FieldProps` and `HTMLFieldProps`, proposed by @hmvp. Closes #760. In short, these types are meant to make field types more compact and easier to maintain in case of a change of guaranteed props types.

Having said that, I'm still not sure about the status of any exported types being a part of the public API. Of course, they'll be available and importable, but we'd have to explicitly mark the semver-compatible ones. For example, I consider the `Override` type internal and I don't think we should maintain its backward compatibility.

### DIFF
--- a/packages/uniforms/src/types.ts
+++ b/packages/uniforms/src/types.ts
@@ -30,6 +30,9 @@ export type DeepPartial<T> = {
   [P in keyof T]?: T[P] extends {} ? DeepPartial<T[P]> : T[P];
 };
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface FilterDOMProps {}
+
 export type GuaranteedProps<Value> = {
   changed: boolean;
   disabled: boolean;
@@ -54,3 +57,27 @@ export type Override<T, U> = U & Omit<T, keyof U>;
 export type Partialize<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
 
 export type ValidateMode = 'onChange' | 'onChangeAfterSubmit' | 'onSubmit';
+
+// Explicit declaration to allow users extend it directly.
+// Would it be possible not to use `declare` here?
+declare module '.' {
+  interface FilterDOMProps {
+    allowedValues: never;
+    changed: never;
+    component: never;
+    disabled: never;
+    error: never;
+    errorMessage: never;
+    field: never;
+    fieldType: never;
+    fields: never;
+    initialCount: never;
+    label: never;
+    name: never;
+    onChange: never;
+    placeholder: never;
+    showInlineError: never;
+    transform: never;
+    value: never;
+  }
+}
--- a/packages/uniforms/src/types.ts
+++ b/packages/uniforms/src/types.ts
@@ -1,4 +1,4 @@
-import { ReactNode, SyntheticEvent } from 'react';
+import { HTMLProps, ReactNode, SyntheticEvent } from 'react';
 
 import { Bridge } from './Bridge';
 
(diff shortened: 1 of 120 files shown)


packages/uniforms/src/types.ts
```diff
@@ -30,6 +30,11 @@ export type DeepPartial<T> = {
   [P in keyof T]?: T[P] extends {} ? DeepPartial<T[P]> : T[P];
 };
 
+export type FieldProps<Value, Base, Extension = {}> = Override<
+  Base,
+  GuaranteedProps<Value> & Extension
+>;
+
```
hmvp, line 37: Why `FieldProps` If its only used in HTMLFieldProps?

packages/uniforms/src/types.ts
```diff
@@ -50,6 +55,12 @@ export type GuaranteedProps<Value> = {
   value?: Value;
 };
 
+export type HTMLFieldProps<Value, Element, Extension = {}> = FieldProps<
```
hmvp, line 58: I chose the order Value, Extension, Element and gave Element a default because most fields use div as an outer wrapper and thus that argument can often be left out, while most fields have something for the extension. Not to disagree with your choice, just explaining mine :slightly_smiling_face:

hmvp:
> This change adds two types: `FieldProps` and `HTMLFieldProps`, proposed by @hmvp. Closes #760. In short, these types are meant to make field types more compact and easier to maintain in case of a change of guaranteed props types.
> 
> Having said that, I'm still not sure about the status of any exported types being a part of the public API. Of course, they'll be available and importable, but we'd have to explicitly mark the semver-compatible ones. For example, I consider the `Override` type internal and I don't think we should maintain its backward compatibility.

I feel that anything you use within `uniforms-unstyled` (and the other widget packages) should be considered a somewhat public API since it is probably not uncommon (and in the docs) to "fork"/extend one or the whole set of components..

packages/uniforms/src/types.ts
radekmie: It's going to be used in `uniforms-antd` and `uniforms-material` as they use custom components.

packages/uniforms/src/types.ts
radekmie: I've thought about it as well, but I see _some_ value in explicitly setting the `Element`.

radekmie:
> I feel that anything you use within uniforms-unstyled (and the other widget packages) should be considered a somewhat public API since it is probably not uncommon (and in the docs) to "fork"/extend one or the whole set of components..

Good point. I think we have to have a stable position on this topic before v3 gets released. Maybe we'll treat types differently than the code? I still don't know.

Another thing is that with #640 it's somehow easier to create a new theme and we can have a slightly different approach here.

Issue #785, opened 2020-08-12
ClassName for SelectField


michael-vasyliv:
It seems like ``className`` does not work for SelectField
inside component
![Screenshot from 2020-08-12 15-38-24](https://user-images.githubusercontent.com/19407350/90015925-f242af00-dcb1-11ea-8a34-521131d0c5eb.png)

in HTML 
![Screenshot from 2020-08-12 15-38-04](https://user-images.githubusercontent.com/19407350/90015957-fbcc1700-dcb1-11ea-9f49-dfe111145110.png)

additional information:
"preact-cli": "3.0.0-rc.7",
"preact": "10.3.4",
"preact-render-to-string": "5.1.4",
"preact-router": "^3.1.0",
"uniforms": "^3.0.0-rc.0",
"uniforms-bootstrap4": "^3.0.0-rc.0",
"uniforms-bridge-json-schema": "^3.0.0-rc.0",

radekmie:
Hi @michael-vasyliv. I can see that there's an incorrect props type, but besides that - do you think that `className` is needed? There are others, specialized ones.

michael-vasyliv:
@radekmie there are so many classes and no one fits me, I want to add class beside ``form-group``. As I understand SelectField extends from BaseField that takes ``className`` as a prop so in my opinion, it should work but I have a bug.

michael-vasyliv:
@radekmie I took all classes that are discribed in https://uniforms.tools/docs/api-fields#selectfield (all props that have suffix className) and no one adds ``className`` beside ``form-group``

michael-vasyliv:
@radekmie or are least there is a question is it a bug or its a designed feature (not accept className in SelectField)? because it works fine for TextField, bug for SelectField it does not work
![Screenshot from 2020-08-12 18-35-07](https://user-images.githubusercontent.com/19407350/90035565-9cc6cc00-dcca-11ea-84b2-4f1d98decb1e.png)

radekmie:
@michael-vasyliv That's odd - we'll check that.

Issue #793, opened 2020-08-21
Select does not show inline error


michael-vasyliv:
Select does not show inline error in uniforms-bootstrap4 and uniforms-bootstrap3
[playground](https://uniforms.tools/playground/#?N4IgDgTgpgzlAuIBcIDqUA2BjA9gWygEIQAacCHMGZUAQwFd4cZaA3KZAM1ozjIaYt2AEUy0AnsgCMABhlkAJgEsWAIwxQFXHnxAZaqzMngR6UMmH1YoACxwYFUCNt7mQMLDah5ayEADsoAHcAAgBlJTxLKDDPb1oAJgAhCCUFAHMoAAoAHX8QgpDA0IiojVivHyzgPMK6kNr6gpglAC8oJBCa_Kb6-HEwDvCTJX90kkbewp4MHCDNADUeMxhOgG0AcgAPGA2SEI3d_Y28PYOMM-2LgF1J-oBfCf87-4BKPNfSdzsggEl_DCjKAAUQgFGcSBMZn4MHE_iwAHl_GF6Ko8EpEEhuK4YXDEf4loCFLR4BwsTooI8QPBKmSQKocDh4DATLQwAAWED3IA)

here is result of inline error in uniforms-bootstrap4
![Screenshot from 2020-08-21 15-45-53](https://user-images.githubusercontent.com/19407350/90892129-80f7b000-e3c5-11ea-9192-17f27ec56495.png)

here is result of inline error in uniforms-antd
![Screenshot from 2020-08-21 15-46-24](https://user-images.githubusercontent.com/19407350/90892137-835a0a00-e3c5-11ea-984f-df098b6195cd.png)

as you can see there is text under input in antd but not in bootstrap4

kestarumper:
Hi @michael-vasyliv,
We'll check this out and if bootstrap supports inline errors we will fix that.

michael-vasyliv:
@kestarumper Hello! I've found the bug and fixed it in the #795

PR #795, opened 2020-08-28
#785 class name for select field

michael-vasyliv:
#785
This PR will fix issue #785 and probably #793, the problem was in a [Spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the ``SelectField`` for bootstrap4, before pass props to ``wrapField`` few properties were removed from props, that is why in ``wrapField`` there were undefined.

### DIFF
--- a/packages/uniforms-bootstrap4/src/SelectField.tsx
+++ b/packages/uniforms-bootstrap4/src/SelectField.tsx
@@ -24,28 +24,34 @@ export type SelectFieldProps = HTMLFieldProps<
   }
 >;
 
-function Select(props: SelectFieldProps) {
-  const {
-    allowedValues,
-    checkboxes,
-    disabled,
-    error,
-    fieldType,
-    id,
-    inline,
-    inputClassName,
-    inputRef,
-    label,
-    name,
-    onChange,
-    placeholder,
-    required,
-    transform,
-    value,
-  } = props;
-
+function Select({
+  allowedValues,
+  checkboxes,
+  disabled,
+  error,
+  fieldType,
+  id,
+  inline,
+  inputClassName,
+  inputRef,
+  label,
+  name,
+  onChange,
+  placeholder,
+  required,
+  transform,
+  value,
+  ...props
+}: SelectFieldProps) {
   return wrapField(
-    props,
+    {
+      ...props,
+      disabled,
+      error,
+      id,
+      label,
+      required,
+    },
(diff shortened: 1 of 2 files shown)


packages/uniforms-bootstrap4/src/SelectField.tsx
```diff
@@ -24,30 +24,28 @@ export type SelectFieldProps = HTMLFieldProps<
...
+    onChange,
+    placeholder,
+    required,
+    transform,
+    value,
+  } = props;
+
   return wrapField(
-    { ...props, id, label },
+    props,
```
radekmie, line 48: This means that the `wrapField` receives a lot of unnecessary props. It'd be much better to list (and pass here) only the ones that are actually needed.

I'm also quite sure that in order to fix #785, it'd be enough to remove line 30 (`className`) as it's not used anywhere. The same goes for #793 and line 44 (`showInlineError`).

It's weird that our ESLint did not catch that. Also, ping @kestarumper and @Monteth.

radekmie (changes requested):

packages/uniforms-bootstrap4/src/SelectField.tsx
michael-vasyliv: wrapField requires not only these properties
```
className
showInlineError
```
as you can see from this list https://github.com/vazco/uniforms/blob/master/packages/uniforms-bootstrap4/src/wrapField.tsx#L26-L38

we should pass more properties such as:
```
className,
disabled,
error,
errorMessage,
required,
showInlineError
```

radekmie: Then these should be added here as well. It's definitely better to list them here rather than pass all.

michael-vasyliv: @radekmie  next props are used only in ``wrapField``
```
className
showInlineError
errorMessage
```
next props are used in ``wrapField`` and ``Select``
```
disabled,
error,
required
```

radekmie (approved):
Thanks again, @michael-vasyliv!

michael-vasyliv:
@radekmie you are welcome. When these changes will go to a new version?

radekmie:
I'm working on it.

michael-vasyliv:
@radekmie thank you.

PR #796, opened 2020-08-31
Add missing JSON Schema properties (closes #748).

### LINKED ISSUE #748: Feature: Missing JSON Schema properties
While implementing a form based on JSON Schema generated by [pydantic](https://github.com/samuelcolvin/pydantic/). I noticed a few JSON Schema properties that are not handled. See also [their docs](https://pydantic-docs.helpmanual.io/usage/schema/#json-schema-types) about JSON Schema properties.

Effectively I need to override `getProps` in the bridge with:
```ts
    getProps(name: string) {
        let props = super.getProps(name);

        if (props.const) {
            props.disabled = true;
            props.default = props.const;
            delete props["const"];
        }

        if (props.maxItems !== undefined) {
            props.maxCount = props.maxItems;
            delete props["maxItems"];
        }

        if (props.minItems !== undefined) {
            props.minCount = props.minItems;
            delete props["minItems"];
        }

        if (props.maximum !== undefined) {
            props.max = props.maximum;
            delete props["maximum"];
        }

        if (props.minimum !== undefined) {
            props.min = props.minimum;
            delete props["minimum"];
        }

        if (props.multipleOf !== undefined) {
            props.step = props.multipleOf;
            delete props["multipleOf"];
        }

        return props;
    }
```

I am not sure about the supported JSON Schema specs and whether these properties are in scope of being supported by default. And I am not sure if it is a valid asumption that `const` is read only. But I feel at least some might be worth to implement in the base JSONSchemaBridge. 

I don't know if this list is complete, its what are missing atm

Monteth:
Add JSON Schema properties: `maxCount`, `minCount`, `max`, `min`, `step`.

### DIFF
--- a/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
@@ -28,16 +28,12 @@ function resolveRef(reference: string, schema: Record<string, any>) {
   return resolvedReference;
 }
 
-const propMapper = {
+const propMapper: Record<string, string> = {
   maxItems: 'maxCount',
   minItems: 'minCount',
   maximum: 'max',
   minimum: 'min',
   multipleOf: 'step',
-};
-
-const isKey = (candidate: string): candidate is keyof typeof propMapper => {
-  return candidate in propMapper;
 };
 
 function distinctSchema(schema: Record<string, any>) {
@@ -284,7 +280,7 @@ export default class JSONSchemaBridge extends Bridge {
     }
 
     Object.keys(ready).forEach(key => {
-      if (isKey(key)) {
+      if (key in propMapper) {
         const newKey = propMapper[key];
         ready[newKey] = ready[key];
         delete ready[key];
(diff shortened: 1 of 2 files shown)


packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -28,6 +28,33 @@ function resolveRef(reference: string, schema: Record<string, any>) {
...
+  },
+  multipleOf(props: Record<string, any> = {}) {
+    props.step = props.multipleOf;
+    delete props['multipleOf'];
+  },
+};
+
+const isKey = (candidate: string): candidate is keyof typeof propMapper => {
+  return candidate in propMapper;
+};
```
radekmie, line 56: As discussed - a simple mapping `{minimum: 'min', ...}` would be enough.

radekmie (changes requested):

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -271,6 +283,14 @@ export default class JSONSchemaBridge extends Bridge {
       }
     }
 
+    Object.keys(ready).forEach(key => {
+      if (isKey(key)) {
```
radekmie, line 287: If you'd add `: Record<string, string>` to `propMapper`, then this function could be replaced with `key in propMapper`.

radekmie (approved):

PR #797, opened 2020-09-01
Add Commercial Services section on the website

MacRusher:
This PR will add a new section on the uniforms.tools website.

### DIFF
--- a/website/index.module.css
+++ b/website/index.module.css
@@ -256,6 +256,31 @@
   margin-top: -8em;
 }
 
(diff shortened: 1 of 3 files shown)


radekmie (approved):

Monteth (approved):
Most of the sections have icons, forms, or images, and this section looks a little boring in comparison as it contains mostly text. Its OK for me, but it would be nice if we can improve it in the future 🚀

PR #800, opened 2020-09-10
Implemented custom AutoField prototype.

### LINKED ISSUE #640: make it more easy to replace components without requiring to implement a custom AutoField

radekmie:
This change provides a new implementation of `AutoField` and an according `AutoFieldContext` to override field detection implementation. Internally a `createAutoField` was added as well in order to reduce the amount of boilerplate theme code to a minimum.

Open questions:
* Is `AutoFieldContext` a good name? Maybe `AutoField.context` or `AutoField.componentDetectorContext` is a better one?
* What about `createAutoField` - should we consider it an internal API for now/forever?
* Right now overriding the default detector requires always return a component. In order to fall back to the default one, one would have to either fetch the default one with `useContext` or implement it in their own project. Another solution would be to treat the provided one with a higher priority and fallback to the default one.

Closes #640.

---

Update 2020-12-21:
* I've decided to move `AutoFieldContext` to `AutoField.componentDetectorContext` in order to reduce API surface. We may change it in the future.
* All `createAutoField`, `AutoField.componentDetectorContext`, and `AutoField.defaultComponentDetector` are considered internal for now.
* Default detector is now accessible through `AutoField.defaultComponentDetector`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,9 +12,9 @@ jobs:
         node-version: [10.x, 12.x, 14.x]
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.3.1
+        uses: actions/checkout@v2.3.4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2.1.1
+        uses: actions/setup-node@v2.1.2
         with:
           node-version: ${{ matrix.node-version }}
       - name: Install
@@ -24,12 +24,12 @@ jobs:
       - name: Test
         run: npm run coverage -- --no-cache --runInBand
       - name: Report coverage
-        uses: codecov/codecov-action@v1.0.12
+        uses: codecov/codecov-action@v1.0.15
       - name: Build docs
         run: npm --prefix website run build
         if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.node-version == '14.x'
       - name: Deploy docs
-        uses: peaceiris/actions-gh-pages@v3.6.4
+        uses: peaceiris/actions-gh-pages@v3.7.3
         with:
           allow_empty_commit: true
           cname: uniforms.tools
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## [v3.0.0-rc.8](https://github.com/vazco/uniforms/tree/v3.0.0-rc.8) (2020-12-05)
+
+- **Breaking:** Due to synchronization of `AutoForm` `model`, autosave is now **always** asynchronous. [\#832](https://github.com/vazco/uniforms/issues/832)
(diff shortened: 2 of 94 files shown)


Monteth:
1.)  `AutoFieldContext` seems like good name for that. 
2.) I don't see why someone would use that, maybe we shouldn't introduce a flood of API that no one will use.
3.) I would say many users may want to replace only one/very few components and they won't write the whole detector as most of its job is still correct.

kestarumper:
1. After some thought, `AutoFieldContext` seems to be the right choice. Imagine writing `AutoField.context.Provider` - it looks strange.
2. I actually like `createAutoField` helper. It reduces boilerplate code and makes sure that we use `connectField(..., { kind: 'leaf' })` for performance reasons. Most of the time, overriding `AutoField` (by writing it yourself) leads to reinventing the wheel. With the use of `createAutoField`, I can focus on **selecting the right component**  based on its props - and that's all that I need.
3. I don't know if that's overkill, but we could `export` `defaultComponentDetector` which accepts a `map` of components
```javascript
{
    BoolField: MyCustomBoolField,
...
}
```
Then if someone would only want to override (or create a new theme), one would pass overridden map and wouldn't have to change the `switch` logic.

If component return logic is more complicated than based on the default, one could write its own `componentDetector` and if needed use and fallback to our exported `defaultComponentDetector`.

kestarumper (dismissed):

radekmie:
@kestarumper 
1. I can imagine `AutoField.context.Provider`. And I think it's actually better than a new export from the theme.
2. No, `createAutoField` won't add `kind: 'leaf'` for you - it'll only use it when it's there.
3. The `defaultComponentDetector` would have to be a function already ([see the `allowedValues` branch](https://github.com/vazco/uniforms/blob/7c9f23a8b1d66a415d78aa1203bb1dd5591a38e7/packages/uniforms-unstyled/src/AutoField.tsx#L16-L20)). There's no need for it anyway, if we'd go with _default if `undefined`_.

kestarumper:
> No, createAutoField won't add kind: 'leaf' for you - it'll only use it when it's there.

I meant only that if you have your own implementation of `AutoField`, then if you don't do the following:
```javascript
return 'options' in component && component.options?.kind === 'leaf'
      ? createElement(component.Component, props)
      : createElement(component, originalProps);
```
then It's lost.

radekmie:
I've updated the description. If everything will be fine, I'll update all themes.

Monteth (dismissed):

kestarumper (dismissed):

Monteth (dismissed):

Monteth (approved):

PR #803, opened 2020-09-16
Disable <SelectField /> items based on predicate (closes #736)

### LINKED ISSUE #736: Ability to specify disabled items in SelectField
Currently, there's no way of disabling particular items in SelectField.

One should write a custom field in order to achieve something like this:

image

It would be nice to have such an ability by default.

f.e.

```
<SelectField
   name="selectField"
   disableItem={(item) => ...}
/>
```
or

```
<SelectField
   name="selectField"
   disableItems={["Item 2", "Item 4"]}
/>
```

Thanks!

kestarumper:
This PR adds the ability to disable certain items based on predicate (similar to `.filter(item => condition(item))`) that can be specified as a prop.

Example: _Disable all odd values_
```javascript
<SelectField
  name="x"
  disableItem={value => value % 2}
/>
```


packages/uniforms-antd/src/SelectField.tsx
```diff
@@ -4,7 +4,10 @@ import CheckboxGroup, {
 } from 'antd/lib/checkbox/Group';
 import { RadioGroupProps } from 'antd/lib/radio';
 import RadioGroup from 'antd/lib/radio/group';
-import SelectAntD, { SelectProps as SelectAntDProps } from 'antd/lib/select';
+import SelectAntD, {
+  SelectProps as SelectAntDProps,
+  SelectValue,
```
radekmie, line 9: It's not used here and linter passed... There's some misconfiguration.

packages/uniforms-material/src/wrapField.tsx
```diff
@@ -41,4 +37,22 @@ function wrap(
...
+}
+
+filterDOMProps.register(
+  'checkboxes',
+  'disableItem',
+  'fullWidth',
+  'helperText',
+  'margin',
+  'variant',
+);
```
radekmie, line 58: I'd say it should be extracted to a separate PR.

radekmie (changes requested):

packages/uniforms-material/src/wrapField.tsx
kestarumper: As discussed in a meeting, it's good that it's fixed but in the future let's split unrelated changes to other PRs.

radekmie (approved):
I'm still thinking if we could come up with a better name.

Monteth (approved):

PR #809, opened 2020-10-12
Added prop autocomplete

mariusrak:
I've added autocomplete property for TextFields into standard templates.


packages/uniforms-antd/src/TextField.tsx
```diff
@@ -14,6 +14,7 @@ function Text(props: TextFieldProps) {
   return wrapField(
     props,
     <Input
+      autoComplete={props.autoComplete}
```
radekmie, line 17: Is this one actually needed? It should not as `{...filterDOMProps(props)}` below does not remove `autoComplete`. Same goes for `uniforms-material` theme.

mariusrak: I wanted to use it with _-unstyled_ but it was assigned to `<div>` instead of `<input>`

radekmie: Yeah, I see that both `uniforms-semantic` and `uniforms-unstyled` required this change, hence I'm asking about `uniforms-antd` and `uniforms-material`. Also, what about `uniforms-bootstrap3` and `uniforms-bootstrap4`? Could you add tests there as well?

mariusrak: I'm not sure about _-antd_ and _-material_, but I've expected same behavior from them as _-unstyled_. I've missed the bootstrap versions, so I'll look at them also.

mariusrak: So, in new commits, tests are added and if it is not needed to explicitly set the value it is not there.

packages/uniforms-bootstrap4/src/TextField.tsx
```diff
@@ -18,6 +18,7 @@ function Text(props: TextFieldProps) {
       className={classnames(props.inputClassName, 'form-control', {
         'is-invalid': props.error,
       })}
+      autoComplete={props.autoComplete}
```
radekmie, line 21: I've just checked and while both Bootstrap themes correctly understand `autoComplete` now, it also appears on the `div` rendered by `wrapField`. To fix that, do add `autoComplete` to the list of omitted values above.

radekmie (changes requested):

packages/uniforms-bootstrap4/src/TextField.tsx
mariusrak: Fixed and added tests for that, hopefully it's okay now.

radekmie (approved):

PR #818, opened 2020-10-24
Fixed non-string labels (fixes #816).

### LINKED ISSUE #816: JSX labels are ignored in v3
The following code JSX renders a field without label:
```tsx
<TextField label={<span style={{ color: 'red' }}>Fancy Label</span>} />
```

radekmie:
This change fixes a bug reported in #816 as well as adds a broad set of tests for `label` handling. We should do the same for `placeholder`, but this will come separately.

### DIFF
--- a/packages/uniforms/src/useField.tsx
+++ b/packages/uniforms/src/useField.tsx
@@ -1,30 +1,31 @@
 import get from 'lodash/get';
 import mapValues from 'lodash/mapValues';
-import { useCallback, useEffect, useMemo } from 'react';
+import { ReactNode, useCallback, useEffect, useMemo } from 'react';
 
 import { joinName } from './joinName';
 import { GuaranteedProps } from './types';
 import { useForm } from './useForm';
 
 function propagate(
-  prop: unknown,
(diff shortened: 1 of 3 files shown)


ArchiCroc:
This will be a nice addition. I was just troubleshooting this when I tried to pass i18n components for label, placeholder, and extra.

packages/uniforms/src/useField.tsx
```diff
@@ -112,7 +113,8 @@ export function useField<
     ...props,
     label,
     name,
-    placeholder,
+    // TODO: Should we assert `typeof placeholder === 'string'`?
```
Monteth, line 116: I don't think so, a placeholder in Antd is of the `ReactNode` type, so we shouldn't assert it so tight.

Monteth (approved):

packages/uniforms/src/useField.tsx
radekmie: That's true, however, widening `placeholder` from `string` to` ReactNode` will require casts or asserts in all HTML fields.

PR #824, opened 2020-10-30
Add Multiselect (closes #737)

### LINKED ISSUE #737: Multi-select field
Hi
 First all: Kudos and Thank you for the vazco/uniforms package. Im using it for first time in a project and I found it really helpful.
 
 I was trying to get a combo-box for selecting multiple items, but I was only able to get a list of checkboxes (or radios). As the length of my allowed-values is more than 20, having a list of checkboxes makes the form not so appealing.
I found this issue https://github.com/vazco/uniforms/issues/34 but still not sure why Uniforms doesn't provide a MultiSelect combobox.

Even more, based on https://github.com/vazco/uniforms/blob/944645fe6d2929e195238e910d2d29fc64ca3083/packages/uniforms-bootstrap4/src/SelectField.tsx
I have modified the renderSelect to be
```

const renderSelect = props => (
  <select
    className={classnames(props.inputClassName, 'c-select form-control', {
      'is-invalid': props.error,
    })}
    disabled={props.disabled}
    id={props.id}
    name={props.name}
    multiple={props.fieldType === Array}
    onChange={event => {
      const item = event.target.value;
      props.onChange(
        props.fieldType === Array ? xor(item, props.value) : item,
      );
    }}
    ref={props.inputRef}
    value={props.value}
  >
    {(!!props.placeholder || !props.required || props.value === '') && (
      <option value="" disabled={props.required} hidden={props.required}>
        {props.placeholder || props.label}
      </option>
    )}

    {props.allowedValues.map(value => (
      <option key={value} value={value}>
        {props.transform ? props.transform(value) : value}
      </option>
    ))}
  </select>
);

const Select = props =>
  wrapField(
    props,
    renderSelect(props),
  );
```
and I've got a working multi-select combo-box on my project.

So, basically my question is: why doesn't Uniforms provide such multiple-select out of the box (at least for bootstrap)?

Monteth:
Add multiselect feature for select input.

### DIFF
--- a/packages/uniforms-bootstrap3/src/SelectField.tsx
+++ b/packages/uniforms-bootstrap3/src/SelectField.tsx
@@ -104,7 +104,7 @@ function Select({
           </option>
         )}
 
-        {allowedValues?.map(value => (
+        {allowedValues!.map(value => (
           <option disabled={disableItem?.(value)} key={value} value={value}>
             {transform ? transform(value) : value}
           </option>
--- a/packages/uniforms-bootstrap3/src/SelectField.tsx
+++ b/packages/uniforms-bootstrap3/src/SelectField.tsx
@@ -104,7 +104,7 @@ function Select({
           </option>
         )}
 
-        {allowedValues!.map(value => (
+        {allowedValues?.map(value => (
           <option disabled={disableItem?.(value)} key={value} value={value}>
             {transform ? transform(value) : value}
           </option>
--- a/packages/uniforms-bootstrap3/src/SelectField.tsx
+++ b/packages/uniforms-bootstrap3/src/SelectField.tsx
@@ -48,9 +48,10 @@ function Select({
   value,
   ...props
 }: SelectFieldProps) {
+  const multiple = fieldType === Array;
   return wrapField(
     { ...props, id, label },
-    checkboxes || fieldType === Array ? (
+    checkboxes ? (
       allowedValues?.map(item => (
         <div
           key={item}
@@ -83,14 +84,21 @@ function Select({
         })}
         disabled={disabled}
         id={id}
+        multiple={multiple}
         name={name}
-        onChange={event =>
-          onChange(event.target.value !== '' ? event.target.value : undefined)
-        }
+        onChange={event => {
+          const item = event.target.value;
+          if (multiple) {
+            const clear = event.target.selectedIndex === -1;
+            onChange(clear ? [] : xor([item], value));
+          } else {
+            onChange(item !== '' ? item : undefined);
+          }
+        }}
         ref={inputRef}
         value={value ?? ''}
       >
-        {(!!placeholder || !required || value === undefined) && (
+        {(!!placeholder || !required || value === undefined) && !multiple && (
           <option value="" disabled={required} hidden={required}>
             {placeholder || label}
           </option>
--- a/packages/uniforms-bootstrap4/__tests__/SelectField.tsx
+++ b/packages/uniforms-bootstrap4/__tests__/SelectField.tsx
@@ -325,6 +325,111 @@ test('<SelectField> - disabled items (options) based on predicate', () => {
   expect(wrapper.find('option[value="b"]').at(0).prop('disabled')).toBe(false);
 });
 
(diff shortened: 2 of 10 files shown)


packages/uniforms-bootstrap4/src/SelectField.tsx
```diff
@@ -87,22 +87,38 @@ function Select({
...
+        {allowedValues?.map(allowedValue => (
+          <option
+            disabled={disableItem?.(allowedValue)}
+            key={allowedValue}
+            value={allowedValue}
+            onClick={event => {
+              if (fieldType === Array) {
+                const item = event.currentTarget.value;
+                onChange(fieldType === Array ? xor([item], value) : item);
+              }
```
radekmie, line 118: I think that `onClick` here is not the best idea - it may have problems on phones (probably) as well as when used with keyboard (definitely). As I remember, the problem was to detect if nothing is selected. Try out the [`selectedIndex`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex) property, that is `-1` if nothing is selected.

packages/uniforms-bootstrap4/src/SelectField.tsx
```diff
@@ -87,22 +84,33 @@ function Select({
...
+        multiple={multiple}
         name={name}
-        onChange={event =>
-          onChange(event.target.value !== '' ? event.target.value : undefined)
-        }
+        onChange={event => {
+          const item = event.target.value;
+          if (multiple) {
+            const clear = event.target.selectedIndex === -1;
+            onChange(multiple ? (clear ? [] : xor([item], value)) : item);
```
radekmie, line 93:
```suggestion
            onChange(clear ? [] : xor([item], value)));
```

packages/uniforms-bootstrap4/__tests__/SelectField.tsx
```diff
@@ -325,6 +325,117 @@ test('<SelectField> - disabled items (options) based on predicate', () => {
...
+        'x.$': { type: String },
+      },
+      { onChange },
+    ),
+  );
+
+  expect(wrapper.find('select')).toHaveLength(1);
+  expect(
+    wrapper.find('select').simulate('change', { target: { value: 'a' } }),
+  ).toBeTruthy();
```
radekmie, line 435: Is this test correct? It doesn't need to include `selectedIndex`?

radekmie (dismissed):
Looks good! Now let's do it in other themes.

packages/uniforms-bootstrap4/__tests__/SelectField.tsx
Monteth: Seems like uncheck the last checked element generates a target without value. Indeed we have to add a target with `selectedIndex` to represent real behavior.

radekmie (approved):

Monteth:
I struggled to add tests in uniforms-material but then I realized there is no logic in multiselect to be tested on our side. So there we have completed and tested multiselect feature.

packages/uniforms-bootstrap3/src/SelectField.tsx
```diff
@@ -83,20 +84,27 @@ function Select({
...
       >
-        {(!!placeholder || !required || value === undefined) && (
+        {(!!placeholder || !required || value === undefined) && !multiple && (
           <option value="" disabled={required} hidden={required}>
             {placeholder || label}
           </option>
         )}
 
-        {allowedValues?.map(value => (
+        {allowedValues!.map(value => (
```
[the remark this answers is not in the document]
Monteth: I think the `allowedValues` should be mandatory in `SelectField` and I don't get why it's like that right now - maybe for easier use of predefined fields. I will switch to optional chaining and let's see what @radekmie would say.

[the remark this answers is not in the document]
radekmie: The thing is that `allowedValues` are required, but not in the `props`. Most often you'll provide those in the schema or through `options`, not `props`.

radekmie (approved):

PR #830, opened 2020-11-25
Organized props filtering (fixes #815)

### LINKED ISSUE #815: [BUG] filterDOMProps and uniforms-material package
Hi,

Did something change in the **filterDOMProps** and **uniforms-material package** in **3.0.0-rc.5**?

I have this example:

```
console.log(filterDOMProps({
      className: "SCComponent-submitBtn-75",
      color: "primary",
      variant: "contained",
      fullWidth: true
    }))
```

This used to return the whole object, but now in the latest version it only returns **className** and **color**, **variant** and **fullWidth** is missing.

You can see the example below, updated both the uniforms and uniforms-material package to the rc.5 and it causes this issue.

https://codesandbox.io/s/currying-snow-jk55s?file=/App.tsx

Any ideas?

radekmie:
This change partially reverts #803 and fixes #815. This whole problem araised due to the way how different `wrapField` works across themes: they are either _external_ (only a few props are used; `uniforms-antd`, `uniforms-material`) or _internal_ (all props are used; `uniforms-bootstrap3`, `uniforms-bootstrap4`) (it's just an ad-hoc naming).

For reviewers: most important part is the removal of `filterDOMProps.register` in `packages/uniforms-material/src/wrapField.tsx`.

### DIFF
--- a/packages/uniforms-material/src/DateField.tsx
+++ b/packages/uniforms-material/src/DateField.tsx
@@ -2,8 +2,6 @@ import TextField, { StandardTextFieldProps } from '@material-ui/core/TextField';
 import React from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
-import wrapField from './wrapField';
-
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
 const dateFormat = (value?: Date) => value && value.toISOString().slice(0, -8);
 const dateParse = (timestamp: number, onChange: DateFieldProps['onChange']) => {
@@ -25,7 +23,6 @@ function Date({
   disabled,
   error,
   errorMessage,
-  fullWidth = true,
   helperText,
   InputLabelProps,
   inputRef,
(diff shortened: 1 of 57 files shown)


packages/uniforms-material/src/DateField.tsx
```diff
@@ -40,11 +39,11 @@ function Date({
     <TextField
       disabled={disabled}
       error={!!error}
-      fullWidth={fullWidth}
+      fullWidth
       helperText={(error && showInlineError && errorMessage) || helperText}
       label={label}
       InputLabelProps={{ ...labelProps, ...InputLabelProps }}
-      margin={props.margin ?? 'dense'}
+      margin="dense"
```
Monteth, line 46: I guess we usually use jsx string props in curly brackets like `{"dense"}`

Monteth (approved):

packages/uniforms-material/src/DateField.tsx
radekmie: We’ve agreed it’s correct and created [a card](https://github.com/orgs/vazco/projects/1#card-50180301) to configure it in our ESLint config.

PR #834, opened 2020-12-03
Adds row support to uniforms-material RadioField

asibilia:
Since `RadioField` is using Material's `RadioGroup` this change would allow users to specify the `row` property for the group (changing the orientation from vertical to horizontal).

From Material's [RadioGroup API](https://material-ui.com/components/radio-buttons/):
> To lay out the buttons horizontally, set the row prop: <RadioGroup row />.

### DIFF
--- a/packages/uniforms-material/src/RadioField.tsx
+++ b/packages/uniforms-material/src/RadioField.tsx
@@ -1,7 +1,7 @@
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormLabel from '@material-ui/core/FormLabel';
 import RadioMaterial, { RadioProps } from '@material-ui/core/Radio';
-import RadioGroup from '@material-ui/core/RadioGroup';
+import RadioGroup, { RadioGroupProps } from '@material-ui/core/RadioGroup';
 import React, { ReactNode } from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
@@ -9,7 +9,7 @@ import wrapField from './wrapField';
 
 export type RadioFieldProps = FieldProps<
   string,
-  RadioProps,
+  RadioProps & RadioGroupProps,
   {
     allowedValues?: string[];
     checkboxes?: boolean;
@@ -31,6 +31,7 @@ function Radio({
   margin = 'dense',
   name,
   onChange,
+  row,
   transform,
   value,
   ...props
@@ -48,6 +49,7 @@ function Radio({
       onChange={(event: any) => disabled || onChange(event.target.value)}
       ref={inputRef}
       value={value ?? ''}
+      row={row}
     >
       {allowedValues?.map(item => (
         <FormControlLabel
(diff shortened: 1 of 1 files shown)


packages/uniforms-material/src/RadioField.tsx
```diff
@@ -48,6 +49,7 @@ function Radio({
       onChange={(event: any) => disabled || onChange(event.target.value)}
       ref={inputRef}
       value={value ?? ''}
+      row={row}
```
radekmie, line 52: Let's keep these ordered:
```suggestion
      row={row}
      value={value ?? ''}
```

packages/uniforms-material/src/RadioField.tsx
```diff
@@ -1,15 +1,15 @@
...
+import RadioGroup, { RadioGroupProps } from '@material-ui/core/RadioGroup';
 import React, { ReactNode } from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
 import wrapField from './wrapField';
 
 export type RadioFieldProps = FieldProps<
   string,
-  RadioProps,
+  RadioProps & RadioGroupProps,
```
radekmie, line 12: As I can see in [`RadioGroup` API](https://material-ui.com/api/radio-group/), the `row` property is actually coming from the [`FormGroup`](https://material-ui.com/api/form-group/). I think it'd be better to add `row?: boolean` below rather than all `RadioGroupProps`, in order to limit incorrect props, e.g. `defaultValue` (they are not going to be used anyway as `...props` is being passed to `<Radio />`, not `<RadioGroup />`).

radekmie (changes requested):
Thanks, @asibilia! I've added two minor requests.

packages/uniforms-material/src/RadioField.tsx
```diff
@@ -2,7 +2,7 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormLabel from '@material-ui/core/FormLabel';
 import RadioMaterial, { RadioProps } from '@material-ui/core/Radio';
 import RadioGroup from '@material-ui/core/RadioGroup';
-import React, { ReactNode } from 'react';
```
asibilia, line 5: @radekmie this was unused

packages/uniforms-material/src/RadioField.tsx
asibilia: out of curiosity why not support both `RadioProps` and `RadioGroupProps` interfaces as their own props like: 

```ts
export type RadioFieldProps = FieldProps<
  string,
  {
    allowedValues?: string[];
    checkboxes?: boolean;
    fullWidth?: boolean;
    helperText?: string;
    margin?: any;
    RadioProps: Pick<RadioProps, 'allowed' | 'props' | 'here'>; // HERE
    RadioGroupProps: Pick<RadioGroupProps, 'allowed' | 'props' | 'here'>; // HERE
    row?: boolean;
    transform?(value: string): string;
  }
>;
```

Then pass those on to the actual components. `material-ui` seems to follow a similar pattern (without the `Pick`) for `InputLabelProps` and `InputProps` on certain components.

radekmie: We can do that, but I don't think there's a need for that here - existing props cover all usable `RadioGroupProps` (with this PR merged). Also, we try to keep fields consistent across themes. If you think it'd make sense, do file your ideas in an issue to have this PR small and focused.

radekmie (approved):

PR #835, opened 2020-12-07
Added readonly attribute.

### LINKED ISSUE #674: html input readonly attribute
Hi, the `readonly` attribute of an html input doesn't seem to be supported. I want to use it in the case where one of the fields is calculated from others, so that the user sees and submits the value but can't modify it via the UI.

The `disabled` attribute doesn't fit my case as I want it to be styled appropriately via bootstrap themes and submitted accordingly.

edit: I'm using SimpleSchema and Bootstrap 3.

martachrzanowska:
This change provides support for the `readonly` attribute.

Closes #674.


packages/uniforms-antd/src/ListDelField.tsx
```diff
@@ -38,9 +39,13 @@ function ListDel({
       disabled={!limitNotReached}
       icon={icon}
       onClick={() => {
-        const value = parent.value!.slice();
-        value.splice(nameIndex, 1);
-        parent.onChange(value);
+        if (readOnly) {
+          undefined;
+        } else {
```
radekmie, line 44:
```suggestion
        if (!readOnly) {
```

packages/uniforms-antd/src/ListAddField.tsx
```diff
@@ -44,9 +45,11 @@ function ListAdd({
       {...filterDOMProps(props)}
       disabled={!limitNotReached}
       icon={icon}
-      onClick={() => {
-        parent.onChange(parent.value!.concat([cloneDeep(value)]));
-      }}
+      onClick={() =>
+        readOnly
+          ? undefined
+          : parent.onChange(parent.value!.concat([cloneDeep(value)]))
```
radekmie, line 51:
```suggestion
        if (!readOnly) {
          parent.onChange(parent.value!.concat([cloneDeep(value)]));
        }
```

packages/uniforms-antd/src/DateField.tsx
```diff
@@ -25,7 +25,9 @@ function Date({
       disabled={props.disabled}
       name={props.name}
       onChange={value => {
-        props.onChange(value ? value.toDate() : undefined);
+        props.readOnly
+          ? undefined
+          : props.onChange(value ? value.toDate() : undefined);
```
radekmie, line 30:
```suggestion
        if (!props.readOnly)
          props.onChange(value ? value.toDate() : undefined);
        }
```

packages/uniforms-unstyled/src/BoolField.tsx
```diff
@@ -24,13 +25,7 @@ function Bool({
...
         id={id}
         name={name}
-        onChange={
-          disabled
-            ? undefined
-            : () => {
-                onChange(!value);
-              }
-        }
+        onChange={() => (disabled || readOnly ? undefined : onChange(!value))}
```
radekmie, line 28:
```suggestion
        onChange={() => (!disabled && !readOnly && onChange(!value))}
```

radekmie (changes requested):
I've added a few examples of how we could simplify this code quite a lot. Everything else is fine :rocket:

radekmie (approved):

radekmie (dismissed):

Monteth (dismissed):

radekmie (approved):

PR #837, opened 2020-12-15
Added migration guide (closes #808).

### LINKED ISSUE #808: Prepare migration guide from v2 to v3
As v3 release is getting closer every day, before an actual launch we'll need a written migration guide. The best start would be to collect all changelog entries.

radekmie:
I've added all the notes from #808 to the website. Right now it's under the _Getting Started_ section and I'm not 100% sure if it's good. Also, I've enabled `hideableSidebar` to make the migration guide more readable.

### DIFF
--- a/docs/migrating-2-to-3.md
+++ b/docs/migrating-2-to-3.md
@@ -1,0 +1,102 @@
+---
+id: migrating-2-to-3
+title: Migrating v2 to v3
+---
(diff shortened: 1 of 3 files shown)


Monteth (approved):
This doc is a detailed summary of why the uniforms v3 is just great.

PR #842, opened 2021-01-05
Cleaned up ESLint rules.

radekmie:
This change reenables `@typescript-eslint/no-unused-vars`, makes a step towards `@typescript-eslint/ban-types` (I still consider `object` fine in most of our code), and gets rid of some other ESLint related warnings.

### DIFF
--- a/packages/uniforms-bridge-simple-schema-2/src/SimpleSchema2Bridge.ts
+++ b/packages/uniforms-bridge-simple-schema-2/src/SimpleSchema2Bridge.ts
@@ -93,9 +93,9 @@ export default class SimpleSchema2Bridge extends Bridge {
   // eslint-disable-next-line complexity
   getProps(name: string, props: Record<string, any> = {}) {
     // eslint-disable-next-line prefer-const
-    let { optional, type, uniforms, ...field } = this.getField(name);
+    const { optional, type, uniforms, ...contextField } = this.getField(name);
 
-    field = { ...field, required: !optional };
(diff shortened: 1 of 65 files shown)


packages/uniforms-bridge-simple-schema-2/src/SimpleSchema2Bridge.ts
```diff
@@ -92,8 +92,7 @@ export default class SimpleSchema2Bridge extends Bridge {
 
   // eslint-disable-next-line complexity
   getProps(name: string, props: Record<string, any> = {}) {
-    // Type should be omitted.
-    // eslint-disable-next-line no-unused-vars, prefer-const
+    // eslint-disable-next-line prefer-const
     let { optional, type, uniforms, ...field } = this.getField(name);
 
     field = { ...field, required: !optional };
```
Monteth, line 98: We can easily avoid using `let` by naming `...field` differently
```suggestion
    const { optional, type, uniforms, ...contextField } = this.getField(name);

    const field = { ...contextField, required: !optional };
```
I know it's barely cosmetical change here, but I believe the `prefer-const` rule is there for a reason.

Monteth (approved):

Monteth (approved):

Monteth (changes requested):
As we see this `let` also was there `for a reason` :D
So we have two options, one is to change the logic of this function, the other to just revert this change and live as nothing happened.

Monteth (approved):

PR #851, opened 2021-01-13
Fixed documentation of "onValidate"

### LINKED ISSUE #847: Incosistent documentation of "onValidate" prop
As per the [migration guide](https://uniforms.tools/docs/migrating-2-to-3#validation-flow-changes) `onValidate` should return a promise, however the examples still use callbacks:

* https://uniforms.tools/docs/api-forms/#props-usage-1
* https://uniforms.tools/docs/api-forms/#asynchronous-validation
* maybe more places, need to check the code

Thanks!

yurtsiv:
Closes #847

### DIFF
--- a/docs/api-forms.md
+++ b/docs/api-forms.md
@@ -81,8 +81,8 @@ import { ValidatedForm } from 'uniforms'; // Or from the theme package.
     }
 
     // ...or any additional validation if an error is already there...
-    if (error) {
-      return;
+    if (isSomeSpecialCase(error)) {
+      return MyAPI.checkOtherCondition(model);
     }
 
     // ...or feed it with another error.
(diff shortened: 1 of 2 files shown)


docs/api-forms.md
```diff
@@ -74,25 +74,25 @@ however, there will be validation checks.
...
     // You can either ignore validation error...
     if (omitValidation(model)) {
-      return callback(null);
+      return null;
     }
 
     // ...or any additional validation if an error is already there...
     if (error) {
-      return callback();
+      return;
```
yurtsiv, line 85: Frankly, I don't understand the comment above and what this example is about. Can you explain it, so maybe I can  rephrase the comment somehow?

radekmie: The point here is that sometimes you'd like to transform the error. For example, you may want to hide an error _X is required_ if the user is offline.

radekmie (approved):

PR #865, opened 2021-02-03
Added error styling for ListField in uniforms-antd (closes #844).

### LINKED ISSUE #844: antd list error not styled
Hi

Thanks for this awesome package - I really like it and use it every day :-)

I've just figured out, that an error in a list is not styled when using antd. 

Steps to reproduce:

1. go to the playground, choose antd.
2. copy paste the following code for the schema and test.

```
new SimpleSchema2Bridge(
  new SimpleSchema({
    date: { type: Date, defaultValue: new Date() },
    adult: Boolean,
    size: {
      type: String,
      defaultValue: 'm',
      allowedValues: ['xs', 's', 'm', 'l', 'xl'],
    },
    rating: {
      type: Number,
      allowedValues: [1, 2, 3, 4, 5],
      uniforms: { checkboxes: true },
    },
    friends: { type: Array, minCount: 0, optional: true, custom() {
        if (true) {
          if (this.isSet) {
            return undefined;
          }
          return SimpleSchema.ErrorTypes.REQUIRED;
        }
        return undefined;
      }, },
    'friends.$': Object,
    'friends.$.name': { type: String, min: 3 },
    'friends.$.age': { type: Number, min: 0, max: 150 },
  })
)
```

Maybe I am doing something wrong, but I think it should be styled.

Thanks so much.
Regards,
Oliver

Monteth:
This PR introduces ListField error stylings in uniforms-antd. 
Based on https://github.com/vazco/uniforms/issues/844.

image
image

No-error styling for comparison:
image
image

### DIFF
--- a/packages/uniforms-antd/src/ListField.tsx
+++ b/packages/uniforms-antd/src/ListField.tsx
@@ -1,5 +1,4 @@
 import QuestionCircleOutlined from '@ant-design/icons/QuestionCircleOutlined';
-import Form from 'antd/lib/form';
 import Tooltip from 'antd/lib/tooltip';
 import classNames from 'classnames';
 import React, {
@@ -49,14 +48,14 @@ function List({
   wrapperCol,
   ...props
 }: ListFieldProps) {
-  const borderStyle = error
+  const wrapperStyle = error
     ? { borderColor: 'rgb(255, 85, 0)', ...style }
     : style;
 
   return (
     <div
       {...filterDOMProps(props)}
-      style={borderStyle}
+      style={wrapperStyle}
       className={classNames([className, 'ant-list', 'ant-list-bordered'])}
     >
       {!!label && (
@@ -72,6 +71,7 @@ function List({
           )}
         </div>
       )}
+
       {!!(error && showInlineError) && <div>{errorMessage}</div>}
 
       {value?.map((item, itemIndex) =>
--- a/packages/uniforms-antd/src/ListField.tsx
+++ b/packages/uniforms-antd/src/ListField.tsx
@@ -32,6 +32,8 @@ const defaultStyle = {
   padding: '10px',
 };
 
+const errorStyle = { borderColor: 'rgb(255, 85, 0)' };
+
 function List({
   children = <ListItemField name="$" />,
   className,
@@ -49,7 +51,9 @@ function List({
   ...props
 }: ListFieldProps) {
   const wrapperStyle = error
-    ? { borderColor: 'rgb(255, 85, 0)', ...style }
+    ? style
+      ? { ...errorStyle, ...style }
+      : errorStyle
     : style;
 
   return (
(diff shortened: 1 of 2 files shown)


packages/uniforms-antd/src/ListField.tsx
```diff
@@ -48,43 +49,57 @@ function List({
...
+          </Tooltip>
+        </span>
+      )}
+    </div>
+  );
+
+  const borderStyle = {
+    ...(error ? { borderColor: 'rgb(255, 85, 0)' } : {}),
+    ...style,
+  };
```
radekmie, line 69: The presence of both `error` and `style` is rather rare, maybe we could optimize for this case? Something like this:
```suggestion
  const borderStyle = 
    error
      ? { borderColor: 'rgb(255, 85, 0)', ...style }
      : style;
```

This would create a new object only when it's needed.

packages/uniforms-antd/src/ListField.tsx
```diff
@@ -48,43 +49,57 @@ function List({
...
-          )}
-        </div>
-      )}
-
-      {!!(error && showInlineError) && <div>{errorMessage}</div>}
+      <Form.Item
+        label={labelNode}
+        hasFeedback
+        style={{ marginBottom: 0 }}
+        validateStatus={error ? 'error' : ''}
```
radekmie, line 81: I guess we could pass the original `validateStatus` from props here, just like in `wrapField`.

packages/uniforms-antd/src/ListField.tsx
```diff
@@ -48,43 +49,57 @@ function List({
...
-            </span>
-          )}
-        </div>
-      )}
-
-      {!!(error && showInlineError) && <div>{errorMessage}</div>}
+      <Form.Item
+        label={labelNode}
+        hasFeedback
+        style={{ marginBottom: 0 }}
```
radekmie, line 80: This should be a constant, so it wouldn't change. Same goes for the `style` below.

radekmie (commented):
While I quite like the way it looks, I hate the idea of adding even more custom styles. Maybe it's just me, but maybe the red outline is enough? This red x icon is not usable with long array s anyway (it's always centered). What do you think?

Monteth:
> While I quite like the way it looks, I hate the idea of adding even more custom styles. Maybe it's just me, but maybe the red outline is enough? This red x icon is not usable with long array s anyway (it's always centered). What do you think?

I think the `x` icon is more consistent with the other fields and this was a purpose for adding it here, but I like the idea of, I would say 'minimal' style, so I will change it.

packages/uniforms-antd/src/ListField.tsx
```diff
@@ -48,13 +48,17 @@ function List({
   wrapperCol,
   ...props
 }: ListFieldProps) {
+  const wrapperStyle = error
+    ? { borderColor: 'rgb(255, 85, 0)', ...style }
```
Monteth, line 52: Is it worth setting this error style to const and spread it here or just leave it, because its usage is rare?

radekmie: Yes, great idea. Something like this:
```ts
const wrapperStyle = error
  ? style ? { ...errorStyle, ...style } : errorStyle
  : style;
```

radekmie (approved):

radekmie (approved):

PR #866, opened 2021-02-05
Updated eslint versions, removed peer dependencies from package.json.

wadamek65:
This PR updates the `eslint-config-vazco` which in turn allows us to remove all of it's peer dependencies from `package.json`. All of the errors have been silenced by turning of the rules and an issue to fix this has been reported in #867 .

Due to the requirement of `npm@7` for package installations and `actions/setup-node` not allowing for npm version customization, we have to install it manually in the installation step.


package.json
```diff
@@ -31,25 +31,16 @@
...
-    "babel-eslint": "10.1.0",
     "classnames": "^2.0.0",
     "core-js": "3.8.3",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.5",
-    "eslint": "6.8.0",
-    "eslint-config-prettier": "6.11.0",
-    "eslint-config-vazco": "5.3.1",
+    "eslint": "7.19.0",
+    "eslint-config-vazco": "6.1.0",
```
radekmie, line 41: Only the latter is needed, right?

.github/workflows/CI.yml
```diff
@@ -17,6 +17,10 @@ jobs:
         uses: actions/setup-node@v2.1.2
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Debug version.
+        run: npm --version
+      - name: Update npm to version 7.
+        run: npm install -g npm@latest
```
radekmie, line 23: Yes, it's needed. However, please do move it to `npm ci` as `npm i -g npm@7 && npm ci`.

packages/uniforms-bootstrap3/src/ListAddField.tsx
```diff
@@ -36,6 +36,8 @@ function ListAdd({
     !disabled && !(parent.maxCount! <= parent.value!.length);
 
   return (
+    // TODO[jsx-a11y]
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
```
radekmie, line 40: I think we could disable them globally in `.eslintrc` for now - this makes too much noise.

radekmie (approved):

Monteth (approved):

PR #870, opened 2021-02-10
Configured sideEffects (closes #843).

### LINKED ISSUE #843: tree shaking and  "sideEffects": false ?
i use uniforms on https://github.com/react-page/react-page to generate forms and i wanted to re-export the bundled uniforms version (using something like `export * from 'uniforms-material'`).

i noticed that this will bundle everything in, even if you don't use any of the re-exports.

I think this is because uniform packages do not declare ` "sideEffects": false`, see https://webpack.js.org/guides/tree-shaking/

i think its quite safe to set that, altough there might actually be some sideeffects (i remember that uniforms extended simpl-schema automatically in some versions, that would be a side effect. not sure if its still the case).

radekmie:
This change configures [`sideEffects`](https://webpack.js.org/guides/tree-shaking/) in all packaged in order to let bundlers do better at tree shaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Next
+
+- **Added:** The [`sideEffects`](https://webpack.js.org/guides/tree-shaking/) flag to let bundlers do better at tree shaking. [\#843](https://github.com/vazco/uniforms/issues/843)
+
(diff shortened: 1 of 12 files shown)


Monteth (approved):

Monteth:
Is there any way to test this new configuration?

radekmie:
@Monteth I've tried to check it locally but couldn't verify it. I guess we'll _test on production_ :shrug:

PR #887, opened 2021-02-26
Improved types

radekmie:
This PR is composed of three parts:
* b0101cf Switched from bound to unbound methods. It allowed re-enabling of `@typescript-eslint/unbound-method`.
* 12e88a1 Improved `onChange` type. Switch from `Value | null` into `any` was needed, to make it possible to use `onChange(otherFieldValue, 'otherField')`. Additionally, the `value` argument is not optional anymore. Ideally, it should be something like this:
    ```ts
    type OnChange<Value> = {
      (value: Value): void;
      (value: any, name: string): void;
    };
    ```
    However, this breaks a lot of inferred types.
* 94d48fc Replaced `ts-ignore` with `ts-expect-error`. These are strictly better, letting us know that it's no longer needed.

**EDIT:**
* 63e4bac I managed to make the `onChange` typed as planned.
* 57c0242 I had to revert 94d48fc for Meteor modules as it broke with our internal declarations. It'll require further investigation.

### DIFF
--- a/packages/uniforms/src/types.ts
+++ b/packages/uniforms/src/types.ts
@@ -12,9 +12,9 @@ export type Context<Model> = {
   error: any;
   model: DeepPartial<Model>;
   name: string[];
-  onChange(key: string, value?: any): void;
-  onSubmit(event?: SyntheticEvent): any | Promise<any>;
-  randomId(): string;
+  onChange: (key: string, value?: any) => void;
+  onSubmit: (event?: SyntheticEvent) => any | Promise<any>;
+  randomId: () => string;
   schema: Bridge;
   state: {
     disabled: boolean;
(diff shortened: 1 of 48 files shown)


Monteth (approved):

packages/uniforms/src/types.ts
```diff
@@ -50,13 +50,18 @@ export type GuaranteedProps<Value> = {
   id: string;
   label: ReactNode;
   name: string;
-  onChange(value?: Value | null, name?: string): void;
+  onChange: OnChange<Value | undefined>;
```
wadamek65, line 53: This change doesn't take into consideration the previous `null` typing. Is this intended? (I see the `null`s were changed into `undefined`s but just wanted to make sure.

wadamek65 (approved):

packages/uniforms/src/types.ts
radekmie: Yes, it was intentional. I wrote `null` here as I prefer them over `undefined`, but none of the `onChange` handlers have had special `null` handling.

PR #895, opened 2021-03-08
Re-enabled jsx-a11y rules and fixed most linter errors.

wadamek65:
In this PR:
- Website/docs adjustments
- Components' adjustments for multiple themes
- Changed selectors in some tests to more semantic clear ones
- Updated `eslint-config-vazco` to version 6.1.1
- Re-enabled all a11y rules back


packages/uniforms-bootstrap3/src/ListAddField.tsx
```diff
@@ -35,13 +35,22 @@ function ListAdd({
...
+  const onClickHandler = () => {
+    if (limitNotReached && !readOnly) {
+      parent.onChange(parent.value!.concat([cloneDeep(value)]));
+    }
+  };
+
   return (
     <div
       {...filterDOMProps(props)}
+      role="button"
```
wadamek65, line 47: @radekmie Please take a look at this file. There are multiple errors were there are `div` elements without semantic meaning with `onClick` handlers which results in multiple different linter errors. There are 2 approaches I could take:
1. Change the `div` to a `button` and fix the styles (I tried it and it's unfortunately not as easy as I would like).
2. Add a proper aria role and keyboard handlers to the element.

This is an example of the 2nd point which I think is elegant enough and we should go with it. There is one thing to keep in mind though: `KeyboardEvent.code` [does not work in IE at all](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#browser_compatibility). I could probably work around it and add more logic but my first question would be - do we need it?

Also another question as the `onKeyDown` handler would basically be the same in all the files. Do you think it would make sense to export a helper function for it and if yes then what would be an appropriate place for it? Something like this:
```ts
const onKeyDownCreator = (onClickHandler: () => void) => (event: KeyboardEvent) => {
  if (event.code === 'Enter') {
    onClickHandler();
  }
}
```

website/pages-parts/LandingPage/Footer.tsx
```diff
@@ -58,7 +58,11 @@ export function Footer() {
               </p>
             </Link>
             <a href="https://vazco.eu">
-              <img className={styles.vazco} src="assets/vazco.svg" />
+              <img
+                alt="Vazco Home"
```
radekmie, line 62: How about a little more explicit _homepage_?

packages/uniforms-bootstrap3/src/ListAddField.tsx
radekmie: I'd definitely go with 2. As we've already discussed, it's not that easy to get rid of all button-like styles in a cross-browser way.

While this code would be the same for multiple packages, I don't think there's a need to share it - we already do that and not all of the packages need that (see `DateField` for an example and the AntD theme as a counterexample).

I also believe we could have one handler for both cases. something like this:
```tsx
function onAction(event: KeyboardEvent | MouseEvent) {
  if (
    limitNotReached &&
    !readOnly &&
    (!(event instanceof KeyboardEvent) || event.code === 'Enter')
  ) {
    parent.onChange(parent.value!.concat([cloneDeep(value)]));
  }
}
```

When it comes to cross-browser compatibility, I believe we could use [`key`, as it's part of the React `SyntheticEvent` API](https://reactjs.org/docs/events.html#keyboard-events).

wadamek65: A few remarks:

The conditional has to look like this:
```ts
(!(event.nativeEvent instanceof KeyboardEvent) || event.key === 'Enter')
```
Which would require me to write a type guard which is another function which might not look nice.

I could also switch this condition to:
```ts
(!('key' in event) || event.key === 'Enter')
```

Which has an additional advantage of not breaking all of our examples in playground/docs. `instanceof KeyboardEvent` returns `false` in all iframe's due to different reference of the prototype chain which would require some more hackery to make this work. 

What do you think? @radekmie

radekmie: Why `event.nativeEvent`? Have you tried importing both `KeyboardEvent` and `MouseEvent` from `react`? It won't cause problems with the iFrame. If not, we could use the one with `'key' in event`.

wadamek65: @radekmie Which `KeyboardEvent` and `MouseEvent` do you have in mind? AFAIK there is only a `SyntheticEvent` class which is the same for both mouse and keyboard handlers and the two you mentioned are only typescript types. Or am I mistaken?

radekmie: There's [this](https://github.com/facebook/react/blob/9198a5cec0936a21a5ba194a22fcbac03eba5d1d/packages/react-dom/src/events/SyntheticEvent.js#L472-L474), but I'm not sure. Just try and let me know.

wadamek65: I don't think this is even exported from React, couldn't even find it in types and it is `undefined` when accessing `React.SyntheticKeyboardEvent`. Also the event's prototype seems to be just `SyntheticEvent` so I don't think this works.
image

wadamek65: Changed all checks to `'key' in event` 🦖

packages/uniforms-bootstrap3/__tests__/ListDelField.tsx
```diff
@@ -46,3 +46,11 @@ test('<ListDelField> - correctly reacts on click', () => {
   expect(wrapper.find('i').simulate('click')).toBeTruthy();
   expect(onChange).toHaveBeenLastCalledWith('x', ['x', 'z']);
 });
+
+test('<ListDelField> - correctly reacts on keyboard enter key', () => {
+  const element = <ListDelField name="x.1" />;
+  const wrapper = mount(element, context());
+
+  wrapper.find('[role="button"]').props().onKeyDown?.({ key: 'Enter' } as any);
+  expect(onChange).toHaveBeenLastCalledWith('x', ['x', 'z']);
```
wadamek65, line 55: Before I go ahead and add tests for all the files I wanted to check a few things with you.

1. `simulate` is deprecated and not recommended by the author. It will also be removed in `enzyme@4` if it ever gets released. More info [here](https://github.com/enzymejs/enzyme/issues/2173). This is why I used `props().onKeyDown` directly.
2. The `onKeyDown` has a very weird type for it's argument which expects a full (and hacked) version of `SyntheticEvent` with all the keys but works just fine with a simple object. This is why I used casting to `any` here.
3. Now that the component is more semantically verbose I also changed the selector to be a little more descriptive and less error-prone to changes. I could also change the other tests to use it as it is pretty much no effort change.

@radekmie @Monteth

radekmie: 1. I'd rather stick to what we do use now and then migrate all tests at once.
2. OK; shouldn't be the case with 1.
3. Sure, go ahead!

package-lock.json
```diff
@@ -3897,6 +3898,9 @@
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
       "integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
+      "dependencies": {
+        "graceful-fs": "^4.1.6"
+      },
       "optionalDependencies": {
         "graceful-fs": "^4.1.6"
       }
```
radekmie, line 3906: It seems that your installation added those whereas mine didn't. Do you have any idea why it is so?

packages/uniforms-bootstrap3/src/ListAddField.tsx
```diff
@@ -35,15 +35,28 @@ function ListAdd({
   const limitNotReached =
     !disabled && !(parent.maxCount! <= parent.value!.length);
 
+  function onAction(
+    event:
+      | React.KeyboardEvent<HTMLDivElement>
+      | React.MouseEvent<HTMLDivElement, MouseEvent>,
+  ) {
```
radekmie, line 42: As we are not using the `currentTarget` property, we can simplify it:
```suggestion
  function onAction(event: React.KeyboardEvent | React.MouseEvent) {
```

radekmie (changes requested):

package-lock.json
wadamek65: I'm not sure - `npm ls graceful-fs` shows that a lot of packages depend on `graceful-fs@4.2.4`. Maybe the lock generation changed across npm versions? Which version do you use? Mine is 7.4.0

wadamek65: I've tried resetting `package-lock.json` to the version currently on master and regenerating it but it produced the same result. 😕

packages/uniforms-bootstrap3/src/ListAddField.tsx
wadamek65: Changed

package-lock.json
radekmie: Mine is v7.7.6. I've done that once again, and yes, it removed it - I guess it's an OS thing. Also, a newer version of npm removed all `resolved` and `integrity` fields - weird. Let's leave it for now.

radekmie (approved):

Monteth (approved):

Issue #889, opened 2021-03-02
v3: no longer possible to know whether a form has been submitted


macrozone:
we validate form fields on the fly, that means we show an error when a field has been filled and somethign is wrong

we achieved this with using `validate="onChange"`, but only show an error if the field was either touched or changed.
We also showed all errors when the user tried to submit the form.

we set touched=true when either the field was blurred or when the form was "submitting". 

Problem is now that submitting is never `true`, so that does not work anymore (we also needed to `useForm().submitting`)

is there a way to find out if the form has ever been submitted?

macrozone:
ok i found a fix:

on the ValidatedForm there is state.validate which is true when the form has been submitted but failed the validation.

While the name is maybe a bit missleading, this is exactly what i need

i prepared a small PR https://github.com/vazco/uniforms/pull/890

radekmie:
Hi @macrozone. I'll comment on both your request and #890 here. There is no way to know whether the form was submitted _in general_. I see that you found the `validate` flag of the `ValidatedForm`. Indeed, it does exactly that, but it's there just to make `onChangeAfterSubmit` work.  I think it makes sense to move this functionality up to the `BaseForm` and make it part of the public API. Would you like to work on it?

macrozone:
> Hi @macrozone. I'll comment on both your request and #890 here. There is no way to know whether the form was submitted in general. I see that you found the validate flag of the ValidatedForm. Indeed, it does exactly that, but it's there just to make onChangeAfterSubmit work. I think it makes sense to move this functionality up to the BaseForm and make it part of the public API. Would you like to work on it?

What do you suggest as property? It's meaning i "form has been tried to be submitted, but validation failed or validation is not yet done". Something like a "submitCount" could maybe also be used for that, paired with "submitSuccessCount" to highlight that "submitCount" also counts unsuccessful submits (where validation failed)

> EDIT: Also, the title says no longer - I'm pretty sure this information was never part of the API - could you explain what you meant?

previously, you could know whether the form was submitted, because the fields would receive `submitting: true` for a short time (before validation). So you could update a state on the fields that this form had been tried to submit. I understand that this was probably not a clean solution to begin with, but it worked without problems.

radekmie:
[...]

PR #890, opened 2021-03-02
wip: validate

macrozone:

PR #906, opened 2021-03-17
feat: submitted state in context 

macrozone:
see https://github.com/vazco/uniforms/issues/889

fixes https://github.com/vazco/uniforms/issues/889

### DIFF
--- a/packages/uniforms-antd/__tests__/_createContext.ts
+++ b/packages/uniforms-antd/__tests__/_createContext.ts
@@ -18,6 +18,7 @@ export default function createContext(
       onChange() {},
       onSubmit() {},
       randomId,
+      submitted: false,
       submitting: false,
       validating: false,
       ...context,
--- a/packages/uniforms-bootstrap3/__tests__/_createContext.ts
+++ b/packages/uniforms-bootstrap3/__tests__/_createContext.ts
@@ -18,6 +18,7 @@ export default function createContext(
       onChange() {},
       onSubmit() {},
       randomId,
+      submitted: false,
       submitting: false,
       validating: false,
       ...context,
--- a/packages/uniforms/__tests__/BaseForm.tsx
+++ b/packages/uniforms/__tests__/BaseForm.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { BaseForm, Bridge } from 'uniforms';
+import { BaseForm, Bridge, Context } from 'uniforms';
 
 import mount from './_mount';
 
@@ -35,7 +35,7 @@ describe('BaseForm', () => {
     );
 
     const context = wrapper.instance().getContext();
-    expect(context).toEqual({
+    expect(context).toEqual<Context<any>>({
       changed: false,
       changedMap: {},
       error,
@@ -52,6 +52,7 @@ describe('BaseForm', () => {
         readOnly: false,
         showInlineError: false,
       },
+      submitted: false,
       submitting: false,
       validating: false,
     });
@@ -91,7 +92,7 @@ describe('BaseForm', () => {
 
     it('have correct `state`', () => {
       const context = wrapper.instance().getContext();
-      expect(context.state).toEqual({
+      expect(context.state).toEqual<Context<any>['state']>({
         disabled: true,
         label: false,
         placeholder: true,
@@ -141,8 +142,9 @@ describe('BaseForm', () => {
       wrapper.setProps({ autosave: true });
       wrapper.instance().getContext().onChange('a', 1);
       await new Promise(resolve => setTimeout(resolve));
-
+      const context = wrapper.instance().getContext();
       expect(onSubmit).toHaveBeenCalledTimes(1);
+      expect(context.submitted).toBe(true);
       expect(onSubmit).toHaveBeenLastCalledWith(model);
     });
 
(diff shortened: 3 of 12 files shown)


packages/uniforms-bootstrap3/__tests__/_createContext.ts
```diff
@@ -28,8 +29,8 @@ export default function createContext(
         placeholder: false,
         readOnly: false,
         showInlineError: false,
-        ...context?.state,
-      },
-    },
+        ...context?.state
+      }
+    }
```
macrozone, line 34: prettier config might have changed. it says `trailingComma: "none"`, so i guess the reformat is correct, but i am unsure

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -31,11 +31,11 @@ describe('BaseForm', () => {
 
   it('have correct context', () => {
     const wrapper = mount<BaseForm<any>>(
-      <BaseForm error={error} model={model} schema={schema} />,
+      <BaseForm error={error} model={model} schema={schema} />
     );
 
     const context = wrapper.instance().getContext();
-    expect(context).toEqual({
+    expect(context).toEqual<Context<any>>({
```
macrozone, line 38: this way, the expected object is type checked. This helps to not forget properties

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -216,37 +218,54 @@ describe('BaseForm', () => {
...
       expect(wrapper.state('resetCount')).toBe(1);
     });
+    it('sets submitted back to false', async () => {
+      const wrapper = createWrapper();
+      const instance = wrapper.instance();
+      expect(instance.getContext().submitted).toBe(false);
+      wrapper.find('form').simulate('submit');
+      expect(instance.getContext().submitted).toBe(true);
+      instance.reset();
+      expect(instance.getContext().submitted).toBe(false);
```
macrozone, line 239: i decided that reset will also reset this new property. I guess one would expect that

macrozone:
@radekmie linting does not work locally, it says that `sh: eslint: command not found`, which makes sense as eslint is not in the devdependencies.

wadamek65:
@macrozone `eslint` is not in the dev dependencies because it is a peer dependency of `eslint-config-vazco`. We are using `npm@7` which installs these for us automatically. Please make sure you are using `npm@7` when installing the dependencies and lint the code properly before submitting this for review.

macrozone:
> @macrozone `eslint` is not in the dev dependencies because it is a peer dependency of `eslint-config-vazco`. We are using `npm@7` which installs these for us automatically. Please make sure you are using `npm@7` when installing the dependencies and lint the code properly before submitting this for review.

i have done this now

packages/uniforms-antd/__tests__/_createContext.ts
```diff
@@ -19,6 +19,7 @@ export default function createContext(
       onSubmit() {},
       randomId,
       submitting: false,
+      submitted: false,
```
radekmie, line 22: Do keep these properties ordered.

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -216,33 +218,50 @@ describe('BaseForm', () => {
...
+        <BaseForm model={model} schema={schema} onSubmit={onSubmit} />,
+      );
 
     it('increase `resetCount`', () => {
+      const wrapper = createWrapper();
       wrapper.instance().reset();
 
       expect(wrapper.state('resetCount')).toBe(1);
     });
+    it('sets submitted back to false', async () => {
```
radekmie, line 232:
```suggestion
    });

    it('sets submitted back to false', async () => {
```

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -256,7 +275,15 @@ describe('BaseForm', () => {
...
     });
 
+    it('sets `submitted` to true', async () => {
+      const wrapper = createWrapper();
+      const instance = wrapper.instance();
+      expect(instance.getContext().submitted).toBe(false);
+      wrapper.find('form').simulate('submit');
+      expect(instance.getContext().submitted).toBe(true);
+    });
     it('sets `submitting` state while submitting', async () => {
```
radekmie, line 285:
```suggestion
    });

    it('sets `submitting` state while submitting', async () => {
```

packages/uniforms/src/BaseForm.tsx
```diff
@@ -251,6 +255,7 @@ export class BaseForm<
       event.preventDefault();
       event.stopPropagation();
     }
+    this.setState({ submitted: true });
```
radekmie, line 258: It'd be better to skip unneeded updates:
```suggestion

    this.setState(state => state.submitted ? null : { submitted: true });
```

packages/uniforms/src/ValidatedForm.tsx
```diff
@@ -119,7 +119,7 @@ export function Validated<Base extends typeof BaseForm>(Base: Base) {
         event.stopPropagation();
       }
 
-      this.setState({ validate: true });
+      this.setState({ validate: true, submitted: true });
```
radekmie, line 122:
```suggestion
      this.setState({ submitted: true, validate: true });
```

radekmie (changes requested):
I've added some comments, most of them are code style-related. Also, please do not create unnecessary PRs - you could've continued on #890.

packages/uniforms/src/BaseForm.tsx
macrozone: react already does that internally.

macrozone:
i personally do not like that level of nitpicking, you should fix your linter if you want to enforce ordering rules of properties. Its bikeshedding after all. But your repo, your rules i gues...

packages/uniforms/src/BaseForm.tsx
radekmie: No, it's not - here's a demo [CodeSandbox](https://codesandbox.io/s/practical-gareth-di8sb?file=/src/App.js). In class components you have to return `null`, in `useState` hook it works only if `newState === oldState`.

macrozone:
i pushed the requested changes. 

There are two problems with the repo by the way:

- prettierconfig is wrong and removes trailing commas, but eslint complains about missing trailing commas. had to run `npm run lint:code -- --fix` to correct this.
- precommit hook (lint-staged) hanged forever (unclear why), so i canceled it, but then all my untaged changes where gone. Happend to me 3 times now.

radekmie (approved):

macrozone:
hooray, thank you!

Monteth (approved):

PR #907, opened 2021-03-17
Respect initial count in not required list fields (fixes #875).

### LINKED ISSUE #875: Bootstrap 4 ListField / ListAddField Always Tries to Set Count to 1
Hi,

I'm using the `bootstrap-4` theme.  When I use the default ListItem component or a custom ListItem component, clicking 'Add' the first time always tries to set the count to 1.

For example, if the initialCount=0, clicking add **once** adds one ListItemField.  This works as expected.

However, if the initialCount=1, clicking add **once** does nothing, clicking add again adds a second ListItemField.

If the initialCount=2, clicking add **once** removes the second ListItemField so that there is only 1.  Clicking it again then adds a second ListItemField back.

This occurred using this code:

```
<ListField
    name="prices"
    addIcon={<Add />}
    removeIcon={<Delete />}
    initialCount={1}
    showInlineError
/>
```

It was also happening for my custom list field, but I'm assuming the fix will apply to both.

Monteth:
See [875](https://github.com/vazco/uniforms/issues/875) for more info.
- [x] fix
- [x] tests


packages/uniforms-antd/src/ListAddField.tsx
```diff
@@ -21,6 +21,7 @@ const defaultStyle = { width: '100%' };
 function ListAdd({
   disabled,
   icon = <PlusSquareOutlined />,
+  initialCount = 0,
```
radekmie, line 24: I think that we don't need this default, right? Bridges do have fallback anyway.

packages/uniforms-antd/src/ListAddField.tsx
```diff
@@ -31,11 +32,10 @@ function ListAdd({
 }: ListAddFieldProps) {
   const nameParts = joinName(null, name);
   const parentName = joinName(nameParts.slice(0, -1));
-  const parent = useField<{ maxCount?: number }, unknown[]>(
-    parentName,
-    {},
-    { absoluteName: true },
-  )[0];
+  const parent = useField<
+    { maxCount?: number; initialCount: number },
```
radekmie, line 36:
```suggestion
    { initialCount?: number; maxCount?: number },
```

radekmie (dismissed):
Code is fine. Let's add tests first though.

packages/uniforms-antd/__tests__/ListField.tsx
```diff
@@ -160,3 +160,25 @@ test('<ListField> - renders correct error style (with specified style prop)', ()
...
+    ),
+  );
+
+  expect(wrapper.find(ListAddField).simulate('click')).toBeTruthy();
+  expect(onChange).toHaveBeenNthCalledWith(1, 'x', [
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+  ]); // renders ( initialCount + 1 ) default values
```
radekmie, line 183: 1. I'd remove this comment.
2. Could you add a test with a different default value as well?

radekmie (changes requested):
In the meantime, the `uniforms-bootstrap5` got created - could you update it as well?

michael-vasyliv:
@radekmie do we need to fix this issue for bootstrap5 as well?

radekmie:
@michael-vasyliv Yes, I already mentioned that.

radekmie (approved):

PR #924, opened 2021-04-08
Fix warning 'act' warning in uniforms-antd (closes #877).

### LINKED ISSUE #877: Investigate act problem in uniforms tests.
Warning message in tests run: 
```
Warning: An update to List inside a test was not wrapped in act(...).
```
Full example here:
https://github.com/vazco/uniforms/runs/1727887630?check_suite_focus=true#step:6:15

Monteth:
(See #877 - some warning is reported in uniforms tests)
The warning is fired in two tests in SelectField in uniforms-antd, and both of them uses 'open' prop on SelectField. It seems like props like 'open' and 'defaultOpen' trigger state change after the first render, and it causes the warning and further potential problems.
As I don't see any benefits of using 'open'-like props - test results are perfectly the same, and they are not present in other themes, so I want to propose to not use them here.  
The other way is to trigger an 'open' state of the select after mount simulating proper event, but I don't know if it's really necessary.

### DIFF
--- a/packages/uniforms-antd/__tests__/SelectField.tsx
+++ b/packages/uniforms-antd/__tests__/SelectField.tsx
@@ -79,8 +79,7 @@ test('<SelectField> - renders a select with correct name', () => {
 });
 
 test('<SelectField> - renders a select with correct options', () => {
-  // @ts-expect-error Is open a valid prop?
-  const element = <SelectField name="x" open />;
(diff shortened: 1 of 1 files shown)


radekmie (approved):
It's all good. I think that we've had it here in the first place, as it worked differently in the previous AntD versions.

PR #927, opened 2021-04-09
Add 'showInlineError' property to AutoFields (closes #827).

### LINKED ISSUE #827: Allow passing showInlineError to AutoFields component
I would expect to be able to set `showInlineError` on the `AutoFields` component, and have it pass down to each `AutoField`, just like you can do this with `showInlineError` on the `AutoForm` component.

There might be other props like this that you can set on `AutoForm`, but not on `AutoFields`?

Monteth:
Added a possibility to pass 'showInlineError' to `AutoFields` in all themes except uniforms-unstyled, which was requested in #827.

I didn't come up with the idea of how to provide the possibility to pass all (or most) props to children `AutoField`. We just have to move them to a common property object and pass them to children, which precludes nice UX mentioned by @Floriferous because the interface differs from the `AutoField`. On the other hand, we have to provide an API to configure the `AutoFields` container component, so we have a override with the child `AutoField`, or we have to move this API to a common prop object - which is breaking change.

Therefore I introduce a minimal change in `AutoFields` allowing you to pass 'showInlineError' prop, and I want you to share your thoughts.

### DIFF
--- a/packages/uniforms-antd/src/AutoFields.tsx
+++ b/packages/uniforms-antd/src/AutoFields.tsx
@@ -8,6 +8,7 @@ export type AutoFieldsProps = {
   element?: ComponentType | string;
   fields?: string[];
   omitFields?: string[];
+  showInlineError?: boolean;
 };
 
 export default function AutoFields({
@@ -15,6 +16,7 @@ export default function AutoFields({
   element = 'div',
   fields,
   omitFields = [],
+  showInlineError,
   ...props
 }: AutoFieldsProps) {
   const { schema } = useForm();
@@ -24,6 +26,12 @@ export default function AutoFields({
     props,
     (fields ?? schema.getSubfields())
       .filter(field => !omitFields.includes(field))
-      .map(field => createElement(autoField, { key: field, name: field })),
+      .map(field =>
+        createElement(autoField, {
+          key: field,
+          name: field,
+          ...(showInlineError !== null ? { showInlineError } : {}),
+        }),
+      ),
   );
 }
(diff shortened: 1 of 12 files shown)


packages/uniforms-antd/src/AutoFields.tsx
```diff
@@ -24,6 +26,12 @@ export default function AutoFields({
     props,
     (fields ?? schema.getSubfields())
       .filter(field => !omitFields.includes(field))
-      .map(field => createElement(autoField, { key: field, name: field })),
+      .map(field =>
+        createElement(autoField, {
+          key: field,
+          name: field,
+          ...(showInlineError !== null ? { showInlineError } : {}),
+        }),
```
radekmie, line 34: Two things here. One, it should be `undefined`, not `null`. Two, we can get rid of one temporary object. My suggestion:
```suggestion
        createElement(
          autoField,
          Object.assign(
            { key: field, name: field },
            showInlineError === undefined ? null : { showInlineError },
          ),
        ),
```

radekmie (changes requested):
I think it's enough :+1:

packages/uniforms-bootstrap3/__tests__/AutoFields.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { AutoField } from 'uniforms-antd';
 import { AutoFields } from 'uniforms-bootstrap3';
```
radekmie, line 3: It should use `AutoField` from the `uniforms-bootstrap3` theme. Same for the others.

radekmie (changes requested):

radekmie (approved):

PR #928, opened 2021-04-12
Added success field style in Bootstrap themes (closes #899).

### LINKED ISSUE #899: Success field for form validation - wrapField(uniforms-bootstrap4)
Hello, 

thank you very much for your project, we are already using `uniforms` in a production. 
I am using `wrapField` from `uniforms-bootstrap4`. When I validate my filled form, and there are errors in a field an error class (`'is-invalid'`)  via `props.error` is set. The invalid fields are painted with a red border and one small red icon is shown. Unfortunately, this behavior is missing for valid inputs in green. I would like to be able to show the user when his input is valid, also with the bootstrap class (`'is-valid'`). 
My idea is to adapt in `wrapField.tsx:56` these lines: 
```
                "is-invalid": error,
                "is-valid": !error && !!value,
            })}
```

Would you be interested in this feature? Thank you for considering my request.

Monteth:
This PR introduces new style for successfully validated fields in Bootstrap themes (requested in #899)
- [x] uniforms-bootstrap3
- [x] uniforms-bootstrap4
- [x] uniforms-bootstrap5

I tried to stay consistent with current success styles, here are some examples from the playground:
image
image

### DIFF
--- a/packages/uniforms-bootstrap3/src/wrapField.tsx
+++ b/packages/uniforms-bootstrap3/src/wrapField.tsx
@@ -8,6 +8,7 @@ import gridClassName from './gridClassName';
 type WrapperProps = Override<
   Omit<HTMLProps<HTMLDivElement>, 'onChange'>,
   {
+    changed?: boolean;
     error?: unknown;
     errorMessage?: string;
     feedbackable?: boolean;
@@ -25,6 +26,7 @@ type WrapperProps = Override<
 // eslint-disable-next-line complexity
 export default function wrapField(
   {
+    changed = false,
     className,
     disabled,
     error,
@@ -59,6 +61,7 @@ export default function wrapField(
       className={classnames(className, 'field', 'form-group', {
         'has-feedback': error && feedbackable,
         'has-error': error,
+        'has-success': !error && changed,
         disabled,
         required,
       })}
--- a/packages/uniforms-bootstrap3/src/wrapField.tsx
+++ b/packages/uniforms-bootstrap3/src/wrapField.tsx
@@ -26,7 +26,7 @@ type WrapperProps = Override<
 // eslint-disable-next-line complexity
 export default function wrapField(
   {
-    changed = false,
+    changed,
     className,
     disabled,
     error,
--- a/packages/uniforms-bootstrap4/src/BoolField.tsx
+++ b/packages/uniforms-bootstrap4/src/BoolField.tsx
@@ -42,7 +42,6 @@ function Bool({ onChange, ...props }: BoolFieldProps) {
         <input
           checked={value || false}
           className={classnames('form-check-input', {
-            'form-check-input': props.error,
             'is-valid': !props.error && props.changed,
           })}
           disabled={disabled}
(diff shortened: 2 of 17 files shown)


packages/uniforms-bootstrap4/src/BoolField.tsx
```diff
@@ -32,14 +32,19 @@ function Bool({ onChange, ...props }: BoolFieldProps) {
...
+        'text-success':
+          !props.error && props.value !== undefined && props.changed,
       })}
     >
       <label htmlFor={props.id} className="form-check-label">
         <input
           checked={value || false}
-          className="form-check-input"
+          className={classnames('form-check-input', {
+            'form-check-input': props.error,
```
wadamek65, line 45: Seems like one of these should be removed? As it is `form-check-input` will always be present.

packages/uniforms-bootstrap4/src/BoolField.tsx
```diff
@@ -32,14 +32,19 @@ function Bool({ onChange, ...props }: BoolFieldProps) {
     { ...props, label: labelBefore, value: props.value },
     <div
       className={classnames(inputClassName, 'form-check', 'checkbox', {
-        'text-danger': error,
         'custom-control-inline': inline,
+        'text-danger': error,
+        'text-success':
+          !props.error && props.value !== undefined && props.changed,
```
wadamek65, line 38: Shouldn't this logic be consistent with the logic for adding the class name to the input?

packages/uniforms-bootstrap4/src/LongTextField.tsx
```diff
@@ -16,6 +16,7 @@ function LongText(props: LongTextFieldProps) {
     <textarea
       className={classnames(props.inputClassName, 'form-control', {
         'is-invalid': props.error,
+        'is-valid': !props.error && !!props.value && props.changed,
```
wadamek65, line 19: Same, the logic for the validity differs in a lot of places:

`!props.error && !!props.value && props.changed`

`!props.error && props.value !== undefined && props.changed`

`!props.error && props.changed`

Is there a reason for this?

wadamek65 (changes requested):

packages/uniforms-bootstrap4/src/BoolField.tsx
Monteth: The `text-success` was added here accidentally - if we want to stay consistent with the 'error' style we should stylize only the label. Thanks for this.

packages/uniforms-bootstrap4/src/LongTextField.tsx
Monteth: I tried to not stylize empty fields depending on their value type, but now I realized that `!error && changed` is probably all I wanted to check.

wadamek65 (dismissed):

wadamek65 (approved):

packages/uniforms-bootstrap3/src/wrapField.tsx
```diff
@@ -25,6 +26,7 @@ type WrapperProps = Override<
 // eslint-disable-next-line complexity
 export default function wrapField(
   {
+    changed = false,
```
radekmie, line 29: Do we need this default value?

radekmie (approved):

radekmie (approved):

PR #931, opened 2021-04-15
Setup @testing-library, added test suites for all themes for TextField and ListField.

wadamek65:
In this PR:
- Setup `@testing-library/react` and added `@testing-library/user-event` and `@testing-library/jest-dom` helpers.
- Added typings for react-dom.
- Created test suites for all shared tests between all themes for `ListField` and `TextField` components.
- Created a `__suites__` directory in base `uniforms` package along with necessary jest and tsconfig configuration to make it work.


packages/uniforms-antd/__tests__/TextField.tsx
```diff
@@ -1,10 +1,30 @@
+import { screen } from '@testing-library/react';
```
wadamek65, line 1: @radekmie You suggested re-exporting all RTL functions from our own file, but I'm not sure if there is any benefit in doing so with `screen` as it is very specific to this lib. Do you think we should re-export this from `__suites__` as well?

packages/uniforms/__suites__/ListField.tsx
```diff
@@ -0,0 +1,106 @@
...
+        { onChange },
+      ),
+    );
+
+    // Due to differences in how and where the add button is rendered,
+    // we have to use different queries for different themes
+    const addButton =
+      screen.queryByText(/\+/) ??
+      screen.queryAllByRole('img').pop() ??
+      screen.getAllByRole('button')[0];
```
wadamek65, line 96: This I'm not really happy with but there isn't any other way as not only the elements differ between the themes but also their positioning. Other two possible solutions would be to either add a test id to the button in all themes or write this test explicitly for every theme separately.

packages/uniforms/__suites__/TextField.tsx
```diff
@@ -0,0 +1,126 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import createContext, { render } from './renderWithContext';
+
+export function TextFieldTests(TextField: React.ComponentType<any>) {
```
wadamek65, line 7: The typing could be improved but I was having trouble finding an elegant way to get the typings that `connectField` returns. Any suggestions/ideas/hints are welcome.

packages/uniforms/__suites__/ListField.tsx
```diff
@@ -0,0 +1,106 @@
...
+  test('<ListField> - renders ListAddField', () => {
+    render(<ListField name="x" label="ListFieldLabel" />, getDefaultContext());
+
+    expect(screen.getByRole('button')).toBeInTheDocument();
+  });
+
+  test('<ListField> - renders correct label (specified)', () => {
+    render(<ListField name="x" label="ListFieldLabel" />, getDefaultContext());
+
+    // Throws error with no form control found when getting by label
```
wadamek65, line 21: Error is caused due to bad accessibility in this component so it would have to be fixed before we could change the component.

packages/uniforms-antd/__tests__/TextField.tsx
radekmie: No, it's fine. I just thought that the RTL API was slightly different and the `screen` is not _that_ global. It's fine the way it is.

packages/uniforms/__suites__/ListField.tsx
radekmie: Or make the `ListFieldTests` have an option like `findListAddButton`. I believe we'll have more options in the future.

packages/uniforms/__suites__/TextField.tsx
radekmie: On one hand, you can go with something like `Component` from the `createAutoField` helper. On the other, something like `ComponentType<GuaranteedProps<Value>>` (for `TextField`, `Value=string`).

packages/uniforms/__suites__/ListField.tsx
```diff
@@ -0,0 +1,106 @@
...
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import createContext, { render } from './renderWithContext';
+
+function getDefaultContext() {
+  return createContext({ x: { type: Array }, 'x.$': { type: String } });
+}
+
+export function ListFieldTests(ListField: React.ComponentType<any>) {
```
radekmie, line 11: I'd rather go with `runListFieldTests`. Or even `run` and don't re-export from `__suites__`.

packages/uniforms/__suites__/index.ts
```diff
@@ -0,0 +1,5 @@
+import '@testing-library/jest-dom/extend-expect';
```
radekmie, line 1: I'd rather move it to `/scripts/setupMatchers.ts` and add it to `setupFiles` in `jest.config.js`.

packages/uniforms/__suites__/renderWithContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+        readOnly: false,
+        showInlineError: false,
+        ...context?.state,
+      },
+    },
+  };
+}
+
+export function render(
+  element: React.ReactElement,
```
radekmie, line 42: Here, and in other places: I'd rather prefer `ReactElement` instead.

packages/uniforms/__suites__/renderWithContext.tsx
```diff
@@ -0,0 +1,51 @@
+import '@testing-library/jest-dom/extend-expect';
+import { render as rtlRender } from '@testing-library/react';
```
radekmie, line 2: I have a problem with this name, but have no idea for another one... Maybe `renderOnScreen`?

radekmie (changes requested):

packages/uniforms/__suites__/TextField.tsx
wadamek65: `ComponentType<GuaranteedProps<Value>>` reports errors with missing properties that are not passed because they are needed in some tests and `Partial` unfortunately doesn't work for `autoComplete` and `type` props. Any other ideas?

packages/uniforms/__suites__/ListField.tsx
wadamek65: Would you suggest I add a parameter to the run function to pass the custom matcher? Should I do this in this PR or in another one?

radekmie: I'm not sure whether it's called a _matcher_, but if a `() => screen.queryByText(/\+/)` is a _matcher_, then yes. And I'd do it in this one already.

packages/uniforms/__suites__/TextField.tsx
radekmie: Then let's leave it as it is for now.

packages/uniforms/__suites__/renderWithContext.tsx
```diff
@@ -0,0 +1,51 @@
+import '@testing-library/jest-dom/extend-expect';
```
radekmie, line 1: It's no longer needed here, right?

packages/uniforms/__suites__/ListField.tsx
```diff
@@ -0,0 +1,106 @@
...
+function getDefaultContext() {
+  return createContext({ x: { type: Array }, 'x.$': { type: String } });
+}
+
+function defaultAddFieldLocator(): HTMLElement | null | undefined {
+  return screen.getAllByRole('button')[0];
+}
+export function runListFieldTests(
+  ListField: ComponentType<any>,
+  addFieldLocator = defaultAddFieldLocator,
```
radekmie, line 16: 1. Let's put it in an object.
2. I'd rather not have a default one at all.

radekmie (changes requested):

packages/uniforms/__suites__/ListField.tsx
wadamek65: Ad. 1. - do you suggest also changing the API for text field suite even if it only accepts one argument?

radekmie: No no, wrap just the options.

radekmie (approved):

Monteth (approved):

Issue #919, opened 2021-04-01
`anyOf` fields with empty required arrays are treated as required


alexkreidler:
I have a JSON Schema:
```json
"TextBaseline": {
    "anyOf": [
        {
            "const": "alphabetic",
            "type": "string"
        },
        {
            "$ref": "#/definitions/Baseline"
        },
        {
            "const": "line-top",
            "type": "string"
        },
        {
            "const": "line-bottom",
            "type": "string"
        }
    ]
}
```
Which becomes
```json
"TextBaseline": {
    "anyOf": [
        {
            "const": "alphabetic",
            "type": "string"
        },
        {
            "enum": [
                "top",
                "middle",
                "bottom"
            ],
            "type": "string"
        }
        {
            "const": "line-top",
            "type": "string"
        },
        {
            "const": "line-bottom",
            "type": "string"
        }
    ]
}
```
and then the compiled schema becomes:
```json
{
    "anyOf": [
        {
            "const": "alphabetic",
            "type": "string"
        },
        {
            "enum": [
                "top",
                "middle",
                "bottom"
            ],
            "type": "string"
        },
        {
            "const": "line-top",
            "type": "string"
        },
        {
            "const": "line-bottom",
            "type": "string"
        }
    ],
    "properties": {},
    "required": [],
    "type": "string",
    "isRequired": false
}
```

Now a string Field is rendered, but with the property `required` set to an empty array. Surprisingly, this results in the field being required. 

The interface shows the required asterisk, and I cannot submit the form because it doesn't validate.

![uniforms_required](https://user-images.githubusercontent.com/11166947/113359491-e1fd4400-9315-11eb-8cff-2b1e74cdd0a5.png)
![uniforms_devtools](https://user-images.githubusercontent.com/11166947/113359673-43251780-9316-11eb-9521-b620a7bf9c2a.png)

Please fix so an empty array means the field is not required. 

Thanks for this great library!

radekmie:
Hi @alexkreidler. I see where the problem is (field receives `required: []`), but I'm not sure how to proceed here. _Technically_, `required` should be ignored for non-objects. Maybe changing [this condition](https://github.com/vazco/uniforms/blob/657ad910c64f6347458d128ecaea7c1c7bc8e23d/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts#L261-L263) to include non-booleans would be enough. If you'd like to, give it a try. If not, we'll look into it.

alexkreidler:
[...]

PR #936, opened 2021-04-21
Added conditional properties and required field assignment in uniforms-bridge-json-schema (closing #919).

Monteth:
Fixed a bug where a field with no computed required subfields gets an empty 'required' prop. The same situation occurs with the 'properties'.

### DIFF
--- a/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
@@ -728,10 +728,17 @@ describe('JSONSchemaBridge', () => {
       });
     });
 
-    it('works with anyOf for a non-object computed property', () => {
+    it('works with anyOf for a non-object computed property (required default value)', () => {
       expect(bridge.getProps('nonObjectAnyOf')).toHaveProperty(
         'required',
         false,
+      );
+    });
+
+    it('works with anyOf for a non-object computed property (properties not defined)', () => {
+      expect(bridge.getProps('nonObjectAnyOf')).toHaveProperty(
+        'properties',
+        undefined,
       );
     });
 
--- a/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
@@ -166,10 +166,9 @@ describe('JSONSchemaBridge', () => {
             minimum: 0,
           },
         ],
-        required: true,
       },
     },
-    required: ['dateOfBirth'],
+    required: ['dateOfBirth', 'nonObjectAnyOfRequired'],
   };
 
   const validator = jest.fn();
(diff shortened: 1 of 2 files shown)


radekmie (approved):
Add tests, please.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -194,20 +194,27 @@ export default class JSONSchemaBridge extends Bridge {
         .filter(Boolean);
 
       if (combinedPartials.length) {
-        _definition.properties = definition.properties ?? {};
-        _definition.required = definition.required ?? [];
+        const localProperties = definition.properties ?? {};
+        const localRequired = definition.required ?? [];
```
radekmie, line 198: With this approach, we'll mutate `definition` and it'd be better not to. How about this:
```suggestion
        const localProperties = definition.properties ? { ...definition.properties } : {};
        const localRequired = definition.required ? definition.required.slice() : [];
```

radekmie (changes requested):
Also, let's add one more test with non-object field with `anyOf` that is required.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -194,20 +194,33 @@ export default class JSONSchemaBridge extends Bridge {
...
-        _definition.properties = definition.properties ?? {};
-        _definition.required = definition.required ?? [];
+        const localProperties = definition.properties
+          ? { ...definition.properties }
+          : {};
+        const localRequired = definition.required
+          ? Array.isArray(definition.required)
+            ? definition.required.slice()
+            : definition.required
+          : [];
```
Monteth, line 204: From what I know the `definition.required` can be either array or boolean, so I clone an array or assign the previous boolean value here.

packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
```diff
@@ -151,6 +151,23 @@ describe('JSONSchemaBridge', () => {
...
+          {
+            enum: ['top', 'middle', 'bottom'],
+            type: 'string',
+          },
+          {
+            type: 'number',
+            minimum: 0,
+          },
+        ],
+        required: true,
```
radekmie, line 169: It shouldn't be here but in the `required` array two lines later, where the `dateOfBirth` is.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
radekmie: No, [it should be an array](https://json-schema.org/understanding-json-schema/reference/object.html#id3).

packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
Monteth: Yes, and this mistake caused the mess with the `definition.required` not being an array. The new commit fixes it all. 
Thanks for the clarification.

radekmie (approved):

PR #937, opened 2021-04-21
Add node 16 to CI workflow.

Monteth:
Added future-LTS and the current version of node (16)  to GitHub CI workflow.


radekmie (approved):
With this change, I'd go and change our website building to Node 16 as well (these `if`s in the later steps).

radekmie (approved):

PR #942, opened 2021-04-30
Improved test coverage

radekmie:
This PR targets to improve our test coverage. Hopefully, we'll reach 100% once again! To do so, I:
* Added `/* istanbul ignore next */` to our `btoa` and `global` polyfills.
* Configured `tsconfig.json` for `target: 'ESNext'` due to https://github.com/istanbuljs/nyc/issues/727.
* Added and/or updated unit tests.
    * This actually highlighted a slight regression with handling `label={null}`. Nice!
* Removed unnecessary optional chaining.

### DIFF
--- a/packages/uniforms/__tests__/connectField.tsx
+++ b/packages/uniforms/__tests__/connectField.tsx
@@ -55,7 +55,7 @@ describe('connectField', () => {
     } as Context<any>,
   };
 
-  const Test = jest.fn(() => null);
+  const Test = jest.fn(props => props.children || null);
 
   beforeEach(() => {
     Test.mockClear();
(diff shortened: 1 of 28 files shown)


packages/uniforms/__tests__/connectField.tsx
```diff
@@ -184,4 +192,48 @@ describe('connectField', () => {
       },
     );
   });
+
+  describe('when rendered provides correct onChange', () => {
+    it('is defaults to field name', () => {
+      const Field = connectField(Test);
+      const value = 'some value';
+      const wrapper = mount(<Field name="another" />, reactContext);
+      wrapper.find(Test).prop('onChange')(value);
```
wadamek65, line 201: Would it be possible to write these tests in a way that it would be easier to refactor to `react-testing-library` if needed in the future?

wadamek65 (approved):

packages/uniforms/__tests__/connectField.tsx
radekmie: I believe we cannot call `props` of a component in RTL, right? I _could_, however, rewrite it using stubs instead.

wadamek65: Yes that's right, there is no access to the component props in RTL. If it is too time consuming to change these tests then I would say to not do it within this PR, but just wanted to mention this.

Monteth (approved):

PR #961, opened 2021-05-17
Improved DateField label in Material-UI theme.

martachrzanowska:
Improved DateField label in Material-UI theme.

Before:
image

After:
image

### DIFF
--- a/packages/uniforms-material/src/DateField.tsx
+++ b/packages/uniforms-material/src/DateField.tsx
@@ -44,7 +44,7 @@ function Date({
       fullWidth
       helperText={(error && showInlineError && errorMessage) || helperText}
       label={label}
-      InputLabelProps={{ ...labelProps, ...InputLabelProps }}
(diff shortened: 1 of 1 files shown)


packages/uniforms-material/src/DateField.tsx
```diff
@@ -44,7 +44,7 @@ function Date({
       fullWidth
       helperText={(error && showInlineError && errorMessage) || helperText}
       label={label}
-      InputLabelProps={{ ...labelProps, ...InputLabelProps }}
+      InputLabelProps={{ ...labelProps, ...InputLabelProps, shrink: true }}
```
radekmie, line 47: It'd be better to set is a default and let end-users decide:
```suggestion
      InputLabelProps={{ shrink: true, ...labelProps, ...InputLabelProps }}
```

radekmie (changes requested):
Could you describe how it looks before and after? If possible please do attach screenshots.

radekmie (approved):

PR #971, opened 2021-05-26
Improved field type inference (fixes #951).

### LINKED ISSUE #951: uniforms-antd. SelectField. Typing improvements to provide access to antd native props
Motivation:
Ability to pass native antd props like "showSearch", "optionFilterProp", and others.
It already works. As far as I understand via filterDOMProps. But there are some issues with types.
 
Solution:
I suppose the root issue is of using Omit on the union as Omit looks in such a case only on the intersection of props. I attached a solution that works.

Please have a look. If it makes sense for you, I'll create a PR

image

radekmie:
This PR aims to improve the `Override` type as suggested in https://github.com/vazco/uniforms/issues/951#issuecomment-836944631. It revealed a few unnecessary props in the `omit` call as well as some other quirks of `SelectField` in `uniforms-material`.

If possible, @aburtsev, please do check this code.

### DIFF
--- a/packages/uniforms-antd/__tests__/SelectField.tsx
+++ b/packages/uniforms-antd/__tests__/SelectField.tsx
@@ -388,7 +388,7 @@ test('<SelectField checkboxes> - renders a set of checkboxes with correct option
     <SelectField
       checkboxes
       name="x"
-      transform={(x: string) => x.toUpperCase()}
+      transform={(x: unknown) => String(x).toUpperCase()}
     />
   );
   const wrapper = mount(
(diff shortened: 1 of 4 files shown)


wadamek65 (approved):

Monteth (approved):

aburtsev:
> This PR aims to improve the `Override` type as suggested in [#951 (comment)](https://github.com/vazco/uniforms/issues/951#issuecomment-836944631). It revealed a few unnecessary props in the `omit` call as well as some other quirks of `SelectField` in `uniforms-material`.
> 
> If possible, @aburtsev, please do check this code.

I checked the code and checked my use case. Everything works. Thanks!
And I don't see the errors anymore in other libs tests that I saw before. I saw them when I first time tried to apply my fix. Probably some recent types improvements fixed them.

PR #972, opened 2021-05-28
Add autofill property to native props

karolisg:
This change is required for disabling autofill on form level  - https://stackoverflow.com/questions/15738259/disabling-chrome-autofill


radekmie:
Hi @karolisg and sorry for the delay! I believe this PR is not needed, as your change would prevent passing down the `autocomplete` prop, not allow it. Have you tried simply adding `autocomplete="off"` to the form?

karolisg:
> Hi @karolisg and sorry for the delay! I believe this PR is not needed, as your change would prevent passing down the `autocomplete` prop, not allow it. Have you tried simply adding `autocomplete="off"` to the form?

Yes, you're right

PR #984, opened 2021-07-21
Add dependencies duplication and npm vulnerabilities reports to the CI workflow.

Monteth:
Add [dependencies duplication](https://docs.npmjs.com/cli/v7/commands/npm-dedupe) and [npm vulnerabilities](https://docs.npmjs.com/cli/v7/commands/npm-audit) reports to the CI workflow.
The CI run on PR to master branch will produce a warning if there are some possible fixes spotted.

The dependencies duplication report is checked by the `npm dedupe` command, while the vulnerabilities report by the `npm audit fix`.

Example warning in the GH action:
image
And in the workflow logs:
image

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,6 +18,11 @@ jobs:
         with:
           cache: npm
           node-version: ${{ matrix.node-version }}
+      - name: Check dependency vulnerabilities
+        run: |
+          npm audit fix
+          git status -s | grep " M package-lock.json" && echo "::"warning file=package-lock.json"::"Found vulnerabilities in the project dependencies. Run npm audit to show the audit details. More info: https://docs.npmjs.com/cli/v7/commands/npm-audit
+          git reset --hard HEAD
       - name: Check dependency duplications
         run: |
           npm dedupe
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Check differences
         run: git status -s
           | grep " M package-lock.json"
-          && echo "::""warning file=package-lock.json""::"Duplicate dependencies found. Run npm dedupe to remove them, or add --dry-run parameter to get more info."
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -18,6 +18,14 @@ jobs:
...
           cache: npm
           node-version: ${{ matrix.node-version }}
+      - name: Find dependency duplications
+        run: npm dedupe
+      - name: Check differences
+        run: git status -s
+          | grep " M package-lock.json"
+          && echo "warning file=package-lock.json Duplicate dependencies found. Run npm dedupe to remove them, or add --dry-run parameter to get more info."
+      - name: Reset changes
+        run: git reset --hard HEAD
```
radekmie, line 28: 1. I'd rather do it in a single step.
2. Can we add a similar one for `npm audit fix`?

radekmie (changes requested):

.github/workflows/CI.yml
```diff
@@ -18,6 +18,16 @@ jobs:
         with:
           cache: npm
           node-version: ${{ matrix.node-version }}
+      - name: Check dependency vulnerabilities
+        run: |
+          npm audit fix | : # skip non-0 exit
```
Monteth, line 23: While the non-zero exit code causes the workflow to fail, I covered it with an empty action ":" to allow custom error handling - rising a GH Actions warning in this case.

piotrpospiech (approved):

wadamek65 (approved):
BTW does dedupe work only for the root `package-lock.json`? If so it may be worth looking into to also dedupe package dependencies (especially the ones that have more than just TypeScript).

.github/workflows/CI.yml
radekmie: `|| true` would be enough.

Monteth:
> BTW does dedupe work only for the root `package-lock.json`?

We discussed that and indeed, it works only for the root level, but it's not an issue in this repo as we have only one `-lock` here.

radekmie (approved):

PR #986, opened 2021-07-24
Add esm target

kestarumper:
This PR introduces a new build target _ES5 with modules_ and changes the `"module"` keyword in _package.json_ to point to the `esm/index.js` build instead of `es6/index.js`. 

Before we used to have ES6 compiled code set as the `"module"` which does not support full ES6 feature. It only allows ES5 with `import/export` module resolution. This caused problems described in [this card](https://github.com/orgs/vazco/projects/1#card-50179677).

<details>
<summary>Screenshot</summary>

image
</details>


kestarumper:
I did a test by copying the _esm_ built files to the `node_modules/uniforms-bridge-json-schema` and I can confirm that Gatsby now builds correctly (you can extend the bridge).

CC: @radekmie

wadamek65 (approved):

Monteth (approved):
LGTM

radekmie (approved):

PR #988, opened 2021-07-28
Adjusted uniforms-material's TextField type.

### LINKED ISSUE #987: Incorrectly typed props in TextField from uniforms-material
`TextField` component props inside `uniforms-material` are typed using `StandardTextFieldProps` from `@material-ui/core/TextField`
Inside material-ui, `TextField`'s props are typed using
`type TextFieldProps = StandardTextFieldProps | FilledTextFieldProps | OutlinedTextFieldProps;`
This results in some props being typed incorrectly, for example `variant` prop has type `"standard" | undefined`, while it should be `"standard" | "filled" | "outlined" | undefined`

Using `uniforms-material 3.5.1`

wadamek65:
The previous type was only relevant for the "standard" variant of TextField. This PR changes the used type to the more general TextField one.

Fixes #987

### DIFF
--- a/packages/uniforms-material/src/DateField.tsx
+++ b/packages/uniforms-material/src/DateField.tsx
@@ -1,4 +1,4 @@
-import TextField, { StandardTextFieldProps } from '@material-ui/core/TextField';
+import TextField, { TextFieldProps } from '@material-ui/core/TextField';
 import React from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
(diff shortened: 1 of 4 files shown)


Monteth (dismissed):

radekmie (approved):

piotrpospiech (dismissed):

radekmie:
@wadamek65, @Florian3k I guess we should do the same for `DateField`, `LongTextField`, and `NumField`, right?

wadamek65:
@radekmie I'll check if there is also a discrepancy there and will adjust the types there as well if there is.

Florian3k (approved):

kestarumper (approved):

radekmie (approved):
@Florian3k Can you confirm that it does the job?

kestarumper (approved):

PR #993, opened 2021-08-06
Make uniforms-material theme use MUI theme props as default.

wadamek65:
Since MUI's `useTheme` returns a default theme even if no provider is present, we can safely use it and add logic for extracting default props.

I've also decided against defaulting to theme props in list-related components (`ListAddField`, `ListDelField`, `ListField`, `ListItemField`) since IMO they are very opinionated and rely on these custom props heavily. Using default props could break the out-of-the-box functionality and force the user to manually specify all fields instead of just using `ListField`.

Unfortunately, since `wrapField` is called conditionally in `SelectField` I could not use the hook there and had to manually add the logic in every component that uses it.

I've noticed no visual regressions in the playground (screenshot below), but it would be nice if someone also double-checked it to make sure I didn't miss anything.

image

Also improved some `margin` types by using the same types as MUI.


packages/uniforms-material/src/BoolField.tsx
```diff
@@ -1,3 +1,4 @@
+import { useTheme } from '@material-ui/core';
```
radekmie, line 1: Due to implied tree shaking, can we import `@material-ui/core/styles` or even `@material-ui/core/styles/useTheme` instead? Just like we do with the components.

packages/uniforms-material/src/BoolField.tsx
```diff
@@ -33,11 +34,18 @@ function Bool(props: BoolFieldProps) {
...
+  const theme = useTheme();
+  const formControlThemeProps = theme.props?.MuiFormControl;
   const SelectionControl =
     appearance === 'checkbox' || appearance === undefined ? Checkbox : Switch;
 
   return wrapField(
-    { fullWidth: true, margin: 'dense', ...props, component: 'fieldset' },
+    {
+      fullWidth: formControlThemeProps?.fullWidth ?? true,
+      margin: formControlThemeProps?.margin ?? 'dense',
```
radekmie, line 45: How about _not_ setting them at all if they are configured in the theme? That's probably even more idomatic.

packages/uniforms-material/src/ErrorField.tsx
```diff
@@ -19,12 +20,15 @@ function Error({
...
+
   return !error ? null : (
     <FormControl
       error={!!error}
-      fullWidth={!!fullWidth}
-      margin={margin === 'dense' ? margin : undefined}
-      variant={variant}
+      fullWidth={themeProps?.fullWidth ?? !!fullWidth}
+      margin={themeProps?.margin ?? margin}
+      variant={themeProps?.variant ?? variant}
```
radekmie, line 31: To match other fields this should do something like this:
```suggestion
      fullWidth={fullWidth ?? themeProps?.fullWidth ?? true}
      margin={margin ?? themeProps?.margin ?? 'dense'}
      variant={variant}
```

Meaning that you can still override the theme with props.

packages/uniforms-material/src/ErrorField.tsx
```diff
@@ -1,3 +1,4 @@
+import { PropTypes, useTheme } from '@material-ui/core';
```
radekmie, line 1: Can you check whether it's emitted in the JavaScript code? Again, I'd rather not import the entire `@material-ui/core` anywhere.

packages/uniforms-material/src/LongTextField.tsx
```diff
@@ -19,24 +20,29 @@ const LongText = ({
...
+  const theme = useTheme();
+  const themeProps = theme.props?.MuiTextField;
+
+  return (
+    <TextField
+      disabled={disabled}
+      error={!!error}
+      fullWidth={themeProps?.fullWidth ?? true}
+      helperText={(error && showInlineError && errorMessage) || helperText}
+      inputProps={{ readOnly, ...(themeProps?.inputProps ?? {}) }}
```
radekmie, line 33: Do you think this is the right order here? I doubt that someone will ever specify theme-wise `readOnly` flag, but I think we'd like to override it. Or even override it only if there's `true` set.

Also, there's no need to `?? {}` here.

radekmie (changes requested):

radekmie:
Also, can we add tests to verify that theming actually works as expected? That would include defaults/theme/props/theme+props.

packages/uniforms-material/src/ErrorField.tsx
kestarumper: We have to remember to remove the default values though (unless we want to set them to `null` to indicate that we want to take the value from theme but I guess this is weird).

kestarumper (changes requested):
In general, I would try and unify the order of overriding the props. I would go with the one suggested by @radekmie in [this comment](https://github.com/vazco/uniforms/pull/993/files#r684960838).

**Importance (descending):**
1. Direct props
2. Theme  props
3. Default props

packages/uniforms-material/__tests__/BoolField.tsx
```diff
@@ -3,12 +3,76 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
...
+      expect.objectContaining({
+        fullWidth: true,
+        margin: 'dense',
+      }),
+      undefined,
+      expect.anything(),
+    );
+  });
+
+  test('<BoolField> - explicit props are passed when MUI theme props are specified', () => {
```
wadamek65, line 53: @radekmie While I was writing this test I realized that `BoolField` doesn't even accept these props explicitly. Is this an oversight in types? Because if it's not and we don't want it to have these props then we probably shouldn't be overriding the defaults here.

packages/uniforms-material/src/ErrorField.tsx
```diff
@@ -1,8 +1,9 @@
-import { PropTypes, useTheme } from '@material-ui/core';
+import type { PropTypes } from '@material-ui/core';
```
radekmie, line 1: I think we are not using `import type` anywhere. Let's stick to `import`s, at least for now.

packages/uniforms-material/__tests__/BoolField.tsx
radekmie: It depends. If it's not needed from the point of the UI, then sure, let's not change it. But if it is, then I'd rather update the types. Overall, it'll work in runtime.

packages/uniforms-material/src/ErrorField.tsx
wadamek65: I added this as you were concerned whether or not the whole MUI package would be emitted: https://github.com/vazco/uniforms/pull/993#discussion_r684961200

This ensures that it will not be included in the compiled code and I think overall is a good practice in cases like these that doesn't require us to manually check whether or not every import will get tree shaken (as I haven't found this being exported from anywhere else).

wadamek65:
@radekmie I've pushed the tests for `BoolField`. I think these should suffice but before I go ahead and replicate them for all other components would you mind taking a look and confirming that these are OK?

packages/uniforms-material/src/ErrorField.tsx
radekmie: I understand, but then I'd rather use `import type` everywhere. Ideally, ESLint or `tsc` should tell us that an import should be type-only.

radekmie:
Yep, all tests look fine.

Monteth (dismissed):

kestarumper (approved):

packages/uniforms-material/src/RadioField.tsx
```diff
@@ -38,8 +37,16 @@ function Radio({
   value,
   ...props
 }: RadioFieldProps) {
+  const theme = useTheme();
+  const formControlThemeProps = theme.props?.MuiFormControl;
   return wrapField(
-    { ...props, component: 'fieldset', disabled, fullWidth, margin },
+    {
+      fullWidth: formControlThemeProps?.fullWidth ?? true,
```
piotrpospiech, line 44: isn't it all the time `true`?

packages/uniforms-material/src/TextField.tsx
```diff
@@ -20,15 +21,18 @@ function Text({
...
 }: TextFieldProps) {
+  const theme = useTheme();
+  const themeProps = theme.props?.MuiTextField;
+
   return (
     <TextField
       disabled={disabled}
       error={!!error}
-      fullWidth
+      fullWidth={themeProps?.fullWidth ?? true}
```
piotrpospiech, line 31: isn't it all the time `true`?

packages/uniforms-material/src/DateField.tsx
```diff
@@ -37,16 +38,19 @@ function Date({
...
 }: DateFieldProps) {
+  const theme = useTheme();
+  const themeProps = theme.props?.MuiTextField;
+
   return (
     <TextField
       disabled={disabled}
       error={!!error}
-      fullWidth
+      fullWidth={themeProps?.fullWidth ?? true}
```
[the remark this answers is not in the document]
wadamek65: No, if `themeProps?.fullWidth` is equal to `false` then it will be `false`. It's true only if it's not defined. (Same to your other comments)

radekmie (approved):

kestarumper (approved):

Monteth (approved):

flean:
How does this work, I have a project and Autoforms doesn't use my theme

radekmie:
Hi @flean. Based on #1038 that you've reported, you're using the new `@mui/material` package and `uniforms-material` are based on `@material-ui/core`. I'll answer you in the other issue, but if it won't be all, then feel free to comment here.

PR #999, opened 2021-08-18
Update DateField.tsx fix timezone problem

minhna:
To make the date field displays the local timezone. It's confusing displaying UTC timezone.

### DIFF
--- a/packages/uniforms-material/src/DateField.tsx
+++ b/packages/uniforms-material/src/DateField.tsx
@@ -4,7 +4,7 @@ import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
 /* istanbul ignore next */
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
-const dateFormat = (value?: Date) => value && value.toISOString().slice(0, -8);
+const dateFormat = (value?: Date) => value && value.toString().slice(0, -8);
 const dateParse = (timestamp: number, onChange: DateFieldProps['onChange']) => {
   const date = new DateConstructor(timestamp);
   if (date.getFullYear() < 10000) {


radekmie:
Hi @minhna. Unfortunately, your change is incorrect. The format we are using now (`YYYY-MM-DDTHH:mm`) is the one that `value` of a `input[type=datetime-local]` should have ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#value)). What is important, it is **not** depending on user locale - the standard defines what is a [date](https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#date_strings), [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#time_strings), and [local datetime](https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#local_date_and_time_strings).

If you'd like to have a better UX with a custom date picker, do implement it as a custom field instead. This topic was raised many times, and there are no plans for supporting anything else than `datetime-local` for dates in the core themes.

minhna:
I apology the wrong format, but it has "local" in name then it's only local when you live in England.

radekmie:
No, as it's entirely up to you, what you'll do with that date.

I don't want to go into details here, but from my experience, it's much easier to use UTC dates everywhere and adjust them only on display. If you really want to convert them into a specified timezone (or the one the user has; keep in mind your server won't know which one is it), you can do it in the `onSubmit` handler.

Ideally, we'd use something like [`Temporal.PlainDate`](https://tc39.es/proposal-temporal/docs/#Temporal-PlainDate) or [`Temporal.PlainDateTime`](https://tc39.es/proposal-temporal/docs/#Temporal-PlainDateTime), but I think it'll take at least a couple of years to be widely used.

PR #1000, opened 2021-08-18
Fix initialValue in NumField (closes #969).

### LINKED ISSUE #969: Material NumField with defaultValue cannot be cleared
When using uniforms, simpl-schema (v2) and material, all at version 3.5.1 (latest version as of writing) when we clear a NumberField (in an AutoForm with an attached schema and model) the defaultValue will be set, whereas this works fine for a TextField.

Assuming the defaultValue of a NumberField is 10, when I backspace 0 it will show 1, but once I backspace 1 it will show 10 again.

Playground example: https://uniforms.tools/playground/#?N4IgDgTgpgzlAuIBcIDqUA2BjA9gWygEIQAacCHMGZUAQwFd4cZaA3KZAM1ozjIaYt2AEUy0AnsgCMABhlkAJgEsWAIwxQFXHnxAZaqzMngR6UMmH1YoACxwYFUCNt7mQ0WgoDyAOwySkblcyGCwbKDxaZBAfKAB3AAIAZSU8SygksIjaACYAIQglBQBzKAAKAB0fBITYxJS0jUzwyLLgKpqaiFp4JR9ipAT26s6a-HEwKEGAOXo8QwgSDtGEx256DHgANR4zQdklkZqAX0OTgEoq89IQGDs4gEk_PqgAUQgKZ0CdN1oYcR8WF8SXoqjwSkQ32CID-AKBPh2GCKPQ4ULgpxA8BaqJAkXgTiUPBAxyAA

I believe these lines combined may be the culprit (setting the value to undefined instead of empty string will cause value to be set to initialValue):
https://github.com/vazco/uniforms/blob/master/packages/uniforms-material/src/NumField.tsx#L50
https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/useField.tsx#L85

Monteth:
Before you open a pull request, please check if a similar one already exists or has been closed before. Detailed information about the process of contributing can be found in [CONTRIBUTING.md](https://github.com/vazco/uniforms/blob/master/.github/CONTRIBUTING.md).

Do link all relevant issues and pull requests. If there is none, please do file an issue first to discuss it upfront.

### DIFF
--- a/packages/uniforms/src/useField.tsx
+++ b/packages/uniforms/src/useField.tsx
@@ -81,11 +81,13 @@ export function useField<
   let initialValue: Value | undefined;
   let value: Value | undefined = props.value ?? valueFromModel;
 
-  if (value === undefined) {
-    value = context.schema.getInitialValue(name, props);
-    initialValue = value;
-  } else if (props.value !== undefined && props.value !== valueFromModel) {
-    initialValue = props.value;
+  if (!changed) {
+    if (value === undefined) {
+      value = context.schema.getInitialValue(name, props);
+      initialValue = value;
+    } else if (props.value !== undefined && props.value !== valueFromModel) {
+      initialValue = props.value;
+    }
   }
 
   if (options?.initialValue !== false) {


wadamek65 (approved):

packages/uniforms/src/useField.tsx
```diff
@@ -81,14 +85,17 @@ export function useField<
...
-  if (value === undefined) {
-    value = context.schema.getInitialValue(name, props);
-    initialValue = value;
-  } else if (props.value !== undefined && props.value !== valueFromModel) {
-    initialValue = props.value;
-  }
-
   if (options?.initialValue !== false) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    onChangeCalled = useRef(false);
```
radekmie, line 90: I think that it'd be better to do something like this:
```ts
const usesInitialValue = options?.initialValue !== false;
const onChangeCalled = usesInitialValue ? useRef(false) : { current: false };
```

Then this code as well as the `onChange` handler will be easier.

radekmie (approved):

wadamek65 (approved):

kestarumper (approved):

PR #1003, opened 2021-08-24
Added explicit form type casting.

wadamek65:
This PR introduces casting of all form types explicitly across all themes where previously they were typed as any due to TS constraints.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -245,3 +245,21 @@ const ChangedForm = ({ model }) => (
   </AutoForm>
 );
 ```
+
+### Why am I suddenly getting type errors in my form components?
+
+After introduction of TypeScript in `uniforms@3.0.0`, in the initial versions all form components in theme packages were typed as `any`.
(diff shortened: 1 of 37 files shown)


radekmie (approved):
It is definitely better than it was now. However, we'd need to prepare solid and clear documentation of the change, including workarounds (if needed).

wadamek65:
@radekmie I'd be happy to add the documentation within this PR if you could provide a little more details on where it should be.

radekmie:
1. A short comment for the `CHANGELOG.md` about this change in general.
2. A note in the FAQ section, on what to do when these typings actually broke somebody's code (i.e., cast to `any` and report us a bug).

Monteth (approved):

wadamek65:
@radekmie I've added some docs in the FAQ section. I suppose it will be a good basis for when the changelog is actually created or should I also prepare this and put it somewhere right now?

kestarumper (approved):

radekmie (approved):

Issue #1004, opened 2021-08-25
Types are incompatible with `antd@>=4.16.11`


yurtsiv:
After running `tsc` in my project, I got the following error:

```
node_modules/uniforms-antd/es5/ListAddField.d.ts:13:86 - error TS2694: Namespace '"/home/stepy/Dev/team-cv/node_modules/antd/lib/_util/type"' has no exported member 'Omit'.                                       
                                                                                                                                                                                                                   
13 } & import("antd/lib/button/button").BaseButtonProps & import("antd/lib/_util/type").Omit<React.AnchorHTMLAttributes<any>, "type" | "onClick"> & { 
```

(there are a few more errors, but the reason is the same)

**Packages used:**

```
uniforms@3.5.5
uniforms-antd@3.5.5
antd@4.16.11
```

**Cause:**

`Omit` type got removed in [antd@4.16.11](https://github.com/ant-design/ant-design/blob/master/CHANGELOG.en-US.md#41611)

**Possible solution:**

Bump up `antd` version. The PR is coming soon.

PR #1005, opened 2021-08-25
Fixed types in uniforms-antd for AntD v4.16.13

yurtsiv:
Relevant issue: https://github.com/vazco/uniforms/issues/1004

As proposed, I updated `antd` to v4.16.13 which caused a little type mismatch in `NumberField` which I've fixed.

### DIFF
--- a/packages/uniforms-antd/__tests__/NumField.tsx
+++ b/packages/uniforms-antd/__tests__/NumField.tsx
@@ -185,10 +185,10 @@ test('<NumField> - renders an InputNumber which does not react to change (empty)
   );
 
   expect(wrapper.find('input')).toHaveLength(1);
-  expect(
-    wrapper.find('input').simulate('change', { target: { value: '' } }),
-  ).toBeTruthy();
-  expect(onChange).not.toHaveBeenCalled();
+
+  wrapper.find(InputNumber).props().onChange!('');
+
+  expect(onChange).toHaveBeenLastCalledWith('x', undefined);
 });
 
 test('<NumField> - renders an InputNumber which does not react to change (same value)', () => {
@@ -201,10 +201,10 @@ test('<NumField> - renders an InputNumber which does not react to change (same v
   );
 
   expect(wrapper.find('input')).toHaveLength(1);
-  expect(
-    wrapper.find('input').simulate('change', { target: { value: '1' } }),
-  ).toBeTruthy();
-  expect(onChange).not.toHaveBeenCalled();
+
+  wrapper.find(InputNumber).props().onChange!('1');
+
+  expect(onChange).toHaveBeenLastCalledWith('x', 1);
 });
 
 test('<NumField> - renders an InputNumber which correctly reacts on change (zero)', () => {
--- a/packages/uniforms-antd/__tests__/NumField.tsx
+++ b/packages/uniforms-antd/__tests__/NumField.tsx
@@ -175,7 +175,7 @@ test('<NumField> - renders an InputNumber which correctly reacts on change (deci
   expect(onChange).toHaveBeenLastCalledWith('x', 2);
 });
 
-test('<NumField> - renders an InputNumber which does not react to change (empty)', () => {
+test('<NumField> - renders an InputNumber which correctly reacts on change (empty)', () => {
   const onChange = jest.fn();
 
   const element = <NumField name="x" />;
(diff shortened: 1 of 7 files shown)


packages/uniforms-antd/__tests__/NumField.tsx
```diff
@@ -185,13 +185,13 @@ test('<NumField> - renders an InputNumber which correctly reacts on change (empt
   );
 
   expect(wrapper.find('input')).toHaveLength(1);
-  expect(
-    wrapper.find('input').simulate('change', { target: { value: '' } }),
-  ).toBeTruthy();
+
+  wrapper.find(InputNumber).props().onChange!('');
```
radekmie, line 189: I don't understand where is this coming from. Isn't the new version of `InputNumber` based on `input` as well?

radekmie (changes requested):
I don't really like the idea of this PR. The problem is that our `.d.ts` files are referencing internal AntD types, not that there's some other problem. I'd rather look into that first. For now, please do answer my question about the test.

radekmie:
As I've checked, this type-only change is enough:
```diff
diff --git a/packages/uniforms-antd/src/ListAddField.tsx b/packages/uniforms-antd/src/ListAddField.tsx
index 1bca2c25..ced75843 100644
--- a/packages/uniforms-antd/src/ListAddField.tsx
+++ b/packages/uniforms-antd/src/ListAddField.tsx
@@ -3,6 +3,7 @@ import Button, { ButtonProps } from 'antd/lib/button';
 import cloneDeep from 'lodash/cloneDeep';
 import React from 'react';
 import {
+  ConnectedField,
   FieldProps,
   connectField,
   filterDOMProps,
@@ -57,7 +58,12 @@ function ListAdd({
   );
 }
 
-export default connectField<ListAddFieldProps>(ListAdd, {
+// There's no way to tell TypeScript NOT TO expand the type alias. Creating a
+// local type helps, at least in the current version.
+// https://github.com/microsoft/TypeScript/issues/34556
+type ListAddFieldType = ConnectedField<ListAddFieldProps>;
+
+export default connectField(ListAdd, {
   initialValue: false,
   kind: 'leaf',
-});
+}) as ListAddFieldType;

```

radekmie:
@yurtsiv Could you look into this anytime soon?

packages/uniforms-antd/__tests__/NumField.tsx
yurtsiv: It is, but for some reason

```typescript
wrapper.find('input').simulate('change', { target: { value: '' } });
```

doesn't trigger `onChange`. Seems to be related to the [breaking changes](https://ant.design/changelog#4.13.0) in `antd@4.14.0`, although it's not explicitly stated there.

radekmie: OK, and could you check, whether or not the type fix I proposed is enough?

radekmie: Also, it looks like they updated their tests in the same way: https://github.com/ant-design/ant-design/pull/29373/files#diff-7fcfad35ca4d1e8f875d130aa8186b0157946e828e7ffb95908d6f928b3f6e1fR18. On the other hand, https://github.com/react-component/input-number/blob/c4ae6f89c21da3f746c6b40196015cc5c659b484/tests/input.test.tsx#L153-L158 would suggest, that we'd need to blur the input after the change.

yurtsiv:
@radekmie Sorry for such a late response.

How did you test the solution above? After applying your solution and installing the package from the file system, it still expands that type alias. Also, there's the same problem in `ListDelField.tsx`, so if you didn't apply the trick there, it would fail.

I'm using `typescript@4.4.2`

radekmie:
@yurtsiv I only applied it in the repo, build the project, and checked the `.d.ts` file. With this change, the `Omit` type from AntD is not used anymore. And yes, if there are other cases, like `ListDelField`, it has to be applied there as well.

yurtsiv:
@radekmie Ok, for some reason, I thought installing a package also rebuilds it. Seems like it's not the case. Anyway, your solution seems to work fine. I've reverted all the previous changes and added your solution where needed.

radekmie (approved):

wadamek65 (approved):

PR #1022, opened 2021-09-13
Fixed leaf component detection (fixes #1019).

### LINKED ISSUE #1019: createAutoField: Cannot use 'in' operator to search for 'options'
## Problem description
When providing a schema like
```ts
const schema = {
  type: 'object',
  properties: {
    username: {
      uniforms: {
        component: 'TEXTFIELD', // <-- This causes error
      }
    }
  },
  required: ['username'],
};
```
the `createAutoField` throws errors (or rather `AutoField` created from it) when checking for `options` property with the following error:
```
TypeError
Cannot use 'in' operator to search for 'options' in TEXTFIELD
```

## Background
I've tried migrating from a custom AutoField to the one created by `createAutoField` but we were using a _component map pattern_ and we were using the `uniforms.component` schema property for the mapping.
```ts
// example, not an actual code
type componentMap = {
    [key: string]: React.ComponentType<T>;
}
``` 
So when `useField` is executed inside the `AutoField` created by `createAutoField` it uses this `component: "TEXTFIELD"` and then we get an error during check for `'options' in component` because `component` is a `string` and and not an `object`.

## Details
- Affected package: `"uniforms": "3.5.1"`
- Reproduction: [Sandbox](https://codesandbox.io/s/sleepy-shirley-o4mix)

radekmie:
In this PR I fixed the `kind: 'leaf'` component detection, fixing #1019.

### DIFF
--- a/packages/uniforms/src/createAutoField.tsx
+++ b/packages/uniforms/src/createAutoField.tsx
@@ -7,7 +7,7 @@ import {
   useContext,
 } from 'react';
 
-import { connectField } from './connectField';
+import { ConnectedField } from './connectField';
 import { Context } from './types';
 import { useField } from './useField';
 
@@ -18,7 +18,7 @@ export type AutoFieldProps = {
 };
 
 /** @internal */
-export type Component = ComponentType<any> | ReturnType<typeof connectField>;
+export type Component = ComponentType<any> | ConnectedField<any>;
 
 /** @internal */
 export type ComponentDetector = (
@@ -36,7 +36,7 @@ export function createAutoField(defaultComponentDetector: ComponentDetector) {
 
     invariant(component, 'AutoField received no component for: %s', props.name);
 
-    return 'options' in component && component.options?.kind === 'leaf'
+    return isConnectedLeafComponent(component)
       ? createElement(component.Component, props)
       : createElement(component, rawProps);
   }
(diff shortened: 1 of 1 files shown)


packages/uniforms/src/createAutoField.tsx
```diff
@@ -46,3 +46,7 @@ export function createAutoField(defaultComponentDetector: ComponentDetector) {
     defaultComponentDetector,
   });
 }
+
+function isConnectedLeafComponent(x: Component): x is ConnectedField<any> {
+  return typeof x === 'object' && 'options' in x && x.options?.kind === 'leaf';
```
wadamek65, line 51: Isn't the optional chaining redundant here since you check for the existence of `options` property beforehands?
```suggestion
  return typeof x === 'object' && 'options' in x && x.options.kind === 'leaf';
```

wadamek65 (changes requested):

packages/uniforms/src/createAutoField.tsx
radekmie: I just wanted to make sure it'll be fine if someone uses `options: undefined`, because before TypeScript 4.4 it's actually a valid value ([link](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types)).

wadamek65 (approved):

radekmie:
https://github.com/vazco/uniforms/issues/1019#issuecomment-930030063

PR #1076, opened 2021-12-16
Use reusable workflows

Monteth:
As we discussed at the internal planning meeting ([card](https://github.com/orgs/vazco/projects/1#card-72596364)) we want to simplify and unify the ci workflow steps, so I created the [open-standards](https://github.com/vazco/open-standards) repo under the [Vazco](https://github.com/vazco) organization for workflows like npm dedupe and vulnerability checks.
This PR introduces a change to use workflows from [open-standards](https://github.com/vazco/open-standards) repo.

The reusable workflows have to be run as standalone jobs (what distinguishes them from official actions), and we have to deal with its consequences: 
1. We have to run checkout and node setup steps 2 times additionally.
2. We can run both actions only once per workflow run, not like it was - once every node matrix run (6 checks less)  
3. It is a standalone job (obviously), and it brings UI changes - now it is more visible, what can be pros or a con, depending on context or approach. It's great if we want to put more emphasis on this element, but unfavorable if it's not that crucial action.

Explanation picture for point 3.:
screenshot from GitHub actions

References: 
[learn-github-actions/reusing-workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows)

Disclaimer:
This PR probably shouldn't be merged before [this PR](https://github.com/vazco/open-standards/pull/1).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
 jobs:
+  npm_dedupe_check:
+    uses: vazco/open-standards/.github/workflows/check_dependency_duplications.yml@v1
   CI:
     runs-on: ubuntu-latest
     strategy:
@@ -23,8 +25,6 @@ jobs:
           npm audit fix || true # skip non-0 exit
           git status -s | grep " M package-lock.json" && echo "::"warning file=package-lock.json"::"Found vulnerabilities in the project dependencies. Run npm audit to show the audit details. More info: https://docs.npmjs.com/cli/v7/commands/npm-audit
           git reset --hard HEAD
-      - name: Check dependency duplications
-        uses: vazco/open-standards/.github/workflows/check_dependency_duplications.yml@v1
       - name: Install
         run: npm i -g npm@7 && npm ci
       - name: Lint
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,10 +24,7 @@ jobs:
           git status -s | grep " M package-lock.json" && echo "::"warning file=package-lock.json"::"Found vulnerabilities in the project dependencies. Run npm audit to show the audit details. More info: https://docs.npmjs.com/cli/v7/commands/npm-audit
           git reset --hard HEAD
       - name: Check dependency duplications
-        run: |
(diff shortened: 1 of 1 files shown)


wadamek65:
Pasting our discussion on slack for visibility:

Could you perhaps take a look if maybe one of these approaches could be better instead?
* Define the action as a task using a docker action (https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action)
* Take a look at DangerJS to see if maybe it has plugins for these operations (https://danger.systems/js/)

.github/workflows/CI.yml
```diff
@@ -18,16 +21,8 @@ jobs:
...
-        run: |
-          npm audit fix || true # skip non-0 exit
-          git status -s | grep " M package-lock.json" && echo "::"warning file=package-lock.json"::"Found vulnerabilities in the project dependencies. Run npm audit to show the audit details. More info: https://docs.npmjs.com/cli/v7/commands/npm-audit
-          git reset --hard HEAD
-      - name: Check dependency duplications
-        run: |
-          npm dedupe
-          git status -s | grep " M package-lock.json" && echo "::"warning file=package-lock.json"::"Duplicate dependencies found. Run npm dedupe to remove them, or add --dry-run parameter to get more info.
-          git reset --hard HEAD
+      - name: Npm dependency duplication check
```
radekmie, line 24: Can we keep the step name as it was? I think it was better before.

radekmie (changes requested):
Could `npm_vuln_check` be a step as well?

wadamek65 (dismissed):

zendranm (dismissed):

wadamek65:
@Monteth 

> Please take a look if it suits us. I think the vuln report is very nice, but the dedupe is not telling much.

Where can I see that?

Monteth:
> @Monteth
> 
> > Please take a look if it suits us. I think the vuln report is very nice, but the dedupe is not telling much.
> 
> Where can I see that?

Go to Checks
image
Then click on the CI workflow
image

wadamek65:
@Monteth Ah, for some reason I was expecting a separate job and forgot it was built into our current CI workflow 🤦 looks good 👌

PR #1077, opened 2021-12-20
Updated Docusaurus.

radekmie:
In this PR, I updated the Docusaurus to the latest beta version. Most importantly, it includes the migration to the new version of DocSearch - our documentation search provider. They introduced a new version recently, and I simply followed the [migration guide](https://docsearch.algolia.com/docs/migrating-from-legacy/) (including the new `appId` and `apiKey`).


website/package.json
```diff
@@ -22,18 +22,18 @@
...
+    "uniforms-bootstrap3": "3.7.0",
+    "uniforms-bootstrap4": "3.7.0",
+    "uniforms-bootstrap5": "3.7.0",
+    "uniforms-bridge-graphql": "3.7.0",
+    "uniforms-bridge-json-schema": "3.7.0",
+    "uniforms-bridge-simple-schema": "3.7.0",
+    "uniforms-bridge-simple-schema-2": "3.7.0",
+    "uniforms-material": "3.7.0",
+    "uniforms-semantic": "3.7.0",
+    "uniforms-unstyled": "3.7.0",
```
Monteth, line 36: So I guess since now we have to update it manually on every publication of uniforms. What is the cause of this change?

Monteth (approved):

website/package.json
radekmie: I'm convinced that Lerna will actually update them on release.

PR #1078, opened 2021-12-21
JSON Schema with Dot-Fieldnames

### LINKED ISSUE #963: Valid JSON Schema with Dot-Fieldnames breaks JSONSchemaBridge 3.5.1
Hey!

We are using uniforms with uniforms-bridge-json-schema/JSONSchemaBridge.
We have a valid JSON Schema, but the schema is for a tool that has dots in the field/property names like `mystructure.value.enabled`.
Unfortunately JSONSchemaBridge seems to expect that dots in names are addressing a hierarchy instead of just a plain name with dots. That expectation might be true from an HTML form view, but is not valid for JSON/JSON Schema. So we have a valid schema but can't convert it successfully with AutoForm.
We receive a `Field not found in schema: "mystructure.value.enabled"` error.

I expect only when you traverse nested objects in the JSON Schema you should expect a nested structure, but not base this on property names.

Or is there anything I don't see? Help would be really appreciated.

kestarumper:
This PR closes #963.

**NOTE**: This is yet another approach that works, but will probably be changed.

When I was working on the implementation of [2 and 2.5](https://github.com/vazco/uniforms/issues/963#issuecomment-990859246) I encountered some problems which would make the API inconvenient. Instead, I implemented the _escape paths_. This is more convenient when used with React because you can pass a string to the `name` prop instead of an array.

Given schema
```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    "type": "object",
    "properties": {
      "x": {
        "type": "object",
        "properties": {
          "y.z": {
            "type": "string"
          }
        }
      }
    }
}
```

we can access it like the following
```jsx
<AutoField name='x.["y.z"]' />
```

You can escape the path with _brackets_ like `normal.["path.with.dots.in.name"].second.["another.path.with.dots"].0`.

### DIFF
--- a/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
@@ -173,6 +173,32 @@ describe('JSONSchemaBridge', () => {
       objectWithoutProperties: { type: 'object' },
       withLabel: { type: 'string', uniforms: { label: 'Example' } },
       forcedRequired: { type: 'string', uniforms: { required: true } },
+      'path.with.a.dot': {
+        type: 'object',
+        properties: {
+          'another.with.a.dot': {
+            type: 'string',
+          },
+          another: {
+            type: 'object',
+            properties: {
+              with: {
+                type: 'object',
+                properties: {
+                  a: {
+                    type: 'object',
+                    properties: {
+                      dot: {
+                        type: 'number',
+                      },
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
     },
     required: ['dateOfBirth', 'nonObjectAnyOfRequired'],
   };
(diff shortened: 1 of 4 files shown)


packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
```diff
@@ -173,6 +173,32 @@ describe('JSONSchemaBridge', () => {
...
+                        type: 'number',
+                      },
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
```
radekmie, line 201: Please add another test, where the dotted field is nested under a not dotted one. I'm pretty sure it'll work, but it's better to have a test for that.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -150,7 +150,9 @@ export default class JSONSchemaBridge extends Bridge {
         fieldInvariant(name, !!definition);
       } else if (definition.type === 'object') {
         fieldInvariant(name, !!definition.properties);
-        definition = definition.properties[next];
+        definition =
+          definition.properties[next] ||
+          definition.properties[next.slice(2, -2)];
```
radekmie, line 155: How about making a function that'd trim the `[""]`? Then we'd have a similar one for the code in lines 301-303.

packages/uniforms/src/joinName.ts
```diff
@@ -7,7 +9,15 @@ export function joinName(...parts: unknown[]) {
...
-          name.push(...part.split('.'));
+          name.push(
+            ...flatten(
+              part
+                .split(/\[|\]/)
+                .map(subkey =>
+                  /^['"]/.exec(subkey) ? `[${subkey}]` : subkey.split('.'),
+                ),
+            ).filter(Boolean),
+          );
```
radekmie, line 20: This will have to be simplified before merging.

radekmie (changes requested):
1. What if a field is called `[""]`? I know, it's a terrible name, but it's possible.
2. More generally, what if the field contains an arbitrary number of `[""]`? Like, `[""][""]`, `["[""]"]`, or `["[""]"]["[""]"]`?

PR #1098, opened 2022-03-18
"Looking for more" section

piotrpospiech:
Changed "looking for more" section. It's fully responsive.

Before:
Old layout

After:
New layout


website/index.module.css
```diff
@@ -702,6 +677,17 @@
   background-color: var(--periwinkle) !important;
 }
 
+.commercial-bullets > li {
+  font-weight: 400;
```
radekmie, line 681: How about `normal` instead?

radekmie (approved):

radekmie (approved):

PR #1102, opened 2022-04-01
New form builder section

zendranm:
This PR changes design of the `CommercialServices` section.

screen


radekmie (commented):
1. It'd be better to have less CSS (if possible).
2. I'd say we'd like to have a screenshot using a smaller resolution to make it look better on smaller screens.

wadamek65 (commented):
I ran this locally to take a look and I think some things may require tweaks:

1. I don't think the button styles are quite right. Unlike other buttons, when you hover over this one, you can see the native href underline. It's very subtle but it's there.
2. The clickable area is also much larger than the button itself. Not crucial but ideally should be adjusted as well.
3. The form builder screenshot seems to be a little cut-off from the right side (the horizontal padding is uneven). Could look better IMO.

radekmie (approved):

PR #1103, opened 2022-04-01
Implemented variable DateField type 'date' | 'datetime-local'. (closes #1099).

### LINKED ISSUE #1099: Add the option to select DateField type
Currently, there is only the option to select a date and time on the DateField component. 
https://github.com/vazco/uniforms/blob/master/packages/uniforms-material/src/DateField.tsx#L61

Maybe the `type` prop could be variable even if its default value is "datetime-local"

vcardins:
Allow the user to choose the date HTML5 input type, being either `date` or `datetime-local`.

`type=datetime-local`
image

`type=date`
image


packages/uniforms-material/src/DateField.tsx
```diff
@@ -3,9 +3,11 @@ import TextField, { TextFieldProps } from '@material-ui/core/TextField';
 import React from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
+export type AllowedDateTypes = 'date' | 'datetime-local';
```
radekmie, line 6: I wouldn't export this type as we don't do that for any of them and inline it instead.

packages/uniforms-material/__tests__/DateField.tsx
```diff
@@ -132,18 +132,32 @@ test('<DateField> - renders a Input with correct value (default)', () => {
   expect(wrapper.find(Input).prop('value')).toBe('');
 });
 
-test('<DateField> - renders a Input with correct value (model)', () => {
+const testDateType = (type: AllowedDateTypes = 'datetime-local') => {
```
radekmie, line 135: `test.each` would be better here.

radekmie (changes requested):
Could you also do the same for other themes and update docs accordingly? We'd rather keep them in sync when possible

packages/uniforms-material/__tests__/DateField.tsx
vcardins: @radekmie Updated the PR, please review again. (Sorry if I missed something, found a break at my work to do this)

packages/uniforms-bootstrap3/__tests__/DateField.tsx
```diff
@@ -43,13 +43,18 @@ test('<DateField> - renders an input with correct type', () => {
...
-  const element = <DateField name="x" disabled />;
-  const wrapper = mount(element, createContext({ x: { type: Date } }));
+const cases = ['datetime-local', 'date'] as DateFieldProps['type'][];
 
-  expect(wrapper.find('input')).toHaveLength(1);
-  expect(wrapper.find('input').prop('disabled')).toBe(true);
-});
+test.each(cases)(
+  '<DateField> - renders a Input with correct type : "%s"',
+  (type: DateFieldProps['type']) => {
```
radekmie, line 50: 1. This can be simplified with `as const`:
    ```ts
    test.each(['date', 'datetime-local'] as const)(
      '<DateField> - renders a Input with correct type : "%s"',
      type => {
    ```
2. Revert the removal of the `renders an input with correct disabled state` test?

docs/api-fields.md
```diff
@@ -162,6 +162,7 @@ import { BoolField } from 'uniforms-unstyled';
 |       `min`       |                                                                          Minimal value. Date object.                                                                           |                        All                         |
 | `showInlineError` |                                                          Field inline error. _Some description would be great, huh?_                                                           | antd, bootstrap3, bootstrap4, bootstrap5, semantic |
 |   `timeFormat`    |                                                           Display time picker in ampm (12hr) format or 24hr format.                                                            |                      material                      |
+|      `type`       |                                                 Display date and time or date only picker. Format value according to its type                                                  |                      material                      |
```
radekmie, line 165: It should say all themes except for `antd`.

radekmie (changes requested):
Some leftovers!

packages/uniforms-unstyled/__tests__/DateField.tsx
```diff
@@ -35,13 +35,16 @@ test('<DateField> - renders a input with correct name', () => {
   expect(wrapper.find('input').prop('name')).toBe('x');
 });
 
-test('<DateField> - renders an input with correct type', () => {
```
radekmie, line 38: This test is still removed (same goes for other themes).

radekmie (changes requested):
Other than that it's :+1:

wadamek65 (dismissed):

vcardins:
@radekmie Is there anything else I could do?

radekmie:
Thanks, @vcardins!

PR #1106, opened 2022-04-19
Implemented variable DateField type ('date' | 'datetime-local')

vcardins:
Before you open a pull request, please check if a similar one already exists or has been closed before. Detailed information about the process of contributing can be found in [CONTRIBUTING.md](https://github.com/vazco/uniforms/blob/master/.github/CONTRIBUTING.md).

Do link all relevant issues and pull requests. If there is none, please do file an issue first to discuss it upfront.


radekmie:
I guess this one is supposed to cover both #1103 and implement it for #1091, right? As #1103 is already merged, I'll close this one, and we'll implement it in #1091 separately.

CC @wadamek65.

PR #1111, opened 2022-04-29
Improved SelectField typings in uniforms-material

ErnestTeluk:
In this pull request, the correct type definitions have been added to the SelectField (material) component

### DIFF
--- a/packages/uniforms-material/__tests__/SelectField.tsx
+++ b/packages/uniforms-material/__tests__/SelectField.tsx
@@ -62,7 +62,7 @@ describe('@RTL - SelectField tests', () => {
     });
     const explicitProps = {
       fullWidth: false,
-      margin: 'dense' as const,
+      margin: 'normal' as const,
     };
 
     const { container } = render(
--- a/packages/uniforms-material/__tests__/SelectField.tsx
+++ b/packages/uniforms-material/__tests__/SelectField.tsx
@@ -62,13 +62,11 @@ describe('@RTL - SelectField tests', () => {
     });
     const explicitProps = {
       fullWidth: false,
-      margin: 'normal' as const,
+      margin: 'dense' as const,
     };
 
     const { container } = render(
       <ThemeProvider theme={theme}>
-        {/* When passing explicit props, TS errors regarding missing variant pop up */}
-        {/* @ts-expect-error */}
         <SelectField name="x" {...explicitProps} />
       </ThemeProvider>,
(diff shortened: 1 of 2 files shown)


packages/uniforms-material/__tests__/SelectField.tsx
```diff
@@ -62,13 +62,11 @@ describe('@RTL - SelectField tests', () => {
     });
     const explicitProps = {
       fullWidth: false,
-      margin: 'normal' as const,
+      margin: 'dense' as const,
```
wadamek65, line 65: I don't think that's a "fix" though? It makes the error go away but I believe `'normal'` should also be an acceptable prop value or is it not after all?

radekmie:
Please do improve the PR title and description. Also, tests are failing.

packages/uniforms-material/__tests__/SelectField.tsx
ErnestTeluk: yes, prop normal is normally acceptable, but mui has a hollow type definition in the `InputBase.d.ts` file, the definition says `margin? 'dense' | 'none'`. I fixed this by adding a prop margin in the interface of our SelectField

packages/uniforms-material/src/SelectField.tsx
```diff
@@ -170,15 +174,20 @@ function Select(props: SelectFieldProps) {
...
     'margin',
     'textFieldProps',
     'variant',
-  ]);
+  ] as const;
+  const filteredProps = omit(filterDOMProps(props), filteredOmittedItems);
+  const filteredPropsOmitted = filteredProps as Omit<
+    typeof filteredProps,
+    typeof filteredOmittedItems[number]
+  >;
```
radekmie, line 190: I've debugged it myself, and it looks like only the `checkboxes` and `disableItem` properties are a problem for the `omit` inference. Here's my approach, that reduces the fuss of type handling:
```diff
   const filteredProps = omit(filterDOMProps(props), [
-    'checkboxes',
-    'disableItem',
+    'checkboxes' as never,
+    'disableItem' as never,
     'fullWidth',
     'helperText',
     'margin',
     'textFieldProps',
     'variant',
   ]);
```

It works because we "hide" these two problematic properties from `omit` and let it do the trick. Let me know if you like it, as I think it's much cleaner. (Of course, it requires a solid comment _why_ it's there.)

radekmie (changes requested):

radekmie (approved):

PR #1117, opened 2022-05-27
Website SEO improvements

zendranm:
This PR introduces:
- In navigation "Enterprise" button replacing `to` with `href`. It doesn't trigger page reload and fixes the button color.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -70,7 +70,7 @@ module.exports = {
         },
         {
           label: 'Enterprise',
-          to: '/#commercial-services',
+          href: '/#commercial-services',
           position: 'left',
         },
       ],


docs/api-bridges.md
```diff
@@ -6,7 +6,7 @@ title: Bridges
 To make use of any schema, uniforms have to create a _bridge_ of it - a unified schema mapper.
 
 <p align="center">
-  <img src="/img/bridge-concept.svg" />
+  <img src="/img/bridge-concept.svg" alt=''/>
```
radekmie, line 9: This shouldn't pass Prettier. Can you check why it's not working here? If it's not able to check that, then at least revise it manually: double quotes, space before the tag end, and maximum line width.

website/components/Heading.tsx
```diff
@@ -7,8 +7,8 @@ export type HeadingProps = JSX.IntrinsicElements['h1'];
 
 export function Heading({ children, className, ...props }: HeadingProps) {
   return (
-    <h1 {...props} className={classNames(styles.heading, className)}>
+    <h2 {...props} className={classNames(styles.heading, className)}>
```
radekmie, line 10: Can you make sure whether the ordering of headings is correct with this change? I can't remember the tool that @wadamek65 recommended some time ago.

website/pages-parts/LandingPage/OpenSource.tsx
```diff
@@ -11,19 +11,17 @@ export function OpenSource() {
...
+      <h3
         className={classNames(
           styles.centered,
           styles.text,
           styles['heading-helper'],
         )}
       >
         Always Open Source.
-      </p>
+      </h3>
```
radekmie, line 22: Here's one example of potentially incorrect headings - `h3` preceedes `h2`.

radekmie (changes requested):

website/components/Heading.tsx
wadamek65: You can use this plugin: https://chrome.google.com/webstore/detail/headingsmap/flbjommegcjonpdmenkdiocclhjacmbi

zendranm: Thanks for the nice plugin. I checked it and there was only one incorrect ordering - the one from [this comment](https://github.com/vazco/uniforms/pull/1117#discussion_r884553295).

docs/motivation.md
```diff
@@ -10,7 +10,7 @@ Basically, there are a few types of forms with different capabilities.
 Most of the time you’ll be using either AutoForm or ValidatedForm, but there are quite a few more to choose from:
 
 <p align="center">
-  <img src="/img/uniforms-graph.svg" />
+  <img src="/img/uniforms-graph.svg" alt="Graph showing inheritance hierarchy of different forms." />
```
wadamek65, line 13: I don't think these alts are good. You should try imagining in a situation where you just cannot see this image at all and see if the alt gives you any useful information.

This as an example: Above this image you have a mention "There’s a very interesting class-based inheritance concept for forms." so we already know there is inheritance in uniforms.

Now we have an image with an alt that says "Graph showing inheritance hierarchy of different forms.". We already know there is an inheritance concept from the text above. Does this alt add any additional useful info? We get to know that a graph is supposed to be here, and it's supposed to be showing forms. But how does the graph look like or what kind of forms are present on it will never be known to the person reading this alt. This is why in my opinion this description is not useful and either should be rephrased to specify exactly what is contained within the graph or the alt should be empty.

docs/motivation.md
```diff
@@ -19,8 +19,7 @@ Most of the time you’ll be using either AutoForm or ValidatedForm, but there a
 > -- <cite>[Radosław Miernik](https://github.com/radekmie)</cite>
 
 <p align="center">
-  <img src="/img/hoc-hell.png" />
-  <small style={{display: 'block'}}>Screenshot of React DevTools when lots of HOC’s are applied to a component.</small>
+  <img src="/img/hoc-hell.png" alt="React DevTools when lots of HOC’s are applied to a component." />
```
wadamek65, line 22: I honestly do not know what to do with this. I'm not a big fan of this image/example in the first place as to me it is lacking some context describing what causes such HoC hell to happen. But to not overcomplicate things I think we can just let this stay, maybe mention how many HoCs are actually rendered for easier comparison.

docs/motivation.md
```diff
@@ -36,8 +35,7 @@ const ValidatedQuickForm = ValidatedForm.Validated(QuickForm.Quick(BaseForm));
 ```
 
 <p align="center">
-  <img src="/img/uniforms-traits.png" />
-  <small style={{display: 'block'}}>AutoValidatedQuickSemanticForm (using 5 "traits") viewed in React DevTools.</small>
+  <img src="/img/uniforms-traits.png" alt='AutoValidatedQuickSemanticForm (using 5 "traits") viewed in React DevTools.' />
```
wadamek65, line 38: Would be nice to mentioned that there is only one component rendered shown in the image.

website/pages/playground.tsx
```diff
@@ -5,7 +5,10 @@ import { Playground } from '../components/Playground';
 
 export default function PlaygroundPage() {
   return (
-    <Layout>
+    <Layout
+      title="Check how it works | uniforms"
+      description="Check how uniforms work, and enjoy the schema-first approach of one of the most popular React form libraries and enjoy building forms like never before."
```
wadamek65, line 10: This sentence sounds really strange `and enjoy` phrase is repeated twice and it's only once delimited by a comma. If we want it to stay I think it should look like this:

```suggestion
      description="Check how uniforms work, enjoy the schema-first approach of one of the most popular React form libraries, and enjoy building forms like never before."
```

But honestly, I think this should be rephrased.

wadamek65 (changes requested):

docs/motivation.md
zendranm: I consulted this with content creator and the new proposition is: "AutoForm class inheritance from ValidatedQuickForm which subsequently inherits from QuickForm and ValidatedForm"

docs/motivation.md
zendranm: After consultation: "Single AutoValidatedQuickSemanticForm component (using 5 "traits") viewed in React DevTools."

website/pages/playground.tsx
zendranm: After consultation: "Check how uniforms work, enjoy the schema-first approach of one of the most popular React form libraries, and build forms as easily as never before."

wadamek65 (approved):

radekmie (approved):

PR #1124, opened 2022-06-06
Configured PRs labeler

radekmie:
To make triaging issues easier, I configured an automatic PR labeler. In the future, we could mimic a similar behavior in the `.github/CODEOWNERS` file if needed, e.g., when we'll have dedicated people for a given theme.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,66 @@
+'Area: Bridge':
+  - 'packages/uniforms-bridge-graphql/**/*'
+  - 'packages/uniforms-bridge-json-schema/**/*'
+  - 'packages/uniforms-bridge-simple-schema-2/**/*'
+  - 'packages/uniforms-bridge-simple-schema/**/*'
+'Area: Core':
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,66 @@
...
+  - 'packages/uniforms-antd/**/*'
+'Theme: Bootstrap 3':
+  - 'packages/uniforms-bootstrap3/**/*'
+'Theme: Bootstrap 4':
+  - 'packages/uniforms-bootstrap4/**/*'
+'Theme: Bootstrap 5':
+  - 'packages/uniforms-bootstrap5/**/*'
+'Theme: MUI':
+  - 'packages/uniforms-material/**/*'
+'Theme: Material-UI':
```
wadamek65, line 61: Shouldn't this be reversed?

wadamek65 (changes requested):

.github/labeler.yml
radekmie: Definitely!

PR #1125, opened 2022-06-06
Fixed SimpleSchema types (fixes #1121).

### LINKED ISSUE #1121: Fix SimpleSchema types
Currently, whenever we use `SimpleSchema` we assert the schema type to `any` because otherwise, it will result in a type error saying that property `uniforms` should not be present.

A possible solution is noted in [this comment](https://github.com/vazco/uniforms/pull/1091#discussion_r861807416).

Example occurrence of such assertion: https://github.com/vazco/uniforms/blob/1112a282da87770aead7d8a6bad2a91da17d1dbf/reproductions/schema/all-fields-schema.tsx#L35

radekmie:
In this PR, I switched all but one test from using an ad-hoc `SimpleSchema` mocks to using actual schemas. This required typing the `uniforms` property in `SchemaDefinition` as well as some further cleanup.

The only problem that is still there, is the fact that while `meteor/aldeed:simple-schema` is _technically_ typed now (LSP sees it), it doesn't work after in all environments -- either build or lint fails. As build is more important, I chose to `@ts-ignore` it for now (`@ts-expect-error` makes lint fail).


packages/uniforms-bridge-simple-schema/__tests__/SimpleSchemaBridge.ts
```diff
@@ -80,9 +82,9 @@ describe('SimpleSchemaBridge', () => {
         throw 'ValidationError';
       };
     },
-  };
+  } as unknown as SimpleSchema;
```
wadamek65, line 85: Why do we need an assertion here?

wadamek65 (approved):

packages/uniforms-bridge-simple-schema/__tests__/SimpleSchemaBridge.ts
radekmie: Because a real SimpleSchema instance has many more fields.

PR #1140, opened 2022-06-29
Unified implementation of DateField type handling.

### LINKED ISSUE #1138: Date Field not updating visually
When a field is configured to be a date only field (not date and time) the field value gets updated but the form field does not update. The form field continue to shows the placeholder. I tried to debug the issue but couldn't get to the core issue. It seems like the field temporarily shows the selected value then gets rerendered to show the placeholder.

Let us know what is happening. Do provide us with:

- Latest version of uniforms with all style frameworks. This issue happens in Firefox and Chrome.
- Setup Uniforms to use the following configuration:
```
new SimpleSchema2Bridge(
  new SimpleSchema({
   "date":{
      "type": Date,
      "defaultValue":new Date(),
      "uniforms":{
         "type":"date"
      }
   },
   "adult": Boolean,
   "size":{
      "type":"String",
      "defaultValue":"m",
      "allowedValues":[
         "xs",
         "s",
         "m",
         "l",
         "xl"
      ]
   },
   "rating":{
      "type":Number,
      "allowedValues":[
         1,
         2,
         3,
         4,
         5
      ],
      "uniforms":{
         "checkboxes":true
      }
   }
  })
)
```
- [Playground Link](https://uniforms.tools/playground/#?N4IgDgTgpgzlAuIBcIDqUA2BjA9gWygEIQAacCHMGZUAQwFd4cZaA3KZAM1ozjIaYt2AEUy0AnsgCMABhlkAJgEsWAIwxQFXHnxAZaqzMngR6UMmH1YoACxwYFUCNt7mQ0WgoDyAOwySkblcyGCwbKDxaZBAfKAB3AAIAZSU8SygksIjaACYAIQglBQBzKAAKAB0fBITYxJS0jUzwyLLgKpqEipAFWngobqR26s6a7vhxMAHkBOE-8w7Rrp6obnoMeAA1HjNButn5soBKEkXR7vofJU4cCDwYQeGl85AJqcHu3v7us86AXzOf1OI2WnnW8EGCTyOHsUFoPmBnW6MCUAC9pkNfmNXpMMd0kiYlD5it1EUtPqsGBtthhdshunhSVjQRgMDg4poaWYHsgANrMpEgAAePLJz2WooF2MZpClywwTJBz26QoVIGZAF1AWLuhA-kSSTRmeNcYMAHL0PCGCBiwU8NkchRc2CDflKpZSW1LHJe0YAZl9nQALIGagBWTW-i5XG53HmY90vLJYADWqhwQpdxlMA0TAJB-ZqfyOVSOpBAMDscQAkn4iVAAKIQCjOQI6Ny0GDiHxYXxJeiqPBKRBIExmfhdnu-GlFeYuOBA14tDgoPD0JQgP5AA)
image

wadamek65:
Fixes #1138

Unified the implementation of date type logic across `DateField`s in themes. The only theme that was not adjusted was `antd` due to the fact that this is handled by a separate prop which is `showTime` that is already implemented. `antd` does not accept a `type` prop like the native date input does.

to do:
* [x] Copy the date test to all other themes
* [x] Test manually once more

<details>
<summary>Screenshots</summary>

tested for this playground schema

```
new SimpleSchema2Bridge(
  new SimpleSchema({
      date: {
    type: Date,
    defaultValue: new Date(),
    uniforms: {
      type: 'date',
    },
  },
  date2: {
    type: Date,
    defaultValue: new Date(),
  },
  })
)
```

unstyled

image

semantic

image

mui

image

material

image

bootstrap5

image

bootstrap4

image

bootstrap3

image

</details>

### DIFF
--- a/packages/uniforms-bootstrap3/src/DateField.tsx
+++ b/packages/uniforms-bootstrap3/src/DateField.tsx
@@ -38,11 +38,9 @@ function Date({
   placeholder,
   readOnly,
   value,
-  type,
+  type = 'datetime-local',
   ...props
 }: DateFieldProps) {
-  const dateType = type === 'date' ? type : 'datetime-local';
-
   return wrapField(
     { ...props, id },
     <input
@@ -65,7 +63,7 @@ function Date({
       placeholder={placeholder}
       readOnly={readOnly}
       ref={inputRef}
-      type={dateType}
+      type={type}
       value={dateFormat(value, type) ?? ''}
     />,
   );
--- a/packages/uniforms-bootstrap3/src/DateField.tsx
+++ b/packages/uniforms-bootstrap3/src/DateField.tsx
@@ -4,9 +4,12 @@ import { connectField, HTMLFieldProps } from 'uniforms';
 
 import wrapField from './wrapField';
 
+type DateFieldType = 'date' | 'datetime-local';
+
 /* istanbul ignore next */
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
-const dateFormat = (value?: Date) => value?.toISOString().slice(0, -8);
+const dateFormat = (value?: Date, type: DateFieldType = 'datetime-local') =>
+  value?.toISOString().slice(0, type === 'datetime-local' ? -8 : -14);
 
 export type DateFieldProps = HTMLFieldProps<
   Date,
@@ -18,7 +21,7 @@ export type DateFieldProps = HTMLFieldProps<
     max?: Date;
     min?: Date;
     wrapClassName?: string;
-    type?: 'date' | 'datetime-local';
+    type?: DateFieldType;
   }
 >;
 
@@ -63,7 +66,7 @@ function Date({
       readOnly={readOnly}
       ref={inputRef}
       type={dateType}
-      value={dateFormat(value) ?? ''}
+      value={dateFormat(value, type) ?? ''}
     />,
   );
 }
--- a/packages/uniforms-mui/src/DateField.tsx
+++ b/packages/uniforms-mui/src/DateField.tsx
@@ -2,8 +2,6 @@ import TextField, { TextFieldProps } from '@mui/material/TextField';
 import React from 'react';
 import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 
-type DateFieldType = 'date' | 'datetime-local';
-
 /* istanbul ignore next */
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
 
@@ -19,7 +17,7 @@ const dateParse = (timestamp: number, onChange: DateFieldProps['onChange']) => {
 export type DateFieldProps = FieldProps<
   Date,
   TextFieldProps,
-  { labelProps?: object; type?: DateFieldType }
+  { labelProps?: object; type?: 'date' | 'datetime-local' }
 >;
 
 function Date({
--- a/packages/uniforms-mui/src/DateField.tsx
+++ b/packages/uniforms-mui/src/DateField.tsx
@@ -39,11 +39,9 @@ function Date({
   readOnly,
   showInlineError,
   value,
-  type,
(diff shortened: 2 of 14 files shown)


packages/uniforms-mui/src/DateField.tsx
```diff
@@ -35,8 +39,11 @@ function Date({
   readOnly,
   showInlineError,
   value,
+  type,
   ...props
 }: DateFieldProps) {
+  const dateType = type === 'date' ? type : 'datetime-local';
```
wadamek65, line 45: This was included in the initial implementation so I added it here for consistency but I'm not sure we need this logic. Thoughts @radekmie ?

packages/uniforms-bootstrap3/src/DateField.tsx
```diff
@@ -63,7 +66,7 @@ function Date({
       readOnly={readOnly}
       ref={inputRef}
       type={dateType}
-      value={dateFormat(value) ?? ''}
+      value={dateFormat(value, type) ?? ''}
```
radekmie, line 69: This could be simplified to:
```ts
      type={type === 'date' ? type : 'datetime-local'}
      value={value?.toISOString().slice(0, type === 'date' ? -14 : -8) ?? ''}
```

Then we signalize that it's the same logic. If needed, we can have a constant for `type === 'date'` (but I don't think we do).

Of course, this applies to all themes.

radekmie (commented):
It'd be nice to have tests for that too.

packages/uniforms-bootstrap3/src/DateField.tsx
wadamek65: Do you think this check makes sense if we specify the "type" type as `'date' | 'datetime-local'`?

```ts
type === 'date' ? type : 'datetime-local'
```

radekmie: Well, I don't see a different option...? Or maybe I misunderstood something.

wadamek65: If `type` is typed as `'date' | 'datetime-local'` then we can just pass `type` directly instead of doing it via the ternary operator:

```diff
- type={type === 'date' ? type : 'datetime-local'}
+ type={type}
```

radekmie: Answering this and the one above: I'm not entirely sure, but I have a weird feeling that it's there for a reason... If you'll check it (locally) that the `type` never has an odd value (like, `String` for example), then I'm fine with that.

packages/uniforms-bootstrap3/src/DateField.tsx
```diff
@@ -62,8 +63,8 @@ function Date({
       placeholder={placeholder}
       readOnly={readOnly}
       ref={inputRef}
-      type={dateType}
-      value={dateFormat(value) ?? ''}
+      type={type}
+      value={dateFormat(value, type) ?? ''}
```
radekmie, line 67: If we'd inline `dateFormat`, then we could inline `DateFieldType` too, making the `DateField` much simpler.

radekmie (commented):
The test looks good :+1:

radekmie (approved):
The code looks good. @kestarumper could you check the PR description and mark this PR as ready once it's done?

kestarumper:
Not sure if this is a regression of this PR, but the _clear_ button is not working when selecting the date with the picker instead of entering by the keyboard.

Other than that, the types `date` and `datetime-local` work correctly.

image

radekmie:
@kestarumper, @wadamek65 Can you check this "clear" button? If it didn't work before, then I guess it's fine.

wadamek65:
To do:
* Clearing the date input for native browser picker is not a regression. It may be worth creating a new issue to address this in the future or see exactly where it stems from.
* Check if this PR resolves: #1162

PR #1141, opened 2022-07-01
JSONSchemaBridge: Get errors for fields with dots

### LINKED ISSUE #1131: Fields with dots in their names not receiving errors
## Problem description
The issue happens with
- `"uniforms": "3.8.1",`
- `"uniforms-bridge-json-schema": "3.8.1"`

Individual field errors are not passed to the fields. This causes them to not display the inline error message nor highlight it [(in some cases it even displays the field as "green" indicating that everything is OK)](https://user-images.githubusercontent.com/13087338/173058145-8fece2c3-5e22-42df-a139-958d07056c27.png). The issue happens in `JSONSchemaBridge` with the use of `Ajv` as a validator. **Note**: The schema validations are still performed (all invalid ones are found) and any errors are displayed by `<ErrorsField />`.

## Technical observation
The field names (field paths, instancePath) produced by Ajv are not escaped the way uniforms do this. For example, `Ajv` will set the `error.instancePath` as `"/nested.field"`. This causes problems in `JSONSchemaBridge.getError` because the escaped field name and `instancePath` don't match.

## Reproduction
- [Codesandbox](https://codesandbox.io/embed/sweet-jasper-tklydk?fontsize=14&hidenavigation=1&theme=dark)

<details>
<summary>Screenshot</summary>

image
</details>

<details>
<summary>Example JSONSchema</summary>

```js
{
  type: "object",
  properties: {
    regularRequired: {
      type: "string"
    },
    "nested.required": {
      type: "string"
    },

    regularMinLength: {
      type: "string",
      minLength: 5,
      default: "koń",
    },
    "nested.minLength": {
      type: "string",
      minLength: 5,
      default: "koń",
    }
  },
  required: ["nested.required", "regularRequired"]
}
```
</details>

<details>
<summary>Half-solution (semi-workaround)</summary>

```ts
import get from 'lodash/get';
import { joinName } from 'uniforms';
import JSONSchemaBridge from 'uniforms-bridge-json-schema';

function pathToName(path: string) {
  path = path.startsWith('/')
    ? path.replace(/\//g, '.').replace(/~0/g, '~').replace(/~1/g, '/')
    : path
        .replace(/\[('|")(.+?)\1\]/g, '.$2')
        .replace(/\[(.+?)\]/g, '.$1')
        .replace(/\\'/g, "'");

  return path.slice(1);
}

export class CustomJSONSchemaBridge extends JSONSchemaBridge {
  getError(name: string, error: any) {
    const details = error?.details;
    if (!Array.isArray(details)) {
      return null;
    }

    const rootName = joinName(nameParts.slice(0, -1));
    const _baseName = nameParts[nameParts.length - 1];
    const baseName = _baseName.includes('.')
      ? pathToName(_baseName)
      : _baseName;

    function findScopedError(error: any) {
      const path = pathToName(error.instancePath ?? error.dataPath);
      const _name = name.includes('.') ? pathToName(name) : name;
      const nameMatchesPath = _name === path;
      const rootMatchesPath = rootName === path;
      const baseNameMatchesMissingProperty =
        baseName === error.params.missingProperty;

      const childErrors = error.params.errors;

      const matches =
        nameMatchesPath || (rootMatchesPath && baseNameMatchesMissingProperty);

      return matches || !!childErrors?.some(findScopedError);
    }

    const scopedError = details.find(findScopedError);

    return scopedError || null;
  }
}
```
</details>

kestarumper:
This PR closes #1131 by extending behavior of `JSONSchemaBridge.getError` which now correctly returns errors for the fields with dots in their names.

The problem was that `Ajv` provides the `error.instancePath` in a "non-escaped way" for field names with dots.
```js
// Example schema
{
  type: 'object',
  properties: {
    'nested.minLength': {
      type: 'string',
      minLength: 5,
      default: 'Horse',
    },
}
```
Uniforms would escape this field name into `["nested.minLength"]` but the `error.instancePath` would be `/nested.minLength`.

The fix uses `joinName.unescape` to unescape individual parts of the `name` and compare it to the `path` created from `error.instancePath`.

### DIFF
--- a/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
@@ -388,6 +388,22 @@ describe('JSONSchemaBridge', () => {
         expect(bridge.getError(name, error)).toEqual(error.details[0]);
       });
     });
+
+    it('works with correct error (complex instance paths - dots in names)', () => {
+      const pairs = [
+        ['["path.with.a.dot"]', '/path.with.a.dot'],
+        ['["path.with.a.dot"].another', '/path.with.a.dot/another'],
+        [
+          '["path.with.a.dot"].["another.with.a.dot"]',
+          '/path.with.a.dot/another.with.a.dot',
+        ],
+      ];
+
+      pairs.forEach(([name, instancePath]) => {
+        const error = { details: [{ instancePath }] };
+        expect(bridge.getError(name, error)).toEqual(error.details[0]);
+      });
+    });
   });
 
   describe('#getErrorMessage', () => {
--- a/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
@@ -391,12 +391,13 @@ describe('JSONSchemaBridge', () => {
 
     it('works with correct error (complex instance paths - dots in names)', () => {
       const pairs = [
-        ['["path.with.a.dot"]', '/path.with.a.dot'],
-        ['["path.with.a.dot"].another', '/path.with.a.dot/another'],
-        [
-          '["path.with.a.dot"].["another.with.a.dot"]',
-          '/path.with.a.dot/another.with.a.dot',
-        ],
+        ['["a.b.c"].["d.e/f"].g', '/a.b.c/d.e~1f/g'],
+        ['a.["b"]', '/a/b'],
+        ['a.["b"].c', '/a/b/c'],
+        ['a.["b.c"]', '/a/b.c'],
+        ['a.["b.c"].d', '/a/b.c/d'],
+        ['["a"].b', '/a/b'],
+        ['["a.b"].c', '/a.b/c'],
       ];
 
       pairs.forEach(([name, instancePath]) => {
--- a/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
+++ b/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
@@ -103,7 +103,7 @@ export default class JSONSchemaBridge extends Bridge {
     }
 
     const nameParts = joinName(null, name).map(joinName.unescape);
-    const unescapedName = joinName(...nameParts);
+    const unescapedName = joinName(nameParts);
     const rootName = joinName(nameParts.slice(0, -1));
     const baseName = nameParts[nameParts.length - 1];
     const scopedError = details.find(error => {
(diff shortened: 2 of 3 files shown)


wadamek65 (approved):

reproductions/schema/json-schema.tsx
```diff
@@ -15,34 +15,64 @@ function createValidator(schema: object) {
 const schema = {
   type: 'object',
   properties: {
-    a: {
-      type: 'number',
-      title: 'a',
-      uniforms: { title: 'Horse' },
+    regularRequired: {
```
radekmie, line 18: I know that the current example schema is not great, but I don't think that this one is any better.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -102,13 +102,14 @@ export default class JSONSchemaBridge extends Bridge {
       return null;
     }
 
-    const nameParts = joinName(null, name);
+    const nameParts = joinName(null, name).map(joinName.unescape);
+    const unescapedName = joinName(...nameParts);
```
radekmie, line 106: `joinName` can handle arrays:
```suggestion
    const unescapedName = joinName(nameParts);
```

packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
```diff
@@ -388,6 +388,22 @@ describe('JSONSchemaBridge', () => {
...
+
+    it('works with correct error (complex instance paths - dots in names)', () => {
+      const pairs = [
+        ['["path.with.a.dot"]', '/path.with.a.dot'],
+        ['["path.with.a.dot"].another', '/path.with.a.dot/another'],
+        [
+          '["path.with.a.dot"].["another.with.a.dot"]',
+          '/path.with.a.dot/another.with.a.dot',
+        ],
+      ];
```
radekmie, line 400: Let's add a couple more tests:
```
a.["b"]
a.["b"].c
a.["b.c"]
a.["b.c"].d
["a"].b
["a.b"].c
```

Also, how does it work with `/` in the name?

radekmie (changes requested):

reproductions/schema/json-schema.tsx
kestarumper: Yeah, I was using it to test visually if the validation works and the fields get highlighted. It just covers some "with dot names" fields.

Should I come up with some better schema or just revert it to the previous one?

packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts
kestarumper: > Also, how does it work with `/` in the name?

I believe it's covered  by the other tests (like [this one](https://github.com/vazco/uniforms/blob/76181387fc065ceb153bc6dfc4c9cb61eb27e836/packages/uniforms-bridge-json-schema/__tests__/JSONSchemaBridge.ts#L366))

reproductions/schema/json-schema.tsx
radekmie: A better schema would be best. Maybe we should copy the one from the playground?

radekmie (approved):

PR #1143, opened 2022-07-08
Improved documentation for useForm error.

wadamek65:
Extended the current invariant message and added an entry to the FAQ.

* [x] `useForm` hook error documentation
* [x] [bluebird promise error documentation](https://github.com/vazco/uniforms/discussions/1047#discussioncomment-2891337)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -294,3 +294,14 @@ const anyProps: any = {
 }
 <AutoForm {...anyProps} />
 ```
+
+### Why am I getting "useForm must be used within a form" error?
+
+`uniforms` uses a `React.Context` in order to keep the state of the whole form.
+The provider for this context is created by `BaseForm`, and in turn all the other Form components inheriting it.
+
+There are two most common issues causing this problem:
+
+1. The component calling this function does not have a Form component above it anywhere in the component tree. To fix
+   this, wrap this component within a parent Form component (does not have to be direct).
+2. There are multiple versions of `uniforms` installed in your `node_modules`. This usually happens when you have more than one version of the core `uniforms` package installed which can happen when you have a mismatch of versions between any of your `uniforms` related dependencies. Ensure all your versions are matching, clean any `node_modules` directories and reinstall dependencies to resolve this error.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -295,13 +295,21 @@ const anyProps: any = {
 <AutoForm {...anyProps} />
 ```
 
-### Why am I getting "useForm must be used within a form" error?
+### "`useForm` must be used within a form"
 
-`uniforms` uses a `React.Context` in order to keep the state of the whole form.
-The provider for this context is created by `BaseForm`, and in turn all the other Form components inheriting it.
+uniforms uses a `React.Context` in order to keep the state of the whole form.
+The provider for this context is rendered by `BaseForm`, and in turn all the other form components inheriting it.
 
 There are two most common issues causing this problem:
 
 1. The component calling this function does not have a Form component above it anywhere in the component tree. To fix
+
    this, wrap this component within a parent Form component (does not have to be direct).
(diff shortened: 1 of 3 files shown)


wadamek65:
@radekmie I also wanted to extend the documentation regarding bluebird as mentioned [here](https://github.com/vazco/uniforms/discussions/1047#discussioncomment-2891337) but I'm not sure what exactly there is to extend. If you can give me some pointers I'll add them to this PR.

docs/faq.md
```diff
@@ -294,3 +294,14 @@ const anyProps: any = {
...
+### Why am I getting "useForm must be used within a form" error?
+
+`uniforms` uses a `React.Context` in order to keep the state of the whole form.
+The provider for this context is created by `BaseForm`, and in turn all the other Form components inheriting it.
+
+There are two most common issues causing this problem:
+
+1. The component calling this function does not have a Form component above it anywhere in the component tree. To fix
+   this, wrap this component within a parent Form component (does not have to be direct).
+2. There are multiple versions of `uniforms` installed in your `node_modules`. This usually happens when you have more than one version of the core `uniforms` package installed which can happen when you have a mismatch of versions between any of your `uniforms` related dependencies. Ensure all your versions are matching, clean any `node_modules` directories and reinstall dependencies to resolve this error.
```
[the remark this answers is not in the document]
radekmie: We do so, as we can either do it every time or not do that at all **automatically**. Doing it manually would require an extra step while publishing, plus I think it's not possible in Lerna easily.

radekmie, line 307: I'm a little confused about the formatting:
1. Why is `uniforms` in backticks? If you wanted to highlight that _this package_ does that, then let's say "The `uniforms` package [...]", as right now we do use the name uniforms as the entire solution. (It's much better in the last sentence.)
2. "The provider for this context is ~~created~~ rendered by [...]"
3. "[...] other ~~Form~~ form components [...]" (Same applies to the next lines.)
4. Why is 1. split into two lines but 2. is not? I'd rather do 2., but I'm not sure what Prettier thinks about that.

packages/uniforms/src/useForm.ts
```diff
@@ -6,6 +6,16 @@ import { Context } from './types';
...
-  invariant(context !== null, 'useForm must be used within a form.');
+  invariant(
+    context !== null,
+    `useForm must be used within a form.
+
+Two most common reasons for this error are:
+1. Component calling this function doesn't have a parent Form component in the tree.
+2. A duplicate uniforms dependency is installed in node_modules.
+
+For more info check FAQ: https://uniforms.tools/docs/faq/#why-am-i-getting-useForm-must-be-used-within-a-form-error
```
radekmie, line 17: I think that the title could be better :sweat_smile:

radekmie: Maybe "`useForm` must be used within a form" (in quotes) would be enough?

kestarumper (approved):

docs/faq.md
```diff
@@ -294,3 +294,22 @@ const anyProps: any = {
...
+
+1. The component calling this function does not have a Form component above it anywhere in the component tree. To fix
+
+   this, wrap this component within a parent Form component (does not have to be direct).
+
+2. There are multiple versions of `uniforms` installed in your `node_modules`. This usually happens when you have
+
+   more than one version of the core `uniforms` package installed which can happen when you have a mismatch of
+
+   versions between any of your `uniforms` related dependencies. Ensure all your versions are matching,
```
Monteth, line 313: Maybe it would be helpful to specify what we mean by saying `uniforms related dependencies`, because we specifically mean bridges and themes, right?

Monteth (approved):

radekmie (approved):

docs/faq.md
radekmie: This can be done in a following PR.

PR #1147, opened 2022-07-15
Support GraphQL Enums and show them as dropdown (continuation)

kestarumper:
This PR continues the work of #1135 (thanks @simonhoss) by adding priority to the `options` and using enum values only as a last resort.


kestarumper:
Closing this because I managed to push directly to the original branch.

PR #1153, opened 2022-08-02
Updated testimonials

nexxtGen:
In this task, the testimonials have been changed.

In website folder:
- New OvalLink component added
- Photo / avatar for testimonial
- CSS Style changes
- I'm using dangerouslySetInnerHTML for description props testimonial because we want a link in the description.

1_12em


website/components/OvalLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+import styles from '../index.module.css';
+
+export type OvalLinkProps = JSX.IntrinsicElements['span'] & { to: string };
+
+export function OvalLink({ children, className, to, ...props }: OvalLinkProps) {
+  return (
+    <Link className={styles['oval-link']} to={to}>
+      <span
+        {...props}
+        className={classNames(styles['oval-link-icon'], className)}
```
radekmie, line 14: This class suggests that it's always going to be an icon. If so, I'd rather have an `IconLink` component.

Also, is it really that different from the others we have? Maybe we could _slightly_ extend an existing one instead?

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -10,6 +11,9 @@ export type TestimonialProps = {
   mirror?: boolean;
   position: string;
   who: string;
+  avatar: string;
+  linkGithub: string;
+  linkLinkedin: string;
```
radekmie, line 16: Let's keep these ordered.

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -39,17 +46,33 @@ export function Testimonial({
...
+              className={styles['small-image-icon']}
+            />
+          </OvalLink>
+        </div>
         <p
           className={classNames(styles.text, styles['testimonial-description'])}
-        >
-          {description}
-        </p>
+          dangerouslySetInnerHTML={{ __html: description }}
```
radekmie, line 74: Instead of allowing arbitrary HTML here, make the `description` field a `ReactElement`.

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -70,17 +93,23 @@ export function Testimonials() {
...
-          description="uniforms it’s one and the only library that allows you to have greater flexibility on top of the React platform to building forms you like, with great robustness with advanced support of many different forms inputs."
-          position="Team Lead"
-          who="Wojtek Trocki"
+          company="Resolve"
+          description="Uniforms is the backbone of our data-intensive web-applications. We have about 200 different forms, from very simple ones, to ones that are filled with complex data-loading conditional form components, which create an incredible UX for our users. And if you really need to push the limits of what you can do with forms, I would highly recommend <b><a href='https://www.vazco.eu/' alt='Vazco website' target='blank'>reaching out to Vazco</a></b> themselves for expert advice."
+          position="CTO and Co-Founder at Resolve"
+          who="Florian Bienefeldt"
+          avatar="img/florian-carre.jpg"
+          linkGithub="https://github.com/Floriferous"
+          linkLinkedin="https://ch.linkedin.com/in/florianbienefelt"
```
radekmie, line 102: 1. The image name doesn't match the name.
2. The name doesn't match the LinkedIn link.
3. Let's move all avatars into `img/avatar/{GitHub handle}` instead.
4. It'd be better to have all avatars in the same format (`.jpg` is fine).

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -39,17 +46,33 @@ export function Testimonial({
...
+              className={styles['small-image-icon']}
+            />
+          </OvalLink>
+          <OvalLink to={linkLinkedin}>
+            <img
+              alt="github"
+              src="assets/linkedin.svg"
+              className={styles['small-image-icon']}
+            />
+          </OvalLink>
```
radekmie, line 70: 1. It should say "GitHub", not "Github"; same for "LinkedIn", not "Linkedin".
2. Instead of using custom icons here, use the `react-feather` package that we use elsewhere.
3. It actually shows that the `OvalLink` component could be a `IconLink` instead.

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -70,17 +93,23 @@ export function Testimonials() {
       <Heading>Testimonials</Heading>
       <div className={styles.testimonials}>
         <Testimonial
-          company="Red Hat"
-          description="uniforms it’s one and the only library that allows you to have greater flexibility on top of the React platform to building forms you like, with great robustness with advanced support of many different forms inputs."
-          position="Team Lead"
-          who="Wojtek Trocki"
+          company="Resolve"
+          description="Uniforms is the backbone of our data-intensive web-applications. We have about 200 different forms, from very simple ones, to ones that are filled with complex data-loading conditional form components, which create an incredible UX for our users. And if you really need to push the limits of what you can do with forms, I would highly recommend <b><a href='https://www.vazco.eu/' alt='Vazco website' target='blank'>reaching out to Vazco</a></b> themselves for expert advice."
```
radekmie, line 97:
```suggestion
          description="uniforms is the backbone of our data-intensive web-applications. We have about 200 different forms, from very simple ones, to ones that are filled with complex data-loading conditional form components, which create an incredible UX for our users. And if you really need to push the limits of what you can do with forms, I would highly recommend <b><a href='https://www.vazco.eu/' alt='Vazco website' target='blank'>reaching out to Vazco</a></b> themselves for expert advice."
```

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -70,17 +93,23 @@ export function Testimonials() {
...
+          who="Florian Bienefeldt"
+          avatar="img/florian-carre.jpg"
+          linkGithub="https://github.com/Floriferous"
+          linkLinkedin="https://ch.linkedin.com/in/florianbienefelt"
         />
         <Testimonial
-          company="Paystack"
-          description="The simplicity with which one can dive into details and break a form apart into it's primitive provides the much-needed peace of mind."
+          company="Toptal"
+          description="Uniforms is my go-to solution for quite a while. Great holistic approach to tackle forms. I especially love the approach to making custom form layouts. Developer experience par-excellence"
```
radekmie, line 106:
```suggestion
          description="uniforms is my go-to solution for quite a while. Great holistic approach to tackle forms. I especially love the approach to making custom form layouts. Developer experience par-excellence"
```

radekmie (changes requested):
Sorry for the delay!

website/pages-parts/LandingPage/Testimonials.tsx
nexxtGen: Good idea to get rid of dangerouslySetInnerHTML. Thanks!

website/pages-parts/LandingPage/Testimonials.tsx
nexxtGen: Details matter. It is now more systematic.

website/components/OvalLink.tsx
nexxtGen: I stayed with a separate component. But that's right, we could use the oval component, expand it with a props link.

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -18,6 +23,9 @@ export function Testimonial({
   mirror,
   position,
   who,
+  avatar,
+  linkGithub,
+  linkLinkedin,
```
radekmie, line 28: (minor) Let's keep these ordered as well.

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -39,12 +47,17 @@ export function Testimonial({
...
-          <b>{company}</b>
+          <b>{who}</b>
+          <br />
+          {position} at <b>{company}</b>
           <br />
-          {position}
         </p>
+        <div className={styles['testimonial-links-wrapper']}>
+          <IconLink color="#412a94" icon={GitHub} to={linkGithub} />
+          <IconLink color="#412a94" icon={Linkedin} to={linkLinkedin} />
```
radekmie, line 59: I'd rather keep the color in the CSS. _(It shouldn't be a problem, as `react-feather` provides SVG icons with `currentColor` color by default, meaning that we can style them using the `color` property in CSS.)_

website/pages-parts/LandingPage/Testimonials.tsx
```diff
@@ -70,17 +83,44 @@ export function Testimonials() {
...
         />
         <Testimonial
-          company="Paystack"
-          description="The simplicity with which one can dive into details and break a form apart into it's primitive provides the much-needed peace of mind."
+          company="Toptal"
+          description={
+            <>
+              uniforms is my go-to solution for quite a while. Great holistic
+              approach to tackle forms. I especially love the approach to making
+              custom form layouts. Developer experience par-excellence
```
radekmie, line 115:
```suggestion
              custom form layouts. Developer experience par-excellence.
```

radekmie (changes requested):
Last minor tweaks.

website/pages-parts/LandingPage/Testimonials.tsx
nexxtGen: My suggestion was the Badge component. 
In my opinion, we should keep the option (as an option) of giving an icon color using the props "color". In the current case I added our color --dusk-blue for testimonial icons using className. I know I could do this by specifying the color of the testimonial icons wrapper. However, it would not fit for me with the concept of reusability of the <IconLink /> component which is a separate organism with its properties. However, if I had to change anything here, write me.

radekmie: I simply don't see a need for that _now_, and there's no point in making the implementation more complex only because we _may_ need it. If we will, we'll do it in the same way - with CSS.

nexxtGen: You're right. Giving colors using CSS is more beneficial for us, because we have a predefined color palette there. I will remove the ability to color icons in IconLink using props color. leave the rest as is?

radekmie: Let's remove this `color` prop and move the `color` from `.testimonial-icon-color` to `.testimonial-links-wrapper` - it should work in the same way but with less code.

nexxtGen: Applying a color to a container (.testimonial-links-wrapper) doesn't work because the <a /> tag has default styling in the application. !important does not resolve this problem. I suggest leaving the props className passed in IconLink in testimonial (.testimonial-icon-color).

nexxtGen: Edit: i remove this className prop and add style by ```.testimonial-links-wrapper a {
  color: ...;
}```

radekmie (approved):

PR #1163, opened 2022-09-14
Make AutoFields render a Fragment wrapper

### LINKED ISSUE #1133: AutoField API Proposal
Hi,
I like the idea of changing fields components by context. Could we push it little further and do this?

```js
import { AutoForm } from 'uniforms'
import { CustomAuto } from './custom-autofield-as-described-in-docs'

<AutoForm.AutoFieldsOverrider autoField={CustomAuto} />
  <AutoForm ... />
</AutoForm.AutoFieldsOverrider >
```

```js
const AutoForm.AutoFieldsOverrider = ({autoField, ...props}) => <AutoForm.AutoFieldsContext.Provider value={autoField} {...props} />
```

And `AutoForm` and `AutoFields` would then check the `AutoForm.AutoFieldsContext` to see which AutoField should be rendered.

Also `AiutoFields` is currently tied to styled package. But it is not neccessary since it is basically only a loop. Or whatever it is, it is not style-related.

wadamek65:
Closes #1133


website/sidebars.json
```diff
@@ -1,7 +1,12 @@
 {
   "docs": {
     "Introduction": ["what-are-uniforms", "motivation", "compare-matrix"],
-    "Getting Started": ["installation", "faq", "migrating-2-to-3"],
+    "Getting Started": [
+      "installation",
+      "faq",
+      "migrating-3-to-4",
+      "migrating-2-to-3"
```
radekmie, line 8: Do you think that it's better to order them like this? Both make sense, actually...

radekmie (approved):

website/sidebars.json
wadamek65: I chose this order because it puts the v3 to v4 tab at the top, closer to the beginning which may be noticed first if someone was looking for it from top to bottom.

PR #1164, opened 2022-09-15
Getinitialvalue props

wadamek65:
Removed support for `initialCount` property on list fields and bridges. `getInitialValue` now doesn't require props being passed to it so it is now memoized. It has also been removed from `filterDomProps`.

### DIFF
--- a/packages/uniforms-antd/__tests__/ListField.tsx
+++ b/packages/uniforms-antd/__tests__/ListField.tsx
@@ -59,23 +59,25 @@ test('<ListField> - renders correct label (specified)', () => {
   );
 });
 
-test('<ListField> - renders correct numer of items with initialCount (specified)', () => {
-  const element = <ListField name="x" initialCount={3} />;
+test('<ListField> - renders correct numer of items with model (specified)', () => {
+  const element = <ListField name="x" />;
   const wrapper = mount(
     element,
-    createContext({ x: { type: Array }, 'x.$': { type: String } }),
+    createContext({ x: { type: Array }, 'x.$': { type: String } }, undefined, {
+      x: [undefined, undefined, undefined],
+    }),
   );
 
   expect(wrapper.find('input')).toHaveLength(3);
 });
 
 test('<ListField> - passes itemProps to its children', () => {
-  const element = (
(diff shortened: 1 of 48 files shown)


packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -94,6 +94,7 @@ export default class JSONSchemaBridge extends Bridge {
     this.getField = memoize(this.getField.bind(this));
     this.getSubfields = memoize(this.getSubfields.bind(this));
     this.getType = memoize(this.getType.bind(this));
+    this.getInitialValue = memoize(this.getInitialValue.bind(this));
```
radekmie, line 97: It'd be nice to keep these ordered.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -218,9 +219,7 @@ export default class JSONSchemaBridge extends Bridge {
     }
 
     if (type === 'array') {
-      const item = this.getInitialValue(joinName(name, '0'));
-      const items = props?.initialCount || 0;
-      return Array.from({ length: items }, () => item);
+      return [];
```
radekmie, line 222: We could actually use `minItems` and do a check, just like I did in https://github.com/vazco/uniforms/pull/1169/files#diff-fcbbc3f73ed7ff24460b9046c0e7b6dd430be61f1a18acbacabe75e5e0939239R93-R95. However, let's do it in a separate PR, to make easily reversible. (I'll raise this topic at our next meeting.)

packages/uniforms-antd/__tests__/ListField.tsx
```diff
@@ -59,23 +59,25 @@ test('<ListField> - renders correct label (specified)', () => {
   );
 });
 
-test('<ListField> - renders correct numer of items with initialCount (specified)', () => {
-  const element = <ListField name="x" initialCount={3} />;
+test('<ListField> - renders correct numer of items with model (specified)', () => {
```
wadamek65, line 62: Not sure on the name change in these

PR #1166, opened 2022-09-20
Remove autoField prop from QuickForm and AutoForm

### LINKED ISSUE #980: Consider removing the `autoField` prop
Since we introduced `AutoField.componentDetectorContext` in #800 and released it in v3.1.0, the `autoField` prop of `QuickForm` (and all its descendant, including `AutoForm`) seems unreasonable. I fully agree with https://github.com/vazco/uniforms/issues/977#issuecomment-869964019:

> It's kind of strange for the implementation not to be passed down to the entire subtree (for me).

My point is, that this prop is there just to "initialize" the `AutoField` a given theme will be using. It's used [only in one place](https://github.com/vazco/uniforms/blob/82b1eee3a0748749e44ba080d8c1f055572a41f5/packages/uniforms/src/QuickForm.tsx#L25) and could be easily removed. However, when removed, we'll always render the "original" `AutoField` of a given theme, which only later will use the `AutoField.componentDetectorContext` to choose the component.

I believe there are cases where an entirely custom `AutoField` makes sense. But I cannot think of one that would not be possible without the `autoField` prop.

My idea looks as follows:
1. Give this issue some time for people to comment on whether they'd have arguments for or against the removal of `autoField`.
2. Deprecate `autoField` by showing a dev-only warning that it's going to be removed in the next major version.
3. Remove it in the next major version.

---

I also thought about making the `autoField` prop automatically populate the `AutoField.componentDetectorContext`. The problem is that it won't work, we expect a [function returning the component](https://github.com/vazco/uniforms/blob/82b1eee3a0748749e44ba080d8c1f055572a41f5/packages/uniforms/src/createAutoField.tsx#L23-L27) in the context, but a component in the `autoField` prop.

wadamek65:
In this PR the `autoField` prop is removed from `QuickForm` and `AutoForm`.

Closes #980.

### DIFF
--- a/packages/uniforms/__tests__/AutoForm.tsx
+++ b/packages/uniforms/__tests__/AutoForm.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import SimpleSchema from 'simpl-schema';
-import { AutoForm, connectField } from 'uniforms';
+import { AutoForm } from 'uniforms';
 import { SimpleSchema2Bridge } from 'uniforms-bridge-simple-schema-2';
 
 import mount from './_mount';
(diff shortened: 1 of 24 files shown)


packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -87,25 +87,6 @@ describe('AutoForm', () => {
   });
 
   describe('when rendered', () => {
-    it('calls `onChange` before render', () => {
```
wadamek65, line 90: Would be best to somehow keep these tests since they are unrelated to the `autoField` prop but I don't really have an idea on how to rewrite it other than move it to a theme instead which I don't think is a good place for it. Any ideas appreciated.

radekmie: Using `AutoField.componentDetectorContext.Provider` to return the `Field` component created here would be enough.

wadamek65: But `AutoField`s are also theme-specific. Can I somehow use the context provider without having one in the base package or did I misunderstand something?

PR #1167, opened 2022-09-20
Move schema component to defaultComponentDetector

### LINKED ISSUE #1114: AutoField.componentDetectorContext is bypassed if schema specify a component
in ReactPage a lot of users asked for having conditional fields in the schema, so by default we support an additional prop `showIf` that can be used in the schema to control whether a field should be shown based on other props. The implementation was fairly easy:

```
import React from 'react';
import { AutoField } from 'uniforms-material';

const AutofieldContextProvider: React.FC = ({ children }) => (
  <AutoField.componentDetectorContext.Provider
    value={(props, uniforms) => {
      const show = props.showIf?.(uniforms.model) ?? true;
      return show
        ? AutoField.defaultComponentDetector(props, uniforms)
        : () => null;
    }}
  >
    {children}
  </AutoField.componentDetectorContext.Provider>
);

export default AutofieldContextProvider;
```

unfortunalty a user noticed that this does not work when you specify a custom component in the schema using uniforms.component prop.

the provider above is never called for these fields.

I think it makes sense at first glance that any custom detector is bypassed when a component is specified, but still, it removes a lot of flexibility like the case above.  Since `autoField` gets removed (https://github.com/vazco/uniforms/issues/980), I actually ran out of ideas how to implement the showIf usecase properly.

the problematic code line is this: https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/createAutoField.tsx#L35

it would probably better if the `componentDetector` is responsible of handling the case where `component` is defined on the field

wadamek65:
In this PR the component supplied by a schema is now moved to the `defaultComponentDetector` for each theme for the `AutoField`.

Closes #1114

### DIFF
--- a/docs/migrating-3-to-4.md
+++ b/docs/migrating-3-to-4.md
@@ -7,4 +7,4 @@ This guide is designed to help you through the migration. If you went through it
 
 ## Breaking API changes
 
-- `componentDetector` in `AutoField`s now always takes precedence over `component` property on a schema. This may make your `AutoField` render a different component when you were using both previously. If that's the case, Move your schema's `component` definition to a [`AutoField.componentDetectorContext.Provider`](/docs/uth-autofield-algorithm/#overriding-autofield) instead.
+- `componentDetector` in `AutoField`s now always takes precedence over `component` property on a schema. This may make your `AutoField` render a different component when you were using both previously. If that's the case, move your schema's `component` definition to a [`AutoField.componentDetectorContext.Provider`](/docs/uth-autofield-algorithm/#overriding-autofield) instead.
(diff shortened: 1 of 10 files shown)


docs/migrating-3-to-4.md
```diff
@@ -6,3 +6,5 @@ title: Migrating v3 to v4
 This guide is designed to help you through the migration. If you went through it and encountered any problems - do let us know. For more information on _why_ certain changes were made, see the [`CHANGELOG.md`](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md). When migrating to v4, use the newest version. Gradual updates will take more time and won't ease this process.
 
 ## Breaking API changes
+
+- `componentDetector` in `AutoField`s now always takes precedence over `component` property on a schema. This may make your `AutoField` render a different component when you were using both previously. If that's the case, Move your schema's `component` definition to a [`AutoField.componentDetectorContext.Provider`](/docs/uth-autofield-algorithm/#overriding-autofield) instead.
```
radekmie, line 10:
```suggestion
- `componentDetector` in `AutoField`s now always takes precedence over `component` property on a schema. This may make your `AutoField` render a different component when you were using both previously. If that's the case, move your schema's `component` definition to a [`AutoField.componentDetectorContext.Provider`](/docs/uth-autofield-algorithm/#overriding-autofield) instead.
```

radekmie (approved):

PR #1172, opened 2022-10-07
Improve JSONSchemaBridge validator types

### LINKED ISSUE #1158: `JSONSchemaBridge` validator return type
**Affected versions** `3.10.0`

## Description
In [`JSONSchemaBridge(schema, validator)`](https://github.com/vazco/uniforms/blob/569b8ff119ff6fa926d23821f9e6570a1dc5f03f/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts#L86) the `validator` function return type is `any` but the `Bridge` [specifies more concrete type (`null | Error | Promise<null | Error>`)](https://github.com/vazco/uniforms/blob/6b1e3b25792b1184c14092c22fc5541b48ec19ae/packages/uniforms/src/Bridge.ts#L127) that can be later returned by the `getValidator`.

## Suggestion
Replace `any` with `null | Error | Promise<null | Error>`.

kestarumper:
## Description
This PR closes #1158 by replacing `any` with more specific validator types.

## Breaking changes
- ~~I had to replace `dataPath` with `instancePath` according to the Ajv v8 `ErrorObject` type. More on that in the [migration v6 to v8 guide](https://ajv.js.org/v6-to-v8-migration.html#new-features)~~
- `getErrorMessages` will always return `string[]` (previously it wasn't casting `number` to `string`, even if `Bridge.getErrorMessages` return type was `string[]`)
- type of `validator` that is passed to `constructor` needs to specifically return a `ValidatorResult | null | undefined` instead of `any` type

### Unknowns
- ~~I wonder if we should _extend_ the `ErrorObject` with the `dataPath` property and be fully backward-compatible~~ - _resolved by extending the type with `dataPath?: string`_.
- ~~I'm not sure about the types that I introduced. I tried to keep it backward-compatible as much as possible, but I don't like it that `getErrorMessages` accepts `string | number`~~ - _resolved by bringing back `unknown`_.
- ~~I'm also unsure if using `Partial<ErrorObject>` (`ErrorObject` from Ajv) is the right direction, but the `getError` heavily uses it (logic for finding the error based on the `instancePath` (previously `dataPath`)~~ - _resolved, we still support `error: unknown`_


kestarumper:
- I've brought back the types of `getError`, `getErrorMessage` and `getErrorMessages` to work with `error: unknown`
- I've also brought back support for `dataPath` from  Ajv < 8

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -1,3 +1,4 @@
+import type { ErrorObject } from 'ajv';
```
radekmie, line 1: I think we can't use `ajv` here. If we do, then it'll be part of our emitted typings and as such, will break if someone will use it without `ajv`. This type is [rather easy](https://github.com/ajv-validator/ajv/blob/16af695b9e19e37fe064145fd33d7a630e9263c0/lib/types/index.ts#L86-L97), so we could inline the part of it we're using.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -77,13 +78,30 @@ function pathToName(path: string) {
   return path.slice(1);
 }
 
+function isValidatorResult(value: unknown): value is ValidatorResult {
+  return typeof value === 'object' && value !== null;
+}
```
radekmie, line 83: This function should also check if `details` is an array.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -77,13 +78,30 @@ function pathToName(path: string) {
...
+type FieldErrorKeywords = string;
+type FieldErrorProperties = Record<string, unknown> & {
+  missingProperty?: string;
+};
+type FieldError = ErrorObject<FieldErrorKeywords, FieldErrorProperties> & {
+  dataPath?: string; // Ajv <8
+};
+type ValidatorResult = Record<string, unknown> & {
+  details?: Partial<FieldError>[];
+};
```
radekmie, line 94: The `Record` is not needed, becasue we only use the `details`.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -108,30 +126,40 @@ export default class JSONSchemaBridge extends Bridge {
     const rootName = joinName(nameParts.slice(0, -1));
     const baseName = nameParts[nameParts.length - 1];
     const scopedError = details.find(error => {
-      const path = pathToName(error.instancePath ?? error.dataPath);
+      const path = pathToName((error.instancePath ?? error.dataPath) || '');
```
radekmie, line 129: I wouldn't call `pathToName` if neither `instancePath` nor `dataPath` are present.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -108,30 +126,40 @@ export default class JSONSchemaBridge extends Bridge {
...
       return (
         unescapedName === path ||
-        (rootName === path && baseName === error.params.missingProperty)
+        (rootName === path &&
+          error.params &&
+          baseName === error.params.missingProperty)
```
radekmie, line 134:
```suggestion
          error.params &&
          baseName === error.params?.missingProperty)
```

radekmie (changes requested):

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
kestarumper: Actually, that one is not necessary because we check for `error.params` truthiness.
On the other hand, if I removed this check and replaced with `?.`, the coverage went down a little because `?.` branch was not covered. I found it difficult to cover, so I just left it.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -111,32 +131,43 @@ export default class JSONSchemaBridge extends Bridge {
...
+    if (isValidatorResult(error)) {
+      const { details } = error;
+      return details.map(error => error.message || '');
+    }
+
+    if (typeof error === 'object') {
+      return [];
+    }
+
+    return ['' + error];
```
Monteth, line 170: Won't you prefer to convert it to string specifically, using `.toString()`? I know it's still JS, but it seems a little bit 'hacky' to me.

Monteth (approved):

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
kestarumper: I don't like `.toString()` if I somehow missed checking for `null` or `undefined`.

radekmie: But both `undefined` and `null` got eliminated in `if (!error)` above.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -78,14 +78,35 @@ function pathToName(path: string) {
   return path.slice(1);
 }
 
+function isValidatorResult(value: unknown): value is ValidatorResult {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    Array.isArray((value as ValidatorResult).details)
```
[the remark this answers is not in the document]
kestarumper: image
Tried that before. The `value` is of type `object` and TS don't like it.
Any ideas?

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
kestarumper: image
🙉

Monteth (dismissed):

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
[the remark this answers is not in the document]
radekmie: Let's merge it as it is and change with newer TypeScript version.

radekmie (approved):

Monteth (approved):

PR #1176, opened 2022-10-11
Drop support for `initialCount` in bridges and `ListField`s

### LINKED ISSUE #1048: Consider removing `props` from `Bridge.getInitialValue`
As described in https://github.com/vazco/uniforms/issues/998#issuecomment-938594418, passing `props` to `Bridge.getInitialValue` was not well-thought:

> [...] It'd be enough to memoize [this `getInitialValue` call](https://github.com/vazco/uniforms/blob/5c2014eb2f82a9cf4c767fe67df7ea7e61cc0d61/packages/uniforms/src/useField.tsx#L92). The problem is that we cannot do that, as it receives `props` that may include arbitrary complex structures. _Why?_ you may ask. And that's a very good question! Only to implement the `initialCount` prop for `ListField` ([JSON schema](https://github.com/vazco/uniforms/blob/5c2014eb2f82a9cf4c767fe67df7ea7e61cc0d61/packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts#L221), [SimpleSchema](https://github.com/vazco/uniforms/blob/5c2014eb2f82a9cf4c767fe67df7ea7e61cc0d61/packages/uniforms-bridge-simple-schema-2/src/SimpleSchema2Bridge.ts#L81)).

As this is the only usage of it within the core themes and we haven't seen any "in the wild", we decided to deprecate this feature and remove it in v4. This issue is a place to raise concerns, questions, or alternative proposals.

wadamek65:
Fixes #1048

Removed support for `initialCount` property on list fields and bridges. `getInitialValue` now doesn't require props being passed to it so it is now memoized. It has also been removed from `filterDomProps`.

Note: Reopened for Hacktoberfest contribution.

### DIFF
--- a/docs/migrating-3-to-4.md
+++ b/docs/migrating-3-to-4.md
@@ -1,0 +1,8 @@
+---
+id: migrating-3-to-4
+title: Migrating v3 to v4
+---
+
+This guide is designed to help you through the migration. If you went through it and encountered any problems - do let us know. For more information on _why_ certain changes were made, see the [`CHANGELOG.md`](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md). When migrating to v4, use the newest version. Gradual updates will take more time and won't ease this process.
+
+## Breaking API changes
--- a/docs/migrating-3-to-4.md
+++ b/docs/migrating-3-to-4.md
@@ -6,3 +6,5 @@ title: Migrating v3 to v4
 This guide is designed to help you through the migration. If you went through it and encountered any problems - do let us know. For more information on _why_ certain changes were made, see the [`CHANGELOG.md`](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md). When migrating to v4, use the newest version. Gradual updates will take more time and won't ease this process.
 
 ## Breaking API changes
+
+- Dropped support for `initialCount` in bridges and `ListField`s. Pass a model object to the form with the appropriate amount of initial items instead.
--- a/packages/uniforms-antd/src/ListAddField.tsx
+++ b/packages/uniforms-antd/src/ListAddField.tsx
@@ -11,18 +11,13 @@ import {
   useField,
 } from 'uniforms';
 
-export type ListAddFieldProps = FieldProps<
-  unknown,
-  ButtonProps,
-  { initialCount?: number }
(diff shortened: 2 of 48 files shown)


docs/migrating-3-to-4.md
```diff
@@ -6,3 +6,5 @@ title: Migrating v3 to v4
 This guide is designed to help you through the migration. If you went through it and encountered any problems - do let us know. For more information on _why_ certain changes were made, see the [`CHANGELOG.md`](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md). When migrating to v4, use the newest version. Gradual updates will take more time and won't ease this process.
 
 ## Breaking API changes
+
+- Dropped support for `initialCount` in bridges and `ListField`s. Pass a model object to the form with the appropriate amount of initial items instead.
```
radekmie, line 10: Let's discuss it briefly, but we should add this file to the main branch and point this PR against it.

packages/uniforms-antd/src/ListAddField.tsx
```diff
@@ -33,10 +28,11 @@ function ListAdd({
 }: ListAddFieldProps) {
   const nameParts = joinName(null, name);
   const parentName = joinName(nameParts.slice(0, -1));
-  const parent = useField<
-    { initialCount?: number; maxCount?: number },
-    unknown[]
-  >(parentName, { initialCount }, { absoluteName: true })[0];
+  const parent = useField<{ maxCount?: number }, unknown[]>(
+    parentName,
```
radekmie, line 32: I guess we could inline `parentName` now? It won't be _as awfully_ formatted as before.

radekmie (changes requested):

packages/uniforms-antd/src/ListAddField.tsx
wadamek65: Not sure, don't have a strong preference because I don't mind the current approach. It's longer but explicit and easy to read, whereas if I inline all of it it would be much less readable IMO.

kestarumper (approved):

radekmie:
Sorry, I closed the `v4` branch. Let's create another one against `master`.

PR #1188, opened 2022-11-04
Adjust dependency versions after 4.0.0-alpha.0 release

### LINKED ISSUE #1187: Fix GitHub Actions CI
It looks like the recent release broke the CI for some reason ([example](https://github.com/vazco/uniforms/actions/runs/3375755191/jobs/5602996243)). While fixing that, we could also take care of the warnings ([see here](https://github.com/vazco/uniforms/actions/runs/3375755191)).

wadamek65:
In this PR:
* Fix for versions in reproductions and website were not matching and were causing a dependency conflict.
* Run `npm dedupe` to deduplicate dependencies.
* Run `npm audit fix` to fix one critical vulnerability.
* Update `actions/checkout` to `v3` which should get rid of the `save-state` warnings.

Fixes #1187


.github/workflows/CI.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         node-version: [14.x, 16.x, 18.x]
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v3
```
radekmie, line 15: Maybe we should use only major GitHub Actions versions? I'm skeptical about it when it comes to code dependencies, but this CI is not business-critical.

radekmie (dismissed):

.github/workflows/CI.yml
wadamek65: Honestly, I don't know. I don't think there is a common standard in actions versioning and most commonly I see just the major versions being used. I thought about locking this to a specific version but after a few minutes of searching for the latest version tag I just decided to use the major one. At the very least I think `actions/checkout` can be trusted. I also don't think I've ever seen using just major versions causing problems yet. 

Conclusion: I'm torn.

radekmie: Then let's pin it to the exact one for now.

radekmie (approved):

PR #1192, opened 2022-11-18
Added `absoluteName` option to `AutoField`.

### LINKED ISSUE #1126: Consider support absolute path render with <AutoField>
## Motivation why you'd like to see this in uniforms.
We have a use case where we want to display a field down to its sister field. Rn, because `uniforms` only supports rendering children fields within a specific field, it is not possible. In general, I think it's also beneficial to enable rendering a field given an absolute path. `uniforms` already provides `useField` with `options` to accept this option, what we need to do is to enable it inside `createAutoField`. 

## Examples of how others could use this feature.
Usage: `<AutoField name="lists.0.name" options={ absoluteName: true} />`

## Possible workarounds, if applicable.
N/A

wadamek65:
In this PR an additional prop `options` was added to `AutoField`. This prop supports an `absoluteName` property which is then passed along to `useField` to enable this functionality for `AutoField`s.

Fixes #1126


packages/uniforms/src/createAutoField.tsx
```diff
@@ -14,6 +14,7 @@ import { useField } from './useField';
 export type AutoFieldProps = {
   component?: Component;
   name: string;
+  options?: { absoluteName?: boolean };
```
wadamek65, line 17: @radekmie Does it make sense to also include `initialValue` here as well since it's also an option for `useField`?

radekmie: I'd still go with the `experimental_` prefix and allow only the `absoluteName` for now. `options` is an extremely popular prop, and hijacking it here would cause problems for a lot of projects.

kestarumper (dismissed):
Great idea with the `__suites__`!

Monteth (dismissed):

kestarumper (approved):

radekmie:
@kestarumper As I wrote in https://github.com/vazco/uniforms/pull/1192#discussion_r1031888872, I'd start only with `experimental_absoluteName` and not the entire `objects`. Mostly because there's only one (for now), and because an object would always have a new reference.

kestarumper (approved):

radekmie (approved):

PR #1193, opened 2022-11-27
Migrate core tests from Enzyme to React Testing Library

ErnestTeluk:
TASK:
- [ ] Rewrite all core test from enzyme to rtl

TODO:
- [x] Fix ts error related with ref in `<ValidateForm />` tests.
- [x] Rewrite the last `describe` in `<ValidateForm />` tests.
- [x] Fix skipped tests in `<AutoForm />` as in `<ValidateForm />`.
- [x] change 'test' to 'it'
- [x] get `context` from jest.fn (last call argument)
- [x] add to `render` rerenderWithProps fn with  React.cloneElement.
- [ ] rewrite `<BaseForm />`, `connectedFields`, `<QuickForm />`


packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -99,85 +121,128 @@ describe('AutoForm', () => {
...
+          </context.Consumer>
+        </AutoForm>,
+        {
+          schema: { type: SimpleSchema2Bridge },
+        },
       );
 
-      wrapper.instance().reset();
-      expect(wrapper.instance().getContext().model).toEqual(model);
+      rerenderWithProps({});
```
kestarumper, line 181: Could we avoid calling `rerenderWithProps({})` if we don't want to override any props? Use `rerender`, this way we can skip the unnecessary `cloneElement` operation.

packages/uniforms/__suites__/render.tsx
```diff
@@ -37,11 +39,19 @@ export function render(
...
         <context.Provider value={contextValue}>{children}</context.Provider>
       );
     },
   });
+
+  const { rerender } = originalRender;
+
+  const rerenderWithProps = (props: Record<any, any>) => {
+    rerender(cloneElement(element, props));
+  };
```
kestarumper, line 54: [opinion]
This `props: Record<any, any>` worries me. It can lead to overriding props with a different type than the component accepts. However, I know we receive `ReactElement` in the `render` function, so there is no way of inferring the `ComponentProps<T>` type. I don't know what to do here.

The beta React docs also [discourage the use of `cloneElement`](https://beta.reactjs.org/apis/react/cloneElement#alternatives).

RTL, on the other hand, also approaches the problem differently [in their docs](https://testing-library.com/docs/example-update-props/) (and [on this blog](https://www.designcise.com/web/tutorial/how-to-update-the-props-of-a-rendered-component-in-react-testing-library)). However they show an example with rendering just a single component, but we usually render `<AutoForm>{a lot of children}</AutoForm>` which might lead to a lot of repetition.

kestarumper (changes requested):

packages/uniforms/__suites__/render.tsx
radekmie: 1. You're right - `cloneElement` is discouraged. But not "in general", but when a render prop would be better. Here we're in a different context and I think it's totally fine.
2. I agree that `Record<any, any>` is far too generic. `Record<string, unknown>` would be already better. But `ReactElement` has two type parameters and actually `Props` is the first one, so we should use that.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -1,23 +1,27 @@
...
-  const model = { a: '1' };
+describe('<AutoForm />', () => {
   const schema = new SimpleSchema2Bridge(
     new SimpleSchema({
       a: { type: String, defaultValue: '' },
       b: { type: String, defaultValue: '' },
       c: { type: String, defaultValue: '' },
     }),
   );
+  const model = { a: '1' };
```
radekmie, line 18: There's no point in touching this code - please revert it.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
     validator.mockClear();
   });
 
-  describe('when changed', () => {
+  describe('when changes', () => {
```
radekmie, line 34: I think the previous naming convention was better. If you have any arguments for changing it, let's discuss it.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
...
     it('updates', () => {
-      // FIXME: AutoForm is not a valid Component.
-      const wrapper = mount<AutoForm | any>(
+      render(
         <AutoForm onChange={onChange} schema={schema} />,
+        { schema: { type: SimpleSchema2Bridge } },
```
radekmie, line 38: What is this argument? It should be used as the schema for the context, but we don't actually need seeding context here - we're rendering forms in these tests. _(Also, this is not a valid schema. At least for any value.)_

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
...
+        { onChange },
       );
-
-      wrapper.instance().getContext().onChange('a', '2');
-
-      expect(onChange).toHaveBeenCalledTimes(1);
-      expect(onChange).toHaveBeenLastCalledWith('a', '2');
```
radekmie, line 40: This test has no assertions.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
...
     });
-
     it('validates', () => {
       // FIXME: AutoForm is not a valid Component.
-      const wrapper = mount<AutoForm | any>(
-        <AutoForm onChange={onChange} schema={schema} />,
+      render(
+        <AutoForm
+          // TODO: delete ts-expect-error if this issue is resolved https://github.com/vazco/uniforms/issues/1165
+          // @ts-expect-error
```
radekmie, line 47: `@ts-expect-error` allows for adding comments and will cause `tsc` fail when no longer needed, so I propose this instead:
```suggestion
          // @ts-expect-error https://github.com/vazco/uniforms/issues/1165
```

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
...
       );
 
-      wrapper.instance().submit();
+      const form = screen.getByRole('form');
+      const input = screen.getByLabelText('A');
+      fireEvent.submit(form);
 
       expect(validator).toHaveBeenCalledTimes(1);
-      expect(validator).toHaveBeenLastCalledWith({});
+      expect(validator).toHaveBeenLastCalledWith({ a: '', b: '', c: '' });
```
radekmie, line 63: Have you checked where is this change coming from? It's rather unexpected.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -27,66 +31,84 @@ describe('AutoForm', () => {
...
-      const context1 = wrapper.instance().getContext();
-      expect(context1).toHaveProperty('changed', false);
-      expect(context1).toHaveProperty('changedMap', {});
-
-      wrapper.instance().getContext().onChange('a', '2');
+      render(
+        <AutoForm schema={schema}>
+          <context.Consumer>
+            {context => mockContext(context?.changed, context?.changedMap)}
+          </context.Consumer>
```
radekmie, line 96: This could be simplified when the spy would get the whole `context`. Something like this:

```ts
const contextSpy = jest.fn<void, [Context<unknown>]>();

// Then:
<context.Consumer children={spy} />
```

packages/uniforms/__tests__/useField.tsx
```diff
@@ -1,100 +1,99 @@
...
-          .find('input')
-          .simulate('change', { target: { value: undefined } }),
-      ).toBeTruthy();
+      expect(input).toHaveAttribute('value', '4');
     });
 
     it('does not apply default value after first change', () => {
-      const wrapper = mount(
-        <AutoForm schema={bridge}>
+      const { getByRole } = render(
```
radekmie, line 60: I would always call it the same. Either `screen` or `renderResult` (then the same in `render.tsx`).

packages/uniforms/__tests__/useField.tsx
```diff
@@ -1,100 +1,99 @@
...
 
-  function Test(rawProps: {
-    children?: ReactNode;
-    name: string;
-    options?: Parameters<typeof useField>[2];
-  }) {
-    const [props] = useField(rawProps.name, rawProps, rawProps.options);
-    return <>{props.children}</>;
-  }
+const TestComponent = connectField((props: Record<string, any>) => {
```
radekmie, line 30: 1. I'd rather use `useField` here.
2. Use proper `props` type.

radekmie (changes requested):
I have skipped `ValidatedForm` for now, as there are already quite a few comments.

packages/uniforms/__suites__/render.tsx
kestarumper: > `ReactElement` has two type parameters and actually `Props` is the first one
Nice!

packages/uniforms/__suites__/render.tsx
```diff
@@ -11,7 +11,9 @@ export function render(
   schema: SimpleSchemaDefinition,
   contextValueExtension?: Partial<Context<unknown>>,
   initialModel?: object,
-) {
+): RenderResult & {
+  rerenderWithProps: (props: Record<any, any>) => void;
+} {
```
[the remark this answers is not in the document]
ErnestTeluk: we need provide this type because when we don't we have this error

```
TS2742: The inferred type of 'render' cannot be named without a reference to '@testing-library/dom/node_modules/pretty-format/build/types'. This is likely not portable. A type annotation is necessary.
```

packages/uniforms/__suites__/render.tsx
```diff
@@ -37,11 +39,19 @@ export function render(
...
     },
   });
+
+  const { rerender } = originalRender;
+
+  const rerenderWithProps = (props: Record<any, any>) => {
+    rerender(cloneElement(element, props));
+  };
+
+  return { rerenderWithProps, ...originalRender };
```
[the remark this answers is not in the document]
ErnestTeluk: when we provide the type of return. I think we can stay with the spread.

packages/uniforms/__suites__/render.tsx
```diff
@@ -37,11 +39,19 @@ export function render(
     formRef: {} as BaseForm<unknown>,
   };
 
-  return renderOnScreen(element, {
+  const originalRender = renderOnScreen(element, {
```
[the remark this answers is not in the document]
ErnestTeluk: fixed

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: fixed

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: fixed

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: fixed

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: fixed

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: fixed, but I need to verify why onChange is called 4 times. I think is unexpected behavior.

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: this is due to the use of <AutoFields> he adds fields from the schema and performs onChangeModel and onChange which gives these values, therefore Validate is called with empty strings

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: I think we should consider changing <AutoFields> because now it adds fields iteratively which makes it execute n times onChangeModel, maybe we should collect them and then do createElement once ?

packages/uniforms/__tests__/AutoForm.tsx
ErnestTeluk: I have changed the render function types and now the schema field is not required, If we don't pass it our component will not have a `<Context.Provider>`. I have also added the correct schema where we use context.

packages/uniforms/__suites__/render.tsx
```diff
@@ -1,47 +1,56 @@
...
+            showInlineError: false,
+            ...contextValueExtension?.state,
+          },
+          formRef: {} as BaseForm<UnknownObject>,
+        };
+        return (
+          <context.Provider value={contextValue}>{children}</context.Provider>
+        );
+      }
+      return <>{children}</>;
```
radekmie, line 45: Do an early return instead - it'll have fewer indents.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -1,92 +1,105 @@
...
-
-      expect(onChange).toHaveBeenCalledTimes(1);
+      const input = screen.getByLabelText('A');
+      fireEvent.change(input, { target: { value: '2' } });
+      expect(onChange).toHaveBeenCalledTimes(4);
       expect(onChange).toHaveBeenLastCalledWith('a', '2');
     });
-
     it('validates', () => {
       // FIXME: AutoForm is not a valid Component.
```
radekmie, line 43: This comment is no longer relevant.

packages/uniforms/__tests__/ValidatedForm.tsx
```diff
@@ -1,193 +1,380 @@
...
 describe('ValidatedForm', () => {
   const onChange = jest.fn();
   const onSubmit = jest.fn();
   const onValidate = jest.fn((model, error) => error);
   const validator = jest.fn();
   const validatorForSchema = jest.fn(() => validator);
+  const contextSpy = jest.fn<ReactNode, [Context<any> | null]>();
 
-  const error = new Error();
+  const error = new Error('test error message');
```
radekmie, line 25: I'd remove this message if it's not needed.

packages/uniforms/__tests__/AutoForm.tsx
```diff
@@ -1,92 +1,105 @@
...
+  const contextSpy = jest.fn<ReactNode, [Context<any> | null]>();
   const model = { a: '1' };
-  const schema = new SimpleSchema2Bridge(
-    new SimpleSchema({
-      a: { type: String, defaultValue: '' },
-      b: { type: String, defaultValue: '' },
-      c: { type: String, defaultValue: '' },
-    }),
-  );
+  const schemaValues = {
```
radekmie, line 17: It's called _schema definition_, not _schema values_.

radekmie (changes requested):
Last round. Also, if you could look into the dip in coverage, that'd be great.

radekmie (approved):
OK, let's get this one closed. We'll improve tests later.

PR #1202, opened 2022-12-16
Update uniforms docs

kestarumper:
## Changes
- Fixed typos
- Replaced legacy ref with `useRef`

### Notes & problems
In reality, if you use the `useForm` without passing in the type (for example, to `AutoForm`), you won't have it typed correctly - you won't be able to call, for example, `formRef.current.reset()`. An explicit type needs to be given to access the methods of the referenced element.

### DIFF
--- a/docs/api-fields.md
+++ b/docs/api-fields.md
@@ -110,7 +110,8 @@ A checkbox.
 ```tsx
 import { BoolField } from 'uniforms-unstyled';
 import { useRef } from 'react'
-const ref = useRef()
+
+const inputRef = useRef();
 
 <BoolField
   appearance="checkbox" // Renders a material-ui Checkbox
@@ -125,7 +126,7 @@ const ref = useRef()
   helpClassName="a b c"
   inline
   inputClassName="a b c"
-  inputRef={ref}
+  inputRef={inputRef}
   labelBefore="Label"
   labelClassName="a b c" // You can either specify them as a single string
   labelClassName=[ 'a', 'b', 'c' ] // or as an array of strings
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -302,14 +304,12 @@ The provider for this context is rendered by `BaseForm`, and in turn all the oth
...
+   To fix this, wrap this component within a parent Form component (does not have to be direct).
 
-   more than one version of the core `uniforms` package installed which can happen when you have a mismatch of
+2. **There are multiple versions of `uniforms` installed in your `node_modules`**.
 
-   versions between any of your `uniforms` related dependencies. Ensure all your versions are matching,
+   This usually happens when you have installed more than one version of the core `uniforms` package. It can happen when you have a mismatch of versions between any of your `uniforms` related dependencies.
 
-   clean any `node_modules` directories and reinstall dependencies to resolve this error.
+   Ensure all your versions match, clean any `node_modules` directories and reinstall dependencies to resolve this error.
```
Monteth, line 315: Maybe let's put it straight forward and say that by 'match', we mean 'they are the same'.

Monteth (dismissed):

docs/api-fields.md
```diff
@@ -109,6 +109,8 @@ A checkbox.
 
 ```tsx
 import { BoolField } from 'uniforms-unstyled';
+import { useRef } from 'react'
+const ref = useRef()
```
radekmie, line 113: 1. Let's keep an empty line between the imports and variables.
2. Add semicolons (Prettier should do that...?).
3. I'd call it `inputRef` here (and `formRef` accordingly).

radekmie (changes requested):

docs/api-fields.md
kestarumper: Huh, prettier does fix it, but it also adds `;` at the end of the `tsx`

e.g.
```tsx
<AutoField name="firstName" inputRef={inputRef} />;
```

radekmie: I think it's totally fine.

radekmie (approved):

Monteth (approved):

Issue #1083, opened 2022-01-10
Explicit form types introduced on v3.6.0


afmfe-iul:
Hey there!
On our project we use uniforms and TypeScript (v4.5.4) extensively and last week we upgraded `uniforms` / `uniforms-semantic` / `uniforms-bridge-json-schema` from v3.5.5 to version v3.7.0, which caused a couple of type errors. After going through the [changelog](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md) we found out that very likely #1003 (from v3.6.0) is where the issues started.

Specifically we are getting errors on `AutoForm` prop types:
1. `className` is not a valid prop, however adding one works;
2. `onChangeModel` is not a valid prop;
3. `onSubmit` is of type `(model: DeepPartial<Model>) => void | Promise<any>`, however we always assumed that it would be `(model: Model) => void | Promise<any>` because if you passed validation (and you typed everything correctly) by the time you reach submission your model would be matching the required type.

[Here](https://codesandbox.io/s/magical-vaughan-pcish?file=/App.tsx:339-385) is a simple example on CodeSandbox (however sometimes the compiler errors don't show because their compiler is fixed at 4.4.4 it seems). If you try this example outside CodeSandbox, with TS 4.5.4, you might get the same results as we.

Currently the workaround we are using is the one sugest on the FAQ (use of `any`).

Would love to get your thoughts on the points above!

radekmie:
Hi @afmfe-iul! You've found the correct suspect here - #1003. And yes, as a current workaround, casting the form to `any`.

Now, let's get to the errors step by step:
1. `className` should be inferred from the `HTMLFormElement` props, but it's not now. I see that as a proper improvement that we could do. However, it's possible to override the actually rendered component, so we should somehow carry this information.
    * Another generic argument makes sense, but it's hard to override.
2. `onChangeModel` is added by the `AutoForm` and [this problem](https://github.com/vazco/uniforms/blob/1b4591a9c14e4c69cb08ca7fd37a886b6dbeee38/packages/uniforms/src/AutoForm.tsx#L26) probably hides the correct type here.
3. I'm still unsure on how to think about that. On the one hand, we are aware that we start with an empty object - not a valid model. But on the other, we should trust the validator in this manner.

I'd love to hear your suggestions here: @afmfe-iul, @wadamek65, @kestarumper, @Monteth.

wadamek65:
[...]

PR #1205, opened 2022-12-27
Stricter types

radekmie:
In this pull request, I narrowed and improved quite a few of core uniforms types. Most notably:
* All `model`s (in the `*Form` components' props, `Context`, `modelTransform`, etc.) are no longer wrapped with `DeepPartial`. The idea is that the form always operates on a "valid" data structure. It covers 3rd issue of #1083.
* Every `Model` has to extend `Record<string, unknown>`. On the one hand it helps with object manipulation here and there, on the other that's potentially safer than `object`. (I'm open for a different base here.)
* Narrowed a lot of `any`s to `unknown`s. The only "severe" ones left are the ones in the bridges, but they deserve a separate PR (and one of them is already started in #1172).

_(Side note: we should switch to Zod or some other schema for our internal tests, as Simple Schema is not really TypeScript friendly.)_


packages/uniforms-unstyled/__tests__/_createContext.ts
```diff
@@ -32,7 +33,7 @@ export default function createContext(
         showInlineError: false,
         ...context?.state,
       },
-      formRef: {} as BaseForm<unknown>,
-    },
+      formRef: {} as BaseForm<Model>,
+    } as Context<Model>,
```
kestarumper, line 37: Not a big fan of `as Type`. It can lead to some errors. Maybe instead, we could use `Context<Model>` as a return type of the `createContext`?

packages/uniforms-bootstrap5/src/AutoField.tsx
```diff
@@ -13,7 +14,7 @@ import TextField from './TextField';
 
 const AutoField = createAutoField(props => {
   if (props.component) {
-    return props.component;
+    return props.component as ComponentType<any>;
```
kestarumper, line 17: Isn't `props.component` already a `ComponentType<any>`?

Btw, there is also a `Component` type that is an alias for
```tsx
/** @internal */
export type Component = ComponentType<any> | ReturnType<typeof connectField>;
```

wadamek65 (approved):

packages/uniforms/src/useForm.ts
```diff
@@ -2,9 +2,9 @@ import invariant from 'invariant';
 import { useContext } from 'react';
 
 import { context as contextReference } from './context';
-import { Context } from './types';
+import { Context, UnknownObject } from './types';
 
-export function useForm<Model>(): Context<Model> {
+export function useForm<Model extends UnknownObject>(): Context<Model> {
```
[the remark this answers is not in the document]
radekmie: I think it's better not to have a default, as that'll force people to use a proper type. We can change it later too.

wadamek65 (approved):

kestarumper (approved):

Issue #1213, opened 2023-01-31
[MUI] and [Material] `min` and `max` properties are not being applied to a Date field


vcardins:
The `min` and `max` properties are necessary to avoid that arbitrary values re entered in a `Date` type field, however, even setting 
```javascript
<AutoField
	type="date"
	variant="standard"
	min={<any value>}
	max={<any value>}
/>
```
they are not applied.

The code

![image](https://user-images.githubusercontent.com/2383329/215632988-f06f8c9f-9b61-4984-b03c-5ac4919ef4eb.png)

generates the html

![image](https://user-images.githubusercontent.com/2383329/215633117-421defca-6c22-448c-8943-3c43584da4b5.png)

radekmie:
Hi @vcardins and thanks for the report! That's true, neither of these two themes supports that, but all of the other themes do. Feel free to file a PR with that!

vcardins:
@radekmie Please review https://github.com/vazco/uniforms/pull/1215.

PR #1215, opened 2023-02-08
Added the missing 'min' and 'max' props to the Material UI and MUI packages

vcardins:
Following up on the issue https://github.com/vazco/uniforms/issues/1213#issuecomment-1416058859, this PR is suppose to fix it.

### DIFF
--- a/packages/uniforms-material/src/DateField.tsx
+++ b/packages/uniforms-material/src/DateField.tsx
@@ -1,10 +1,13 @@
 import { useTheme } from '@material-ui/core';
 import TextField, { TextFieldProps } from '@material-ui/core/TextField';
 import React from 'react';
-import { FieldProps, connectField, filterDOMProps } from 'uniforms';
+import { FieldProps, connectField, filterDOMProps, DateFieldType } from 'uniforms';
 
 /* istanbul ignore next */
 const DateConstructor = (typeof global === 'object' ? global : window).Date;
+
+const dateFormat = (value?: Date, type: DateFieldType = 'datetime-local') =>
+  value?.toISOString().slice(0, type === 'datetime-local' ? -8 : -14);
 
 const dateParse = (timestamp: number, onChange: DateFieldProps['onChange']) => {
   const date = new DateConstructor(timestamp);
@@ -20,6 +23,8 @@ export type DateFieldProps = FieldProps<
   TextFieldProps,
   {
     labelProps?: object;
+    max?: Date;
+    min?: Date;
     type?: 'date' | 'datetime-local';
   }
 >;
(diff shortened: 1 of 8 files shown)


packages/uniforms-material/src/DateField.tsx
```diff
@@ -55,6 +60,8 @@ function Date({
       InputLabelProps={{ shrink: true, ...labelProps, ...InputLabelProps }}
       inputProps={{ readOnly, ...(themeProps?.inputProps ?? {}) }}
       margin={themeProps?.margin ?? 'dense'}
+      max={dateFormat(max)}
```
radekmie, line 63: Please check whether it can be safely passed here.

radekmie (changes requested):
We'd also need tests for that.

kestarumper:
Hi @vcardins,
Could you give us your `node -v` and `npm -v` versions?
Also, please use `npm ci` instead.

vcardins:
My current node and npm version. Also, the error output
image

vcardins:
I use `nvm` and tried with node v- 16.0.0 but also didn't work.

radekmie:
@vcardins I tried it now on Node.js 18.13.0 and npm 8.19.3. However, the error you posted suggests a binary dependency problem. Based on the logs you're using macOS - what CPU is it? I'm on M2 and it works correctly. Also, have you tried clearing the npm cache (`npm cache clear --force`)?

vcardins:
Very frustrating for me, guys. I'll try on my windows machine when at home.
image

ErnestTeluk:
I have tried `node v19.6.0` and `npm 9.4.0` and I get the same error as you but if you run it on the version that @radekmie  suggests it works for me (working on a mac with m1 pro). I see from your screen that after using `npm cache clear --force` and nvm use node you are still using `node v19.06.0` and `npm 9.4.0`. Try using `node 18.13.0` and `npm 8.19.3`.

vcardins:
I manage to fix the `npm` errors but now when I run `npm test` I get
image
One thing that I noticed is `uniforms-mui` package uses @types/react 18.x.x, while the main repo `"@types/react": "17.0.39"`.
It's getting closer but still far. :)

vcardins:
Also, the CI errors that are happening now has nothing to do with my changes. https://github.com/vazco/uniforms/actions/runs/4226780890/jobs/7340589734

radekmie:
Uff, at least we're moving on 😅 I think it may be the problem of transitive dependencies. Did you use `npm ci` or `npm i`? It definitely should be the former.

As for the second point, I think it's actually because of this change - the `type` parameter is not passed to `dateFormat` calls.

vcardins:
I forgot to mention that I had to add @ts-ignore for this to pass. Does it work fine for you?
image

vcardins:
ahahhahah It's ok, we get this :D. I'll push soon

PR #1229, opened 2023-02-24
Replace property `allowedValues` + `transform` with `options`

### LINKED ISSUE #806: Property 'options' does not exist on <SelectField /> (uniforms-semantic)
```
Type '{ name: string; options: { label: string; value: string; }[]; onChange: (repositoryId: string) => void; }' is not assignable to type 'IntrinsicAttributes & { label?: ReactNode; name: string; placeholder?: string | boolean | null | undefined; } & Pick<Partial<GuaranteedProps<string | string[] | undefined>>, "id" | ... 9 more ... | "showInlineError"> & Pick<...> & { ...; }'.
  Property 'options' does not exist on type 'IntrinsicAttributes & { label?: ReactNode; name: string; placeholder?: string | boolean | null | undefined; } & Pick<Partial<GuaranteedProps<string | string[] | undefined>>, "id" | ... 9 more ... | "showInlineError"> & Pick<...> & { ...; }'.
```
**Versions**:
- uniforms `3.0.0-rc3`
- uniforms-semantic `3.0.0-rc3`

**Description**:
Property `options` do not exist but are used to calculate `allowedValues`. The only problem is that they're missing in the typescript definitions.

kestarumper:
This PR closes #806.

### DIFF
--- a/docs/uth-autofield-algorithm.md
+++ b/docs/uth-autofield-algorithm.md
@@ -15,6 +15,10 @@ In [\#800](https://github.com/vazco/uniforms/issues/800), a new way of creating 
 import { createAutoField } from 'uniforms';
 
 const AutoField = createAutoField(props => {
+  if (props.component) {
+    return props.component;
+  }
+
   if (props.options) {
     return props.checkboxes && props.fieldType !== Array
       ? RadioField
--- a/docs/uth-autofield-algorithm.md
+++ b/docs/uth-autofield-algorithm.md
@@ -15,7 +15,7 @@ In [\#800](https://github.com/vazco/uniforms/issues/800), a new way of creating 
 import { createAutoField } from 'uniforms';
 
 const AutoField = createAutoField(props => {
-  if (props.allowedValues) {
+  if (props.options) {
     return props.checkboxes && props.fieldType !== Array
       ? RadioField
       : SelectField;
--- a/packages/uniforms-antd/src/RadioField.tsx
+++ b/packages/uniforms-antd/src/RadioField.tsx
@@ -32,6 +32,7 @@ function Radio(props: RadioFieldProps) {
           key={option.key ?? option.value}
           style={radioStyle}
           value={option.value}
+          disabled={option.disabled}
         >
           {option.label ?? option.value}
         </RadioAntD>
--- a/packages/uniforms-antd/src/RadioField.tsx
+++ b/packages/uniforms-antd/src/RadioField.tsx
@@ -17,6 +17,7 @@ function Radio(props: RadioFieldProps) {
   return wrapField(
     props,
     <RadioAntD.Group
+      {...filterDOMProps(props)}
       disabled={props.disabled}
       name={props.name}
       onChange={event => {
@@ -25,7 +26,10 @@ function Radio(props: RadioFieldProps) {
         }
       }}
       value={props.value ?? ''}
-      {...filterDOMProps(props)}
+      options={props.options?.map(option => ({
+        ...option,
+        label: option.label ?? option.value,
+      }))}
     >
       {props.options?.map(option => (
         <RadioAntD
--- a/packages/uniforms-antd/src/RadioField.tsx
+++ b/packages/uniforms-antd/src/RadioField.tsx
@@ -27,8 +27,12 @@ function Radio(props: RadioFieldProps) {
       {...filterDOMProps(props)}
     >
       {props.options?.map(option => (
-        <RadioAntD key={option.key} style={radioStyle} value={option.value}>
-          {option.label}
+        <RadioAntD
+          key={option.key ?? option.value}
+          style={radioStyle}
+          value={option.value}
+        >
+          {option.label ?? option.value}
         </RadioAntD>
       ))}
     </RadioAntD.Group>,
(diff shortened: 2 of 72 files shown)


docs/uth-autofield-algorithm.md
```diff
@@ -15,7 +15,7 @@ In [\#800](https://github.com/vazco/uniforms/issues/800), a new way of creating
 import { createAutoField } from 'uniforms';
 
 const AutoField = createAutoField(props => {
-  if (props.allowedValues) {
+  if (props.options) {
```
radekmie, line 18: Can you add the `if (props.component)` part here too? It's missing in the docs.

packages/uniforms-antd/src/RadioField.tsx
```diff
@@ -26,9 +27,13 @@ function Radio(props: RadioFieldProps) {
...
     >
-      {props.allowedValues?.map(value => (
-        <RadioAntD key={value} style={radioStyle} value={value}>
-          {props.transform ? props.transform(value) : value}
+      {props.options?.map(option => (
+        <RadioAntD
+          key={option.key ?? option.value}
+          style={radioStyle}
+          value={option.value}
+        >
```
radekmie, line 35: We could add the `disabled` here as well.

packages/uniforms-antd/src/types.ts
```diff
@@ -0,0 +1,7 @@
+/** Option type used in SelectField or RadioField */
+export type Option<Value = unknown> = {
```
radekmie, line 2: We pass the `Value` everywhere, so there's no need for a default here.

packages/uniforms-antd/src/types.ts
```diff
@@ -0,0 +1,7 @@
...
+  disabled?: boolean;
+  label: string;
```
radekmie, line 4: Isn't the `label` optional as well?

packages/uniforms-bootstrap3/src/RadioField.tsx
```diff
@@ -14,38 +15,37 @@ export type RadioFieldProps = HTMLFieldProps<
...
       <div
-        key={item}
+        key={item.key ?? item.value}
         className={classnames(
           props.inputClassName,
           `radio${props.inline ? '-inline' : ''}`,
         )}
       >
-        <label htmlFor={`${props.id}-${escape(item)}`}>
+        <label htmlFor={`${props.id}-${escape(item.key ?? item.value)}`}>
```
radekmie, line 35: There's no need to `escape` the `key`:
```suggestion
        <label htmlFor={`${props.id}-${item.key ?? escape(item.value)}`}>
```

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -306,21 +314,24 @@ export default class JSONSchemaBridge extends Bridge {
...
+        options = Object.entries(options).map(([key, value]) => ({
+          key,
+          label: key,
+          value,
+        }));
       }
     } else if (props.enum) {
-      props.allowedValues = props.enum;
+      options = Object.values(props.enum).map(value => ({
+        label: String(value),
```
radekmie, line 332: I wouldn't add it here, as `label` fallbacks to `value` anyway.

packages/uniforms-bridge-simple-schema-2/__tests__/SimpleSchema2Bridge.ts
```diff
@@ -218,32 +227,57 @@ describe('SimpleSchema2Bridge', () => {
...
+      expect(bridge.getProps('rr').options[0]).toStrictEqual({
+        key: 'k1',
+        label: 'A',
+        value: 1,
+      });
+      expect(bridge.getProps('rr').options[1]).toStrictEqual({
+        key: 'k2',
+        label: 'B',
+        value: 2,
       });
```
radekmie, line 281: It'd be better to test the entire `options`, not only some properties. It'll also make the test smaller.

packages/uniforms-bridge-simple-schema-2/src/SimpleSchema2Bridge.ts
```diff
@@ -109,6 +117,12 @@ export default class SimpleSchema2Bridge extends Bridge {
     return undefined;
   }
 
+  private allowedValueToOption = <T>(value: T): Option<T> => ({
+    key: String(value),
+    label: String(value),
+    value,
+  });
```
radekmie, line 124: 1. As above, there's no need to provide `key` and `label`.
2. There's no point in making it a method - either inline it entirely, or make a helper outside of the class.

radekmie (changes requested):

packages/uniforms-bridge-zod/src/ZodBridge.ts
```diff
@@ -150,11 +159,19 @@ export default class ZodBridge<T extends ZodRawShape> extends Bridge {
         props.minCount = field._def.minLength.value;
       }
     } else if (field instanceof ZodEnum) {
-      props.allowedValues = field.options;
+      props.options = (field.options as ZodEnum<[string]>['options']).map<
+        Option<unknown>
+      >(value => ({
+        key: value,
+        label: value,
```
radekmie, line 166: I'd leave these out as they're optional and default to `value` anyway.

radekmie (approved):

zendranm (approved):

PR #1247, opened 2023-03-17
Migrated `HiddenField` tests to @testing-library/react.

ErnestTeluk:
Rewrite `HiddenField` tests to @testing library/react


packages/uniforms/__suites__/HiddenField.tsx
kestarumper, line 1: [opinion]
I know it's been like this before, but let's think if we can improve the test names.

Can we get rid of phrases _"correctly does X"_ or _"with correct Y"_.
The idea of tests is to be correct. The test descriptions are too vague and should be explicit what it tests and under what assumptions.

Eg.1
- [before]`<HiddenField> - renders an input with correct id (inherited)`
- [after] `<HiddenField> - render an input with inherited id from testId` (or even `inherits id from testId`)

Eg.2.
- [before] `<HiddenField> - renders an input with correct value (model)`
- [after] `<HiddenField> - uses value from model`

etc.

PS: Let me know if you like it or not and if we should leave it as is.

Issue #973, opened 2021-05-31
Configuring label and placeholder in schema


pavankramadugu:
```
const schema = {
  type: 'object',
  properties: {
    email: {
      type: 'string',
      uniforms: { placeholder: true }
    },
  },
};
```
This is not showing placeholder as email in the ui. Using bootstrap5 3.4.0

I also have another query,

```
import LongTextField from "uniforms-bootstrap5"
const schema = {
  type: 'object',
  properties: {
    email: {
      type: 'string',
      uniforms: { component: "LongTextField", placeholder: true }
    },
  },
};
```
The issue here is i'm getting this json from server LongTextField is coming as "LongTextField" so it's not loading the schema. how to approach this when we try to get schema dynamically from server. Due to this i'm unable to use custom comps.

wadamek65:
Hi @pavankramadugu , thanks for the issue report.

1. After some initial investigation, I think you want to set `uniforms.placeholder` property to a string, and pass a `placeholder={true}` prop in your component explicitly to make it show up. [Like in this example in the playground.](https://uniforms.tools/playground#?N4IgDgTgpgzlAuIBcIDqUA2BjA9gWygEIQAacCHMGZUAQwFd4cZaA3KZAM1ozjIaYt2AEUy0AnsgCMABhlkAJgEsWAIwxQFXHnxAZaqzNt5QyYfVigALHBgVQIyeBHqmQ0WgoDyAOwySkbhMyGCwrKDxaZBAfKAB3AAIAZSU8cygksIjaACYAIQglBQBzKAAKAB0fBITYxJS0jUzwyLLgKpqa7KUMJAT26s6a-HEwKD6k5yUfYpIOoYT6HyVOHAg8GD7gBPNaSxs7Bz6AcnhYeGOEgF95mqu5nyuASiqn0hAYGziAST9pqAAohAKI5Ajo3LQYOIfFhfEl6Ko8EpEGDgiBIdDYT4AGo8Iq0M7GOD3EDwFocFCqHA4eAwZy0MAAVhAVyAA) I think there might be some underlying issue/inconsistency in our code though about how the placeholder prop is handled but this will require further investigation. (@radekmie @Monteth pinging for visibility if one of you maybe already has the answer).
2. As for your second question, certainly there is no way to use a custom component out of the box in a schema that is fetched from external sources. A solution that might help is maybe to create a helper function which will help you derive the schema, something like this:

```ts
const CustomLongTextField = () => {}; // Your custom component
[...]

PR #1248, opened 2023-03-17
Remove label and placeholder from context and form, add optional provideDefaultLabelFromFieldName property to bridges.

Monteth:
Changes are described here: https://github.com/vazco/uniforms/issues/973#issuecomment-1235430047.

This version doesn't include the `provideDefaultLabelFromFieldName` bridge property in SimpleSchema, since it always delivers a 'label' property. We cannot distinguish whether it was inferred from the field name or entered manually into the schema.


packages/uniforms-bridge-graphql/__tests__/GraphQLBridge.ts
```diff
@@ -225,12 +224,16 @@ describe('GraphQLBridge', () => {
 
   describe('#getProps', () => {
     describe('labels are derived properly', () => {
-      describe('when props.label is undefined or true', () => {
+      describe('when provideDefaultLabelFromFieldName is true', () => {
+        const bridgeTDefaultLabel = new GraphQLBridge(
+          astT.getType('Post')!,
+          schemaValidator,
+          schemaData,
+        );
```
radekmie, line 232: It's missing the `provideDefaultLabelFromFieldName`, right?

packages/uniforms-bridge-graphql/src/GraphQLBridge.ts
```diff
@@ -125,7 +125,12 @@ export default class GraphQLBridge extends Bridge {
       props.decimal = true;
     }
 
-    props.label ??= upperFirst(lowerCase(field.name));
+    if (
+      this.extras.provideDefaultLabelFromFieldName &&
+      props.label === undefined
+    ) {
+      props.label = upperFirst(lowerCase(field.name));
+    }
```
radekmie, line 133: I think we should make this bridge accept an object now. The thing is, `extras` is a map from field name to its extra props. In that case, if someone will have a `provideDefaultLabelFromFieldName` field (for some reason), it's going to break.

I propose to accept an object _plus_ this new field instead.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -107,6 +107,7 @@ export default class JSONSchemaBridge extends Bridge {
     public validator: (
       model: UnknownObject,
     ) => ValidatorResult | null | undefined,
+    private provideDefaultLabelFromFieldName: boolean = true,
```
radekmie, line 110: Same as with the `GraphQLBridge` - it should be an object already.

packages/uniforms-bridge-zod/src/ZodBridge.ts
```diff
@@ -29,7 +29,10 @@ function isNativeEnumValue(value: unknown) {
 }
 
 export default class ZodBridge<T extends ZodRawShape> extends Bridge {
-  constructor(public schema: ZodObject<T>) {
+  constructor(
+    public schema: ZodObject<T>,
+    private provideDefaultLabelFromFieldName: boolean = false,
```
radekmie, line 34: Same here. But maybe it'd make sense to have the schema as the first argument and the options object as the second? I'm not sure.

packages/uniforms/src/BaseForm.tsx
```diff
@@ -21,12 +21,11 @@ export type BaseFormProps<Model extends UnknownObject> = {
...
   error: unknown;
   id?: string;
-  label: boolean;
   model: Model;
   modelTransform?: (mode: ModelTransformMode, model: Model) => Model;
   noValidate: boolean;
-  onChange?: (key: string, value: unknown) => void;
-  onSubmit: (model: Model) => void | Promise<unknown>;
+  onChange?: (key: string, value: any) => void;
+  onSubmit: (model: Model) => void | Promise<any>;
```
radekmie, line 28: Is this from an incorrect conflict resolution or actual change?

packages/uniforms/src/useField.tsx
```diff
@@ -59,25 +36,12 @@ export function useField<
...
-    state.placeholder,
-    label || labelFallback,
-  );
+  const label = props.label ?? schemaProps.label ?? '';
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const id = useMemo(() => context.randomId(), []);
   const onChange = useCallback(
-    (value?: Value, key: string = name) => {
+    (value?: Value | unknown, key: string = name) => {
```
radekmie, line 44: Same as the `any` above.

website/pages-parts/CommonForms/ShippingForm.tsx
```diff
@@ -21,7 +21,6 @@ export function ShippingForm() {
       schema={schema}
       onSubmit={(model: any) => alert(JSON.stringify(model, null, 2))}
       model={model}
-      placeholder
```
radekmie, line 24: We'd like to add them in the schema then.

radekmie (changes requested):
As I understand, as soon as we migrate to Zod in tests _or_ Simple Schema solves the issue with the `label`s, we could remove all of these `label: ''` from tests?

packages/uniforms-bridge-graphql/src/GraphQLBridge.ts
Monteth: Do you want it to look like this?
```ts
constructor(
  public schema: GraphQLType,
  props: {
    validator: (model: UnknownObject) => unknown,
    provideDefaultLabelFromFieldName: boolean
  }
)
```
Or do you want the `schema` to be in the object too?

Monteth: OK, I can see from the next comments that you are not sure; I would consider dividing parameters to 1. `schema` and 2. `others` if the second parameter would be unused in most cases. In this case of Graphql schema, I think the validator will be used in most cases, the same for JSONSchema, so let's change those bridges to accept one object.

Monteth:
> As I understand, as soon as we migrate to Zod in tests or Simple Schema solves the issue with the labels, we could remove all of these label: '' from tests?

Correct. Since we get rid of `label` in context, we can no longer tell SimpleSchema not to derive the label from the field name, so we have to pass it explicitly. This behavior is specific to SimpleSchema only.

website/pages-parts/CommonForms/ShippingForm.tsx
Monteth: I don't know what I am doing wrong, but the form behaves exactly like on [uniforms.tools](https://uniforms.tools/). The only empty fields `state` and `zip` receives the placeholder in the field level, so it seems like we don't need it.

packages/uniforms-bridge-json-schema/src/JSONSchemaBridge.ts
```diff
@@ -287,8 +297,10 @@ export default class JSONSchemaBridge extends Bridge {
       this._compiledSchema[name],
     );
 
-    props.label ??=
-      props.title ?? upperFirst(lowerCase(joinName(null, name).slice(-1)[0]));
+    if (this.provideDefaultLabelFromFieldName && props.label === undefined) {
+      props.label =
+        props.title ?? upperFirst(lowerCase(joinName(null, name).slice(-1)[0]));
+    }
```
radekmie, line 303:
```suggestion
    props.label ??= props.title;
    if (this.provideDefaultLabelFromFieldName && props.label === undefined) {
      props.label = upperFirst(lowerCase(joinName(null, name).slice(-1)[0]));
    }
```

packages/uniforms-bridge-graphql/src/GraphQLBridge.ts
```diff
@@ -20,13 +20,29 @@ function fieldInvariant(name: string, condition: boolean): asserts condition {
...
-    public extras: UnknownObject = {},
-  ) {
+  extras: UnknownObject;
+  provideDefaultLabelFromFieldName: boolean;
+  schema: GraphQLType;
+  validator: (model: UnknownObject) => unknown;
+
+  constructor({
+    extras = {},
+    provideDefaultLabelFromFieldName = false,
```
radekmie, line 30:
```suggestion
    provideDefaultLabelFromFieldName = true,
```

radekmie (dismissed):

packages/uniforms/__tests__/connectField.tsx
```diff
@@ -206,25 +190,28 @@ describe('connectField', () => {
...
-    expect(wrapper.find(Test).at(4).prop('label')).toBe('');
-    expect(wrapper.find(Test).at(5).prop('label')).toBe('');
+    expect(wrapper.find(Test).at(0).prop('label')).toBe('Field');
+    expect(wrapper.find(Test).at(1).prop('label')).toBe('');
+    expect(wrapper.find(Test).at(2).prop('label')).toBe('Field');
+    expect(wrapper.find(Test).at(3).prop('label')).toBe('Other');
+    expect(wrapper.find(Test).at(4).prop('label')).toBe('Subfield');
+    expect(wrapper.find(Test).at(5).prop('label')).toBe('Subfield');
     expect(wrapper.find(Test).at(6).prop('label')).toBe('Subfield');
+    expect(wrapper.find(Test).at(7).prop('label')).toBe('Subfield');
```
kestarumper, line 215: I counted 9 fields but I see only 8 `expect`(s). I was looking for
```ts
expect(wrapper.find(Test).at(8).prop('label')).toBe('');
```

kestarumper (dismissed):

kestarumper (approved):

radekmie (approved):

PR #1258, opened 2023-04-25
Migrate `RadioField` tests to @testing-library/react

ErnestTeluk:
Rewrite RadioField tests to @testing library/react. Also added a helper to antd to support `data-*` attributes.


packages/uniforms-antd/src/RadioField.tsx
```diff
@@ -4,14 +4,33 @@ import { FieldProps, connectField, filterDOMProps } from 'uniforms';
...
 const radioStyle = { display: 'block' };
 
+// helper function needed because antd radio.group does not support data-attributes https://github.com/ant-design/ant-design/issues/8561
+const filteredDataAttributes = (props: ObjectType) =>
+  Object.keys(props)
+    .filter(key => key.startsWith('data-'))
+    .reduce((newProps: ObjectType, key: string) => {
+      newProps[key] = props[key];
+      return newProps;
+    }, {});
```
radekmie, line 32: I gave it a deeper thought, and I think we should rather make the `data-*` tests optional. I don't think it's good that one theme would render them on the wrapper (thus only once) and some on every item (multiple times).

radekmie (changes requested):

packages/uniforms-antd/__tests__/index.ts
```diff
@@ -39,15 +39,17 @@ describe('@RTL', () => {
...
+  suites.testListDelField(theme.ListDelField);
   suites.testListField(theme.ListField, {
     getListAddField: screen => screen.getByRole('img', { name: 'plus-square' }),
   });
   suites.testLongTextField(theme.LongTextField);
   // FIXME: AntD number input doesn't work with new RTL test implementation
   // suites.testNumField(antd.NumField);
   suites.testQuickForm(theme.QuickForm);
+  // FIXME: AntD radio.group does not support data-attributes https://github.com/ant-design/ant-design/issues/8561
+  // suites.testRadioField(theme.RadioField);
```
radekmie, line 52: If some of the `RadioField` tests work, then please use them, e.g., by adding a `ignoreDataAttributes` flag. (IIRC that was the sole problem.)

If not, then restore the previous test suite.

radekmie (changes requested):

radekmie (approved):

Issue #596, opened 2019-09-15
Support recursive json schemas


crubier:
Recursive JSON schemas such as this one (where the recursi e property is not required) make this library crash. It should not, ideally.

```
{
  "definitions": {
    "address": {
      "type": "object",
      "properties": {
        "billing_address": {
          "title": "Billing address",
          "$ref": "#/definitions/address"
        },
        "street_address": {
          "type": "string"
        },
        "city": {
          "type": "string"
        },
        "state": {
          "type": "string"
        }
      },
      "required": [
        "street_address",
        "city",
        "state"
      ]
    }
  },
  "type": "object",
  "properties": {
    "billing_address": {
      "title": "Billing address",
      "$ref": "#/definitions/address"
    }
  }
}
```

radekmie:
Hi @crubier. It's not always possible, as your schema while being used in `<AutoForm />` would create an infinite number of fields (due to `billing_address` recursion). However, we can make the parsing lazy and allow the following code:

```js
<AutoForm schema={schema}>
  <AutoField name="billing_address">
    <AutoField name="billing_address">
      <AutoField name="street_address" />
    </AutoField>
  </AutoField>

  {/* Or equivalently */}
  <AutoField name="billing_address.billing_address.street_address" />
</AutoForm>
```

crubier:
My actual use case is actually not about adresses, and involve an actual need for arbitrary depth recursion. But the key point here is that the recursion does not have to be infinite since the recursive element is NOT required. So by not displaying non-required recursive fields by default (but allowing the user to click to add it), then there is no infinite recursion (unless the user clicks an infinite amount of times)

radekmie:
> [...] since the recursive element is NOT required.

Sure thing.

> [...] not displaying non-required recursive fields by default (but allowing the user to click to add it) [...]

It'll be possible, but not a built-in behaviour. You'll have to implement it on your own.

crubier:
What would be the best way to do this on our side in your opinion?

radekmie:
At first, wait for the implementation on uniforms site or file a PR with it. Then, create a custom `NestField`-like component that you need, so a component which conditionally renders optional subfields.

radekmie:
@crubier I'm not sure if it was working before (haven't tried it) but it is definitely working in v2.6.6.

PR #1282, opened 2023-08-11
Recursive schema evaluation - stop evaluating initial value for empty array

Monteth:
Uniforms go through the schema to get starting values for the following fields. Therefore, during the initial load, it must go through all the 'leaves' of the schema tree. This eager approach prevents us from processing recursive schemas. However, this doesn't have to be the case.

When we think about different types of recursive schemas, they only make sense when they can be completed in a finite amount of time. This requires specific points where the evaluation can stop. An array has a built-in mechanism for this. It can contain one or multiple values to continue expanding the tree, or it can have 0 elements to close off a branch. This enables us to evaluate a recursive schema and create a finite model. In this model, the 'leaves' can consist of empty arrays.

Consider this schema, where we can add multiple children on different levels, and it will only evaluate the next tree level when we add a new value to the model (+ button).
So the schema of the family tree is potentially infinite, but the model is finite.
```JSON
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "children": {
      "type": "array",
      "items": { "$ref": "#" }
    }
  }
}
```
```JS
const model = {
  "children": [
    {
      "name": "Emma",
      "children": [
        {
          "name": "Olivia"
        },
        {
          "name": "Sophia",
          "children": [
            {
              "name": "Zoe"
            }
          ]
        }
      ]
    }
  ],
  "name": "Alice"
};
```


kestarumper (approved):

kestarumper:
We talked internally with @Monteth, and we also want to include a basic cycle detection mechanism to prevent freezing the browser tab (prevent infinite loops).

ErnestTeluk (approved):

Monteth:
> We talked internally with @Monteth, and we also want to include a basic cycle detection mechanism to prevent freezing the browser tab (prevent infinite loops). 

I've tried to implement that, but since it is a little more complicated than we would like, I want to merge this PR without this check and (eventually) implement it under a dedicated issue.

I'd also reconsider if we really need that because I've spotted only one issue related to JSONSchema infinite loop (#596) that was fixed. This PR fixes one more case where it could take place so that it will be even less helpful but still will have an impact on the package footprint and render performance. 

@kestarumper

package.json
```diff
@@ -20,13 +20,16 @@
     "@testing-library/user-event": "14.4.3",
     "@types/jest": "26.0.20",
     "@types/node": "18.19.50",
+    "buffer": "^5.5.0||^6.0.0",
```
piotrpospiech, line 23: Why two different versions? It is dev depenendency, not a peer dependency. Why is this needed? Where is this used?

package.json
```diff
@@ -20,13 +20,16 @@
...
     "eslint-config-vazco": "7.4.0",
     "eslint-import-resolver-alias": "1.1.2",
     "eslint-import-resolver-typescript": "2.3.0",
     "husky": "8.0.1",
     "jest": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
+    "json-schema-traverse": "1.0.0",
```
piotrpospiech, line 30: Why is it added here? It was already added to the json bridge package.

package.json
```diff
@@ -20,13 +20,16 @@
...
     "lint-staged": "13.0.3",
+    "process": "^0.11.10",
```
piotrpospiech, line 32: Is this used?

packages/uniforms-bridge-json-schema/package.json
```diff
@@ -37,6 +37,7 @@
   },
   "dependencies": {
     "invariant": "^2.0.0",
+    "json-schema-traverse": "^1.0.0",
```
piotrpospiech, line 40: Where did you use it?

piotrpospiech (changes requested):
I think you have some leftovers from trying to implement loops detection.

Monteth:
> I think you have some leftovers from trying to implement loop detection.

Yes, indeed. Sorry for that!

piotrpospiech (approved):

zendranm (approved):

pnpm-lock.yaml
```diff
@@ -6732,7 +6732,7 @@ snapshots:
       '@parcel/source-map': 2.1.1
       '@parcel/types': 2.12.0(@parcel/core@2.12.0(@swc/helpers@0.5.13))(@swc/helpers@0.5.13)
       '@parcel/utils': 2.12.0
-      '@parcel/workers': 2.12.0(@parcel/core@2.12.0(@swc/helpers@0.5.13))(@swc/helpers@0.5.13)
+      '@parcel/workers': 2.12.0(@parcel/core@2.12.0(@swc/helpers@0.5.13))
```
piotrpospiech, line 6735: You should remove `pnpm-lock.yaml` from this PR. It looks like changes caused by packages installed previously.

piotrpospiech (changes requested):

piotrpospiech (approved):

Issue #1130, opened 2022-06-10
Migrate tests from Enzyme to React Testing Library


radekmie:
As Enzyme is no longer supported and we already rely on a community-made workaround for React v17 and there's nothing sure for v18 we should push for getting rid of it entirely.

Support for v18 was requested in #1109 and added in #1129.

radekmie:
Maybe we could try https://github.com/enzymejs/enzyme/issues/2524#issuecomment-1117039658 to buy us more time?

ErnestTeluk:
> Maybe we could try [enzymejs/enzyme#2524 (comment)](https://github.com/enzymejs/enzyme/issues/2524#issuecomment-1117039658) to buy us more time?

unfortunately it did not work (tested on react 17), when it comes to react 18 there is one more blocker which is https://github.com/facebook/docusaurus/issues/7264

kestarumper:
- [x] [Migrated AutoField tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1266)
- [x] [Migrated AutoFields tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1271)
- [x] [Migrated BoolField tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1268)
- [ ] ListField
- [ ] ListItemField
- [x] [Migrated NestField tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1270)
- [x] [Migrate RadioField tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1258)
- [ ] [Rewrite SelectField tests](https://github.com/vazco/uniforms/pull/1294)
- [ ] TextField
- [x] [Migrated SubmitField tests to @testing-library/react](https://github.com/vazco/uniforms/pull/1269)

src: https://github.com/orgs/vazco/projects/4/archive

PR #1299, opened 2024-02-09
Migrated `ListField` tests to @testing-library/react

piotrpospiech:

- added new tests to ListField RTL tests
- removed enzyme ListField tests


packages/uniforms/__suites__/ListField.tsx
```diff
@@ -116,4 +183,46 @@ export function testListField(
...
+          'x.$': String,
+        },
+      );
+
+      expect(screen.getByText(/^Error$/)).toBeInTheDocument();
+    });
+  }
+
+  if (disableInlineError) {
+    test('<ListField> - renders correct error text (specified)', async () => {
```
kestarumper, line 209: Is the name of this test OK? It's no different than the above one but the body is different (e.g. `showInlineError={false}`.

kestarumper (changes requested):

kestarumper (approved):

PR #1301, opened 2024-02-14
Migrated `connectField` tests to @testing-library/react

piotrpospiech:

- added `connectField` RTL tests
- removed enzyme `connectField` tests

Main changes compare to `enzyme` tests:
- removed from `when rendered with label` test following `labelVariants`:
```js
const labelA = <span style={{ color: 'red' }}>Error</span>;
const labelB = <span style={{ color: 'green' }}>OK</span>;
    
['Props', labelB, 'Props'],
['', labelB, ''],
[labelA, '', labelA],
[labelA, 'Schema', labelA],
[labelA, labelB, labelA],
[labelA, undefined, labelA],
[undefined, labelB, labelB],
```

- changed in the `is able to set another field value` test, tested value from `object` to `string` to have the same type in those two fields


packages/uniforms/__suites__/connectField.tsx
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  test('connectField - when called, creates component', () => {
+    const Field = connectField(Test);
+
+    expect(Field).toBeInstanceOf(Function);
+    expect(Field.Component).toEqual(Test);
+  });
+
+  test('connectField - when called, does not set default value', () => {
```
kestarumper, line 75: I'm not sure either about the test name or the test body
Does not set default value or default options? (default value of options maybe?)

I'm asking because the tests below (the ones with the `{ kind: '...' }` are named differently, but the test bodies are similar.

kestarumper (changes requested):

kestarumper (dismissed):

packages/uniforms/__tests__/connectField.tsx
```diff
@@ -119,100 +132,111 @@ describe('connectField', () => {
...
-      ['', labelB, ''],
       ['', undefined, ''],
-      [labelA, '', labelA],
-      [labelA, 'Schema', labelA],
-      [labelA, labelB, labelA],
-      [labelA, undefined, labelA],
       [undefined, '', ''],
       [undefined, 'Schema', 'Schema'],
-      [undefined, labelB, labelB],
       [undefined, undefined, ''],
```
kestarumper, line 157: Why did you remove those cases?

kestarumper (changes requested):

packages/uniforms/__tests__/connectField.tsx
piotrpospiech: I created `Test` component that use `label` as `input` props and it couldn't be rendered.

I changed it now to render `label` above `input`. Rendering `label` is based on rendering `label` in `TextField` in the `uniforms-unstyled`. I needed to change `works with nested labels` test after this change.

kestarumper (approved):
Great work!

PR #1303, opened 2024-02-15
Migrated `BaseForm` tests to @testing-library/react

piotrpospiech:
Is part of #1130

- added new tests to `BaseForm` RTL suites
- removed enzyme `BaseForm` tests

I cannot migrate most of `BaseForm` tests, because it was based on `BaseForm` context or state. RTL doesn't support testing these cases. I didn't find another aproach to test these cases.

Additional issue is with `ignores synchronous errors` test. Using RTL, error cannot be catched.


packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -1,387 +1,76 @@
...
-
-    it('autosaves correctly (`autosave` = true)', async () => {
-      wrapper.setProps({ autosave: true });
-      wrapper.instance().getContext().onChange('a', 1);
-      await new Promise(resolve => setTimeout(resolve));
-      const context = wrapper.instance().getContext();
-      expect(onSubmit).toHaveBeenCalledTimes(1);
-      expect(context.submitted).toBe(true);
-      expect(onSubmit).toHaveBeenLastCalledWith(model);
-    });
```
kestarumper, line 175: We could rewrite it with `fireEvent` and check how many times did `onSubmit` was called.

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -1,387 +1,76 @@
...
-
-    it('clears autosave correctly', () => {
-      wrapper.setProps({ autosave: true, autosaveDelay: 100 });
-      wrapper.instance().getContext().onChange('a', 1);
-      wrapper.unmount();
-
-      expect(onSubmit).not.toBeCalled();
-    });
```
kestarumper, line 225: That one is very important when it comes to the functionality.
https://testing-library.com/docs/react-testing-library/api/#unmount

packages/uniforms/__tests__/BaseForm.tsx
```diff
@@ -1,387 +1,76 @@
...
-      wrapper.setProps({ modelTransform: undefined });
-    });
-
-    it('sets `submitted` to true', async () => {
-      const wrapper = createWrapper();
-      const instance = wrapper.instance();
-      expect(instance.getContext().submitted).toBe(false);
-      wrapper.find('form').simulate('submit');
-      expect(instance.getContext().submitted).toBe(true);
-    });
```
kestarumper, line 321: Idea: Prepare a component that reads `context.submitted` and renders conditionally.
Then we can `expect(Component).toHaveBeenCalledTimes(1)`.

kestarumper (changes requested):
I don't know about the other `it(...)`s but we definitely need to think that through.

I'm not sure how much we can count on the CodeCov code coverage.
If for example the `autosave` functionality is covered somewhere else (e.g. in a theme package) then that's fine, but we can't lose the tested functionalities at all.

kestarumper (approved):
Nice!
I haven't come up with any other tests so if nobody suggests anything we're good to go.

PR #1318, opened 2024-04-19
Migrated `wrapField` tests to @testing-library/react

zaxoavoki:


packages/uniforms/__suites__/wrapField.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+import { renderWithZod } from './render-zod';
+import { skipTestIf } from './skipTestIf';
+
+export function testWrapField(
+  wrapField: (wrapperProps: any, children: ReactNode) => ReactElement,
+  options?: {
+    skipForMUI?: boolean;
+    skipForAntD?: boolean;
+    onlyForBootstrap3?: boolean;
```
piotrpospiech, line 13: I have trouble reading this PR if you use variables like `skip...` and `only...`. In this suite, sometimes you use `options?.skipForAntD` and `!options?.skipForAntD`.

It will be more readable if you stick to one convention. However, I don't think this is possible in this single file if we want to cover all tests.

packages/uniforms/__suites__/wrapField.tsx
```diff
@@ -0,0 +1,249 @@
...
+          { wrapperStyle: { backgroundColor: 'red' } },
+          <div data-testid="x" />,
+        ),
+        schema: z.object({}),
+      });
+      expect(
+        screen
+          .getByTestId('x')
+          .closest('.ant-form-item')
+          ?.getAttribute('style'),
```
piotrpospiech, line 175: If you want to use class selector, you can do this:
```ts
const { container } = renderWithZod(...)
const element = container.getElementsByClassName('.ant-form-item')[0]
```

packages/uniforms/__suites__/wrapField.tsx
```diff
@@ -0,0 +1,249 @@
...
+        element: wrapField({ extra: 'Extra' }, <div data-testid="x" />),
+        schema: z.object({}),
+      });
+      if (options?.skipForAntD) {
+        expect(screen.getByText('Extra')).toBeInTheDocument();
+      } else {
+        expect(screen.getByTestId('x').parentElement).toHaveAttribute('extra');
+      }
+    },
+  );
```
piotrpospiech, line 193: I think there are too many conditions here.

piotrpospiech (commented):
This PR really points out the disadvantage of the suites approach (the same as `SelectField` suites, I guess)

1. We talked we don't want to create flags for specific themes, but I see why it is needed here.
2. Creating tests only for one theme in the suite or changing the `expect` check for the specific theme only makes this suite harder to read.

I think we need to consider keeping theme-specific tests in a separate file. We could keep core tests in the suites and keep theme specific in the theme package or create a separate file (we can have `wrapField` as a directory). The distinction can be necessary as we support different props, e.g., `validateStatus` in the AntD theme.

The same issue is true for the rest of the tests that are waiting to be moved to suites (mainly AntD and MUI).

packages/uniforms/__suites__/wrapField.tsx
zaxoavoki: I fully agree with you that it's unreadable and not elastic at all. Having more component features will require us to add other flags to test theme-specific functionalities somehow. That's shy I'd move all the unique tests for each theme into their own folders and keep core tests in suites. That way, it's easier to see what's what, and we won't run into any problems down the line. But if we stick with using flags for each theme, getting rid of a theme later on is gonna be a real headache. We'd have to spend ages sorting out flags and fixing things up. Instead, we could just remove the test file for that theme in its folder and be done with it.

What do you guys think?

piotrpospiech: I checked the suite file again, taking the test structure we discussed above into account.

* `renders wrapper with (feedbackable=true)` -> We deprecate bootstrap3. I think we can remove this test. I will ignore deprecated themes in the next tests from now.
* `renders wrapper` -> it’s tested only for material UI and MUI. Let’s move it to the MUI theme.
* `renders wrapper with correct class` -> it’s bootstrap specific test. We will support two versions of Bootstrap. We can pass the `wrapClassName` argument and write this test in the suite, but we can also move it to bootsrap4 and bootstrap5 as we decided to remove bootstrap4 in the next year. It will be easier to maintain in the future.
* `renders help block` -> use argument `helpPropsName` with default value - `help`
* `renders help block with specified class` -> the same as `renders wrapper with correct class`
* `renders error block` -> I think it should be named `renders error block (showInlineError=true)` as we have a similar test later in this suite.
* `renders wrapper with label` -> it looks like Material UI and MUI don’t support labels. Let’s use the argument `withoutLabel`.
* `label has custom class (String)` -> the same as `renders wrapper with correct class`
* `label has custom class (Array[String])` -> the same as `renders wrapper with correct class`
* `renders wrapper with a custom validateStatus` -> move it to AntD theme
* `renders wrapper with extra style` -> move it to AntD theme
* `renders wrapper with extra text` -> `extra` is supported props only in the AntD theme. You shouldn’t check this: `expect(screen.getByTestId('x').parentElement).toHaveAttribute('extra');`. Let’s move it to the AntD theme.
* `renders wrapper with an error status (error)` -> do we need to test it as we have a similar test for AntD (`renders error block`)?
* `renders error block (showInlineError=false)` -> is there any way to write this test to make it more general?
* `renders wrapper with label and info` -> move it to AntD.

piotrpospiech (approved):

kestarumper (approved):

PR #1319, opened 2024-04-20
Add support for Zod `.refine` and `.superRefine`

piotrpospiech:
This allows to pass to the `ZodBridge` schema that has `refine` or `superRefine`.

**Example:**

```ts
const schema = z
  .object({
    username: z.string().min(4),
    password: z.string().min(8),
    confirmPassword: z.string().min(8),
  })
  .refine((form) => form.password === form.confirmPassword, {
    message: "Passwords don't match",
    path: ["confirmPassword"],
  });
```

```ts
const schema = z
  .object({
    username: z.string().min(4),
    password: z.string().min(8),
    confirmPassword: z.string().min(8),
  })
  .superRefine((val, ctx) => {
    if (val.password !== val.confirmPassword) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message: "Passwords don't match",
        path: ["confirmPassword"],
      });
    }
  });
``` 

image


kestarumper (changes requested):
Could you add a simple test that creates the `ZodBridge` from a `z.object(...).refine()` and `z.object(...).superRefine()`?

kestarumper (approved):

zaxoavoki (approved):

zendranm (approved):

Monteth (approved):

Issue #1315, opened 2024-04-12
Deprecate packages


kestarumper:
## Upcoming deprecation
We plan on deprecating the following packages in the upcoming `v4` release:
- `uniforms-material` #1338
- `uniforms-bootstrap3` #1341
- `uniforms-bridge-graphql` #1322
- `uniforms-bridge-simple-schema` #1323

This means we will mark the packages as deprecated on NPM and remove them from the repository. Those pacakges can still be downloaded and used but we will not contribute to them.

Migration guide  _to uniforms v4 with deprecated themes_  will be included in the uniforms v4 documentation #1305.

## Planned deprecation
- `uniforms-bootstrap4` will be deprecated after a year.

## Motivation
We will deprecate theme packages in a year after releasing the next version of the theme package that supports the new major version of the UI theme package.

E.g.
1. Bootstrap releases `Bootstrap 6`
2. We start working on `uniforms-bootstrap6.`
3. We release `uniforms-bootstrap6` and  then deprecate `uniforms-bootstrap5` in a year

Maintaining old packages with backward compatibility is our main problem and does not scale well. We want to stay up to date with the newest theme integrations.

When it comes to the bridges, we identified the least popular ones and want to pay more attention to the other ones. There is a plan to focus on improving `uniforms-bridge-zod`.

PR #1322, opened 2024-04-20
Deprecate `uniforms-bridge-graphql` package

piotrpospiech:
Deprecate the GraphQL bridge, as mentioned in this issue (https://github.com/vazco/uniforms/issues/1315).

- Removed the whole package from the `uniforms` repository
- Removed package from labeler
- Removed GraphQL from the Readme file
- Removed `graphql` and `uniforms-bridge-graphql` packages from `package.json` files
- Removed GraphQL example from docs (playground)
- Removed bridge from tsconfig references

> [!NOTE]
> I left the `GraphQLBridge` docs. We will keep that in the uniforms v3 docs.


kestarumper (changes requested):
If we leave the docs about the GraphQL bridge, shouldn't we add a warning that the package has been deprecated? I suggest using [admonition](https://docusaurus.io/docs/next/markdown-features/admonitions).

kestarumper (approved):

zendranm (approved):

Monteth (approved):

Issue #1194, opened 2022-12-01
Compatibility with StrictMode


Akshaya-009:
After updating uniforms and uniforms-unstyled to version `3.10.1`,  unable to see the values in the input field.

Please check the issue and update us on the same.

Attaching Codesandbox link for the reference: 

[Click to open codesanbox](https://codesandbox.io/s/uniform-tools-cbppr4?file=/src/App.js)

wadamek65:
Hi @Akshaya-009 , thank you for the issue report.

The issue you are seeing is not caused by `uniforms@3.10.1` update directly, but rather by the version of `react` being 18. The issue seems to go away if you remove `StrictMode` from the application, so there may be some idempotency problem within core `uniforms` logic. We will investigate this further and try to resolve it, but depending on the outcome, the fix may be too complex and may have to be postponed to `uniforms@4`. We will keep you updated on this issue.

sprasad02:
@wadamek65 Removing `StrictMode` works. Until the fix is released can this be added to the doc so that it might be helpful to other devs upgrading to React 18?

Akshaya-009:
Thank you for the reply @wadamek65 and @sprasad02. Please update once it is fixed in `version 4`.

maslennikov:
I confirm, with Nextjs 13 and strict mode enabled fields don't get updated when typing (although `onChange` callbacks trigger and model gets updated with setstate properly). Solution from #1289 doesn't work for me. Had to disable strict mode.

PR #1324, opened 2024-04-20
Compatibility with StrictMode

piotrpospiech:
This fixes https://github.com/vazco/uniforms/issues/1194.

Overriding `setState` caused this issue. I removed it and added checking if the component is mounted before calling `setState`.

Tested with Vite, React `v18.2.0` and Strict Mode.


kestarumper (approved):
Have you tested it somehow?

zendranm (approved):

mdesousa:
hi @piotrpospiech @zendranm thanks for this fix! wondering if there are plans to patch it in version 3?

piotrpospiech:
Hi @mdesousa, we didn't consider it. Our focus is on version 4, which is nearly finished ([v4.0 Milestone](https://github.com/vazco/uniforms/milestone/8)) and mainly needs documentation updates. We don't plan more changes to v3 packages. If this fix is important for you, I'd recommend trying the v4 beta or waiting for the full v4 release coming soon.

mdesousa:
Unfortunately our app is in production and can't upgrade to a beta library yet. We'd also expect that there will be breaking changes and more testing necessary to upgrade to 4.0... which we'll happily do when it's ready. But in the meantime getting this small patch for version 3 would help us with this issue that is impacting our production app. Thanks!

PR #1326, opened 2024-04-22
Add CodecovCI token

Monteth:
For some time, the Codecov CI has required a token even for OSS projects, so it has been added to this PR. Also, other actions are upgraded to versions running on node 20, as previous runners are deprecated.


.github/workflows/CI.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         node-version: [14.x, 16.x, 18.x, 20.x]
     steps:
       - name: Checkout
-        uses: actions/checkout@v3.3.0
+        uses: actions/checkout@v4.1.1
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3.6.0
```
piotrpospiech, line 18: Did you consider upgrading `actions/setup-node`?

Monteth: Oh, I must have skipped that by accident, thanks for pointing it out!

kestarumper (approved):

zendranm (approved):

PR #1338, opened 2024-05-24
Deprecate `uniforms-material` theme

zendranm:
Deprecating the Material theme, as mentioned in this issue (https://github.com/vazco/uniforms/issues/1315).

- Removed the whole package from the uniforms repository
- Removed package from the labeler
- Removed Material from the Readme file
- Removed the uniforms-material package from the package.json files
- Removed theme from tsconfig references
- Removed `skipTestIf` from test suites


kestarumper (approved):

kestarumper:
I wonder if we should add somewhere in the docs an admonition that this theme has been deprecated.
But I don't know where since there are no dedicated pages for a theme.

wadamek65:
@kestarumper Maybe you could add it in the installation section when picking the material theme? It's probably very niche but better than nothing I suppose. 
image

kestarumper:
@wadamek65, but we're getting rid of the Material tab button fully (this is because we're deleting all dependencies such as styles, etc).

wadamek65:
@kestarumper Then another option would be modifying the README in the package itself, so at least it will be shown on the npm page of the package.

wadamek65:
@kestarumper Then I'm out of ideas 😖

Monteth:
@kestarumper We can release the last version with deprecation info and then remove the package, so in the history, there will be a readme with a message.

Monteth (approved):

ErnestTeluk (dismissed):

ErnestTeluk:
conflict to fix

ErnestTeluk (approved):

kestarumper (approved):

PR #1343, opened 2024-05-29
Initialize model with default values

### LINKED ISSUE #1066: Optional defaultValue does not populate form model, but field only
As you can see in this reproduction: https://codesandbox.io/s/jolly-perlman-ntuxi

Both fields in the form are filled with their defaultValue, yet only str1 is set in the form model. This causes weird bugs as we rely on the model being exactly what is shown on screen.

This is the schema in question:

```js
const schema = new SimpleSchema({
  string1: {
    type: String,
    defaultValue: "yooo" // Shown in field and form model
  },
  string2: {
    type: String,
    defaultValue: "yooo", // Shown in field only
    optional: true
  }
});
```

kestarumper:
Closes #1066.

## Changes
- Model is initialized with the `initialValues` taken from `schema` (`Bridge`) - 
- Model results from a shallow merge of `schema.model` and `props.model` and the latter one takes precedence
- (`__tests__` only) `render(...)` changed wrapper from `<context.Provider>` to `<AutoForm>` because it changed behavior of `<ListField>` (the list of child fields was incorrect)

## Notes
- **BREAKING:** This changes the initial render and doesn't trigger the `Field.onChange` of the fields that set their `initialValue` on the first render. However, the fields that are rendered after, e.g., after adding another field in an `Array`, or due to conditional rendering, will still trigger `Field.onChange`.
- **BREAKING:** Optional / Required doesn't affect setting the model

## Example
- [Playground (localhost)](http://localhost:3000/playground/#?N4IgDgTgpgzlAuIBcICCAbdACAYgSynQBMYsAKAZTwFsx0oKBjACymoEMBKEAGnAgD2YGMlDsArvAEx2ANyjIAZu3Rw-EqTPkARQuwCeyAIwAGE3yJ4ZAI3pElKtSHTtrhZPAjiofOu0ZQzALEUBAOqj4g0OxEAPIAduiGSMoRfDBBAO4Akol48VAAohCCYSmOkeww-vGMCRTi1tR4iOVpIFU1dfEAaip4ROzwCm1OMCxs7MggBZlYVLT0TKwcAEwAQhADAOZQZMAAOvFYWOMr7EhYs_M0dAwTHPtHJzCe-dtGl4fHJ1jw-mAoJcKG94tseM9fkQoMpxOh4H10N5LgcQPoBBjUZCAL4Qn6vLZg1ZfSEnf6A4Gg8GkrDQ2HwxHIrCo9EY1aovG_LBCeB4ATxFSXTzeHGcrCKLZQeIkL5_AFArCoEoGHi0mESBkqJkAbWAV3Y1AVqIAUgJmMdtAIoBysOxdpcACyrLDYgC6LrFAHIJQRpTAAHQAEk9l1i1gAVlBGPAvT6pSQg_6BYaQ_MqbHJX7E3aoKm9eSFQA5cTUNwQVV0jUIrUKgCctY9kO9mYTgf9MSI0BgMFTYcj0Yzvtb7aIndgAYJUAQeblFLThPBavp1aRCs9FFgBtzjZ-zaHAbbylkAi2w2LpdCPdlBcuSogKqXVcZCu1qyMqoAzAB2d24ptxrNDzkE8Wigc8ywPGcbywcDQh3bFOCOBCQFxEB4BWEYOnieB7GxIA)
```
// SimpleSchema2Bridge
{
  string1: {
    type: String,
    defaultValue: "yooo"
  },
  string2: {
    type: String,
    defaultValue: "yooo2",
    optional: true
  },
  friends: { type: Array, defaultValue: [{ name: "John Doe", age: 42 }] },
  'friends.$': Object,
  'friends.$.name': String,
  'friends.$.age': { type: Number, defaultValue: 99 },
  'friends.$.address': Object,
  'friends.$.address.street': { type: String, defaultValue: 'Sesame' },
  'friends.$.favoriteNumbers': { type: Array, defaultValue: [21, 37] },
  'friends.$.favoriteNumbers.$': { type: Number },
}
```


kestarumper:
TODO: add test to check if it works correctly with optional fields added during runtime

packages/uniforms/__suites__/render.tsx
```diff
@@ -1,43 +1,30 @@
...
+import { Context, UnknownObject } from 'uniforms';
 import { SimpleSchema2Bridge } from 'uniforms-bridge-simple-schema-2';
+import { AutoForm } from 'uniforms-unstyled';
 
-const randomId = randomIds();
 export function render<P, Model extends UnknownObject>(
   element: ReactElement<P>,
   schema?: SimpleSchemaDefinition,
-  contextValueExtension?: Partial<Context<Model>>,
+  contextValueExtension?: Pick<Partial<Context<Model>>, 'onChange'>,
```
piotrpospiech, line 11: `contextValueExtension` is no longer needed. It is used only for the `onChange` function. Just create the `onChange` param.

The whole `render` could also be refactored. `schema`, `onChange` and `model` are optional. They can be moved to an object param. We could avoid these kind of uses:

```
render(
  <ListField name="x">
    <Child />
    PlainText
  </ListField>,
  { x: Array, 'x.$': String },
  undefined,
  { x: [undefined, undefined] },
);
```

However, this is a big refactor because we use `render` many times, so maybe we can do it in another PR.

piotrpospiech (changes requested):
Playground link doesn't work for me (I installed dependencies, built everything and started doc website). I've got this error:
```
TypeError: Cannot read properties of undefined (reading 'length')
```

kestarumper:
@piotrpospiech 
> Playground link doesn't work for me (I installed dependencies, built everything and started doc website). I've got this error:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'length')
> ```

Works for me.
I run `npm run reset && npm ci` (_node v20.10.0_).

packages/uniforms/__suites__/render.tsx
```diff
@@ -1,43 +1,30 @@
...
-          schema: new SimpleSchema2Bridge({ schema: new SimpleSchema(schema) }),
-          state: {
-            disabled: false,
-            readOnly: false,
-            showInlineError: false,
-            ...contextValueExtension?.state,
-          },
-          formRef: {} as BaseForm<UnknownObject>,
-        };
+        const bridge = new SimpleSchema2Bridge({
```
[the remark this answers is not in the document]
kestarumper: Good call. It also showed some flaws in our tests. I fixed them.

packages/uniforms/__suites__/render.tsx
kestarumper: Yep, it is too much for this PR (for me). But if you want to refactor, you can commit to this PR.

piotrpospiech:
@kestarumper I tried it with `Node v20.10` and `npm 10.2.5`. I ran `npm run reset && npm ci`. Still the same issue:
```
TypeError: Cannot read properties of undefined (reading 'length')
```

However, it is working for @Monteth.

packages/uniforms/__suites__/render.tsx
piotrpospiech: Let's do this in another PR. Maybe just write `TODO` comment.

piotrpospiech (approved):

ErnestTeluk (approved):

packages/uniforms/src/Bridge.ts
```diff
@@ -84,6 +85,19 @@ export abstract class Bridge {
     );
   }
 
+  /**
+   * Get initial model value recursively.
```
Monteth, line 89:
```suggestion
   * Get the initial model value walking the tree recursively.
```
Maybe it is nitpicking, but you do not recursively run the `getInitialModel` function.

Monteth (approved):
It is a great change. I'd like to see what is the impact on first render time 🤔

PR #1363, opened 2024-08-23
Extend `onChangeModel` - add details about what changed

### LINKED ISSUE #1359: More complex onChange
Hi, we want to implement undo/redo feature. For that we need to know complete model on every change (accessible by `onChangeModel`). But we also need to know which key caused the change. Otherwise we would record every typed letter in an input, which is not how undo/redo should work. To avoid capturing subsequent changes in same field, we can use method `onChange` which provides `key` causing change with new `value`, but now we don't know whole model, so we cannot record it. Now we can only somehow combine key with current model, which causes troubles and bugs especially with nested fields.

Current solution, which I sense is also bug-prone and not quite clean is this:
```js

const handleFieldChange = (key, value) => {
        // This fires after all form.setState() is done
        // so we can catch last model, while knowing which key caused change
        formRef.current.setState(
                () => null,
                () => log("model change", key, value, formRef.current.state.model)
        );
};

return <AutoForm model={model} ref={formRef} onChange={handleFieldChange} />
```

So can we extend onModelChange feature to provide not only new model, but also key and value causing changes?
Thanks.

kestarumper:
Closes #1359.

This PR extends `AutoForm.onChangeModel` to provide additional information `details: { key, value, previousValue }`.
- `key` is the changed key
- `value` is the new value
- `previousValue` is the value before the change

### DIFF
--- a/docs/api-forms.md
+++ b/docs/api-forms.md
@@ -25,9 +25,9 @@ By default, the validation will take place `onSubmit`, and `onChange` **after th
 
 ##### Props:
 
-|      Name       |                                                          Description                                                           |
-| :-------------: | :----------------------------------------------------------------------------------------------------------------------------: |
-| `onChangeModel` | Like `onChange` but for the whole model. Triggered just after `onChange` but with the next model instead of (key, value) pair. |
+|      Name       |                                                                                                              Description                                                                                                               |
+| :-------------: | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
+| `onChangeModel` | Like `onChange` but for the whole model. Triggered just after `onChange` with the next model and information what `{ key, value, previousValue }` caused the change. `previousValue` will be `undefined` if there was no value before. |
 
 **Note:** All `ValidatedQuickForm` props are also accepted and all methods are available.
 In other words, that means that `AutoForm` receives all props listed on this page.
--- a/packages/uniforms/src/AutoForm.tsx
+++ b/packages/uniforms/src/AutoForm.tsx
@@ -1,4 +1,5 @@
 import clone from 'lodash/clone';
+import get from 'lodash/get';
 import isEqual from 'lodash/isEqual';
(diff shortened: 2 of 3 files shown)


zendranm (approved):

packages/uniforms/src/AutoForm.tsx
```diff
@@ -15,7 +16,10 @@ import { ModelTransformMode, UnknownObject } from './types';
 
 export type AutoFormProps<Model extends UnknownObject> =
   ValidatedQuickFormProps<Model> & {
-    onChangeModel?: (model: Model) => void;
+    onChangeModel?: (
+      model: Model,
+      info: { key: string; value: unknown; previousValue: unknown },
```
Monteth, line 21: I don't like the `info` key, what do you think about `context`, `changeContext`, `details`, `changeDetails`, `meta`, `metadata`
```suggestion
      changeContext: { key: string; value: unknown; previousValue: unknown },
```

docs/api-forms.md
```diff
@@ -25,9 +25,9 @@ By default, the validation will take place `onSubmit`, and `onChange` **after th
...
-| :-------------: | :----------------------------------------------------------------------------------------------------------------------------: |
-| `onChangeModel` | Like `onChange` but for the whole model. Triggered just after `onChange` but with the next model instead of (key, value) pair. |
+|      Name       |                                                                                                              Description                                                                                                               |
+| :-------------: | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
+| `onChangeModel` | Like `onChange` but for the whole model. Triggered just after `onChange` with the next model and information what `{ key, value, previousValue }` caused the change. `previousValue` will be `undefined` if there was no value before. |
```
Monteth, line 30: `onChangeModel` does what `onChange` does, and since there is virtually no additional cost, there are no benefits to using `onChange` now.

Monteth (approved):

PR #1366, opened 2024-09-06
Nextra docs

zaxovaiko:
## Changes

Added [`Nextra` docs](https://nextra.site/) to a separate folder. Later, it will be moved into Turborepo. 

Please note that not all interactive components were moved. It will be fixed soon.


zaxovaiko:
Closing PR, since we decided to switch to Astro docs because of a mismatch version error of the React package.

PR #1367, opened 2024-09-08
Replace `lerna` with `turborepo`

piotrpospiech:
### Description

This is minimal setup of `turborepo`. It replaces `lerna` package. I focused on keeping it as simple as possible, so current configuration can be (and should be) updated in the future. For example, `eslint` and `typescript` configs can be moved to internal packages ([docs](https://turbo.build/repo/docs/guides/tools/eslint)). 

During implementation, I had issues with `reproductions` and `website` projects and couldn't make them work. They have been removed for now in order to create this *draft* pull request. Removed files in the `reproductions`, `website` and `docs` can be ignored in the code review process.

### Build process

The biggest change is a new building process. To take full advantage of the new project structure, I changed the build process from incrementally building with `tsc` in the root directory to running each packages' `tsc` build script in parallel. It's faster on cold start, but also incredibly fast when it comes to cache utilization. There are results after running it with Node `v22.8.0` and npm `v10.8.3` (cold start):

| | master | turborepo |
| - | :-: | :-: |
| `build` | 49s | 21s |
| `test` | 37s | 14s |

### Missing features (to be implemented)

- [X] `reproductions` project
- [ ] Documentation (`website` & `docs` directories)
- [ ] Publishing process

`lerna` had already implemented publishing process (package versioning and publishing them to the NPM registry). `turborepo` doesn't have such feature. It is recommended to use [`changesets`](https://github.com/changesets/changesets) ([docs](https://turbo.build/repo/docs/guides/publishing-libraries#versioning-and-publishing)).

I suggest to implement docs and publishing process in other PRs.

### List of changes

- Removed `lerna` package
- Added `turbo` package
- Changed build process
- Removed reproductions and docs (temporary)
- Added `devDependencies` and moved some packages from `dependencies`. It can be refactored further though.
- Simplified `tsconfig` files to remove incremental build
- Added `simple-schema.d.ts` to the `uniforms-bridge-simple-schema-2`. This file was originally located in the `website` directory. This directory was temporary removed and missing file was causing type issues during testing process.
- Removed `publish` script as it was based on `lerna`
- Removed unused packages in the root directory

### How to test changes

Use Node `v22` (check current version in the `.nvmrc`) and npm `v10` (check current version in the `package.json -> packageManager`)

1. Checkout to `turborepo` branch
2. Run `npm run clean` (it deletes packages' builds)
3. Run `npm run reset` (it deletes `node_modules`, turbo cache, etc)
4. Run `npm ci` (it also runs `build` script)
[...]

PR #1368, opened 2024-09-08
Update `typescript`

piotrpospiech:
In this PR, I wanted to update `typescript` package in each `package.json` file from `v4.4.4` to `v5.5.4`. Newer version of `typescript` is required to create extensible `tsconfig` structure. In current version, `extends` option in the `tsconfig` accepts only `string` as argument, but `array` is needed.

To update `typescript` in the root directory, it was also necessary to update the following packages:
- `eslint-config-vazco`
- `jest` (newer version required also to install `jest-environment-jsdom`)
- `ts-jest`

Most of the changes are lint/type fixes caused by updated `eslint-config-vazco` and `typescript` packages.

I updated `JSONSchemaBridge › #getProps › works with anyOf for a non-object computed property (properties not defined)` test as `properties` is not part of tested object. Tested object didn't change in comparison to `master` branch.

I deprecate also Node `v14` in CI workflow and added Node `v22`.


zendranm:
It would be good to update the `CONTRIBUTING.md` content as well.

zendranm (approved):

piotrpospiech:
> It would be good to update the `CONTRIBUTING.md` content as well.

I updated it in the #1367

kestarumper (approved):

zendranm (approved):

zaxoavoki (approved):

PR #1375, opened 2024-10-25
Update react 18 antd5

kestarumper:
Based on the `update-react` branch. Update React to v18 in the `uniforms-antd` package. Updating AntD to v5 is required.

### DIFF
--- a/packages/uniforms-antd/package.json
+++ b/packages/uniforms-antd/package.json
@@ -32,8 +32,8 @@
     "src/*.tsx"
   ],
   "peerDependencies": {
-    "@ant-design/icons": "^4.0.0",
-    "antd": "^4.0.0",
+    "@ant-design/icons": "^5.0.0",
+    "antd": "^5.0.0",
     "react": "^18.0.0 || ^17.0.0 || ^16.8.0"
   },
   "dependencies": {
--- a/packages/uniforms-antd/package.json
+++ b/packages/uniforms-antd/package.json
@@ -46,7 +46,7 @@
   "devDependencies": {
     "@types/invariant": "2.2.37",
     "@types/lodash": "4.17.5",
-    "@types/react": "17.0.39",
+    "@types/react": "18.3.12",
     "@types/simpl-schema": "1.12.8",
     "moment": "2.30.1",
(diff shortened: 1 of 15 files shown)


packages/uniforms-antd/package.json
```diff
@@ -32,8 +32,10 @@
     "src/*.tsx"
   ],
   "peerDependencies": {
-    "@ant-design/icons": "^4.0.0",
-    "antd": "^4.0.0",
+    "@ant-design/icons": "^5.0.0",
+    "antd": "^5.0.0",
+    "dayjs": "^1.11.13",
+    "rc-input-number": "^9.2.0",
```
kestarumper, line 38: Fixes
```
The inferred type of 'default' cannot be named without a reference to '.pnpm/rc-input-number@9.2.0_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/rc-input-number'. This is likely not portable. A type annotation is necessary.ts(2742)
```

piotrpospiech: Let's do this differently. 

1. Keep `rc-input-number@9.2.0` only in `devDependencies`
2. Add `/// <reference types="rc-input-number" />` to the `NumField` file

packages/uniforms-antd/package.json
```diff
@@ -32,8 +32,10 @@
     "src/*.tsx"
   ],
   "peerDependencies": {
-    "@ant-design/icons": "^4.0.0",
-    "antd": "^4.0.0",
+    "@ant-design/icons": "^5.0.0",
+    "antd": "^5.0.0",
+    "dayjs": "^1.11.13",
```
piotrpospiech, line 37: I think we can consider keeping only `antd` and `react` in `peerDependencies` for now.

piotrpospiech (changes requested):
1. AntD tests are skipped. You need to remove `.skip` from them. 5 AntD tests fail. 
2. By the way, our CI fails, so we need to adjust other tests as well. EDIT: This is caused by using `npm` by CI. Let's fix that at the end.

piotrpospiech (approved):

PR #1395, opened 2025-01-17
Added migration guide from 3 to 4 to docs

piotrpospiech:
- added migrating v3 to v4 page
- updated order of pages in getting started section

### DIFF
--- a/website/docs/getting-started/migrating-3-to-4.md
+++ b/website/docs/getting-started/migrating-3-to-4.md
@@ -1,0 +1,36 @@
+---
+id: migrating-4-to-4
+title: Migrating v3 to v4
+sidebar_position: 4
+---
+
+This guide is designed to help you through the migration. If you went through it and encountered any problems - do let us know. For more information on why certain changes were made, see the [CHANGELOG.md](https://github.com/vazco/uniforms/blob/master/CHANGELOG.md). When migrating to v4, use the newest version. Gradual updates will take more time and won't ease this process.
+
+## Breaking API changes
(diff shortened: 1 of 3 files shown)


zaxovaiko (approved):

website/docs/getting-started/migrating-3-to-4.md
```diff
@@ -0,0 +1,36 @@
...
+## Deprecated packages
+
+The following packages are deprecated and are no longer supported in v4.0.0.
+
+- Deprecate `uniforms-bootstrap3` package
+- Deprecate `uniforms-material` package
+- Deprecate `uniforms-bridge-simple-schema` package
+- Deprecate `uniforms-bridge-graphql` package
+
+If you want to continue using these packages, you can use the following commands to download the latest version of the package:
```
Monteth, line 32: We recommend using identical versions of all uniforms packages, so let's also point it out here so they won't use uniforms v4 with uniforms-material v3.

```suggestion
If you want to continue using these packages, we recommend staying with uniforms v3, as we don't guarantee they will work with the newer versions. You can use the following commands to download the latest version of the package:
```

Monteth (approved):

zendranm (approved):

website/docs/getting-started/migrating-3-to-4.md
kestarumper: For me the _latest version_ is referring to `4.x`, not the latest `3.10.x`.

PR #1397, opened 2025-02-26
Update @testing-library/react

piotrpospiech:
I updated testing library to the latest version as React 19 isn't supported in the current version, so it cannot be tested properly.

- updated `@testing-library/react` package
- added `@testing-library/dom` package (required by breaking change in the [v16.0.0 release](https://github.com/testing-library/react-testing-library/releases))

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,9 @@
     "test": "jest"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


Monteth (approved):
The lockfile is outdated, lests refresh it before merge

Issue #1403, opened 2025-03-21
Update `simpl-schema` package


piotrpospiech:
The latest version of the `simpl-schema` package is `3.4.6`. In our `peerDependencies`, we currently only support versions `^1.0.0 || ^0.0.4`. We need to update our `peerDependencies` in the `uniforms-bridge-simple-schema-2` to support newer versions.

Issue created based on the comment in the closed issue:
> Sorry to raise the dead but if we really dont care what version of simpl-schema is there, couldnt we just use something like: 
"simpl-schema": ">=1.0.0",

_Originally posted by @bmartin1134 in https://github.com/vazco/uniforms/issues/298#issuecomment-2741922044_

PR #1404, opened 2025-03-21
Support higher version of peer dependency `simple-schema`

zaxovaiko:
### Dependency updates:

* Updated `peerDependencies` to include support for `simpl-schema` version `^3`
* Updated `devDependencies` to use the latest version of `@types/simpl-schema`.

Ref: #1403


packages/uniforms-bridge-simple-schema-2/package.json
```diff
@@ -38,14 +38,14 @@
     "tslib": "2.8.0"
   },
   "peerDependencies": {
-    "simpl-schema": "^1.0.0 || ^0.0.4",
+    "simpl-schema": "^1.0.0 || ^0.0.4 || ^3",
```
piotrpospiech, line 41: What about `v2`?

piotrpospiech (changes requested):

packages/uniforms-bridge-simple-schema-2/package.json
zaxovaiko: There is no version `2`.
image

Monteth (approved):

packages/uniforms-bridge-simple-schema-2/package.json
zaxovaiko: Lol, ok. I've added `^2` as well, in case someone uses it.

piotrpospiech: No. Don't add it. You missunderstood this.

There is a [Meteor simpl-schema](https://github.com/Meteor-Community-Packages/meteor-simple-schema) package (available in the [atmosphere](https://atmospherejs.com/) registry) and there is a [NPM simpl-schema](https://github.com/longshotlabs/simpl-schema) package, a successor to the Meteor simpl-schema package.

In the `uniforms-bridge-simple-schema-2`, we support the NPM package. In the deprecated `uniforms-bridge-simple-schema`, we had integration with the Meteor package.

As you wrote previously, there is no `v2` of the `simpl-schema`. I only mentioned that NPM package is successor of the Meteor package (so it is basicly the `v2`, but they started versioning the package from the `v1` or `v0` to be precise).

`"simpl-schema": "^3.0.0 || ^1.0.0 || ^0.0.4"` is the correct way to support all package versions.

Monteth: I doubt so; even if there is v2 somewhere, it is not under the `simpl-schema` npm name, so it's redundant. But do we need to care? I don't know.

piotrpospiech (changes requested):
Do we even need to support all versions? We can't guarantee that using previous versions would be available without problems. For example, TS types were changed. There are a lot of breaking changes in the package itself. I would prefer to support the latest major version of the package.

However, droping previous versions would be a breaking change.

packages/uniforms-bridge-simple-schema-2/package.json
zaxovaiko: @piotrpospiech dropping the old version might cause some backward compatibility problems, as you said. Let's do that later.

